PR #42, opened 2023-04-01
Adding a basic Discord rich presence

TheFatBastid:
This adds a basic discord rich presence. It isn't very dynamic at the moment. It just shows the game name and the time playing.
I figure once things get settled a little more it can be expanded to include the song they are playing, what instrument they are playing, if in a multipler lobby, whatever we feel like adding etc etc etc.
Right now I don't think a privacy option is totally needed but in the future I bet some people won't want others to know what song they playing or their score or something like that.


EliteAsian123:
Hi, thanks for your pull request! I've made some code-styling changes and moved the `DiscordController` class, but other than that, it looks good! Thanks very much for your contribution!

PR #54, opened 2023-04-07
very messy settings menu (in the pause menu)

Roystermeat:


EliteAsian123:
I'm working on a cache rework, so I'll merge when I'm done with that. Thanks!

PR #78, opened 2023-04-12
Initial scoring system

muskit:
Full scoring implementation for `FiveFretTrack`: this is the only track type that actually contributes to the score as of right now, though I've tried to set up the code in such a way that allows for adding of other tracks (sustain scoring for `RealGuitarTrack`, in particular, requires more refactoring before implementing).

Font for the score display (`PlayScene/Canvas/ScoreBox/Text`) needs to be changed to match mock-up.


EliteAsian123:
I've merged this to “new score screen.” Work still has to be done. Thanks very much!

PR #79, opened 2023-04-12
Move some Utils methods to more relevant locations, and other misc. changes

TheNathannator:
Just a few small things I couldn't keep away from doing lol

Most of the things in Utils were rather specialized and don't make sense as general-purpose utility methods imo. It's often good practice to try and avoid utility classes, as most of the time the methods there can be placed where they're relevant instead.

Other changes:

- Make sources URL constant's name const-case
- Update LiberationSans SDF - Fallback.asset because Unity kept modifying it lol


EliteAsian123:
Thanks!

PR #87, opened 2023-04-14
Fix bot overstrumming after adding strum leniency

RileyTheFox:
As the bot is set to strum on every note, this causes overstrumming with chords as the event is fired for every note in the chord, this PR fixes that by only setting strum leniency in normal play


EliteAsian123:
Will merge for now, a better solution may be needed later.

PR #94, opened 2023-04-15
Implement new audio engine

RileyTheFox:
Rewrote audio engine code to use BASS instead of Unity.

Removed the delta time option for song time. All song time now uses audio position from BASS (as it's more reliable)

Work done:
- Add IAudioManager interface for implementing audio libraries
- Added BASS implementation of IAudioManager
- Initialize BASS
- Loading songs and SFX
- Stem muting
- Play and pause

Things which need adding:
- Volume control for stems and SFX
- New sound effects in StreamingAssets/sfx folder


RileyTheFox:
Added some reverb to the stems when star power is active. It only applies reverb to the instruments/players which have star power activated. The effect might need some adjusting as well as a volume increase, which I tried but didn't notice a change so a gain DSP might need to be written

PR #96, opened 2023-04-15
Fix lefty flip on 4 and 5 lane drums

### LINKED ISSUE #56: Implement lefty flipped drums correctly.
Lefty drums does two things:
* Mirrors the chart while keeping the color of the lanes the same. (This means we now have a red cymbal)
* Makes the green cymbal input trigger a red cymbal.

It its not a simple mirroring of the screen like guitar is.

For 5 lane something similar can be done here.

----
Rock band 3 screenshots for reference

Righty drums:
Screenshot_2023-04-07_18-11-48

Lefty Drums:
Screenshot_2023-04-07_18-12-05

santorfo:
All that was done was reversing the order of the drum colors back to the original, whilst the chart is still flipped, as expected.

Fixes #56

### DIFF
--- a/Assets/Script/PlayMode/DrumsTrack.cs
+++ b/Assets/Script/PlayMode/DrumsTrack.cs
@@ -42,14 +42,6 @@ namespace YARG.PlayMode {
 
 			noKickMode = SettingsManager.GetSettingValue<bool>("noKicks");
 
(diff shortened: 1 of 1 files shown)


DarkRTA:
this does not fix #56 as the inputs are still incorrect

PR #99, opened 2023-04-15
Fix lefty flip drum inputs

### LINKED ISSUE #56: Implement lefty flipped drums correctly.

santorfo:
#96 did not fix the inputs, only the note colors, as DarkRTA mentioned.
I've now flipped the inputs when they're being processed as a hit, except when bot mode is on.
Fixes #56 for real this time.

### DIFF
--- a/Assets/Script/PlayMode/DrumsTrack.cs
+++ b/Assets/Script/PlayMode/DrumsTrack.cs
@@ -186,6 +186,28 @@ namespace YARG.PlayMode {
 		}
 
 		private void DrumHitAction(int drum, bool cymbal) {
+			// invert input in case lefty flip is on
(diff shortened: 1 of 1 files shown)


santorfo:
@DarkRTA this should fix #56 for good now

PR #116, opened 2023-04-17
(DC-16) Fix bug with strumming non-overlapping sustains

raphaelgoulart:
Hit detection ignores notes that are currently being held, _regardless of whether it's an overlapping sustain or not_. Changed so it only happens on extended sustains.
(Please ignore all commits except latest two)


EliteAsian123:
For future reference: https://discord.com/channels/1086048856678084609/1097123850451419206

PR #133, opened 2023-04-19
Added music player and shuffled around fonts

TheFatBastid:
No fonts should be under TMP's dir. This disabled the <font> rich tag as well. Thanks to Kadu for the music player buttons!


EliteAsian123:
I'll merge this after the graphics update

EliteAsian123:
Sorry this merge is taking so long. I'm just waiting on full CON support, as that will probably change how some things are handled.

EliteAsian123:
FINALLY! I've just merged manually at this point, since so much has changed. I've taken your code and refactored it, and it's now in! Sorry for the *very very very very very very very very very very* long wait. My bad for not merging it when it was first opened.

PR #144, opened 2023-04-20
Taps

NarrikSynthfox:
Tap notes, visuals need to be reworked but functionality is there.


NarrikSynthfox:
Don't merge until after merging #151

After that is merged, I'll sync the taps branch and implement the model for the tap notes

NarrikSynthfox:
Tap models are now implemented.

PR #147, opened 2023-04-20
add functional helpbar, read more

jtanan44:
Adds a functional help bar to all current menu screens. replaces stock unity ui back button in several instances. Adds animation for menu buttons on entry (not controlled). changes the "update available" button to not display new version string, and add Kadu's UI for it


RileyTheFox:
These changes are cool, I like them! I do have some feedback for them:

I think the update box should use a darker blue background. It's currently a bit hard on the eyes to read the text when it's white and hovered over.
image

Perhaps also greyscale the image used for the button so the colour can be configured in Unity?

I also think the current version should still be visible when there is an update available, it's useful for the information to be visible even if the client is outdated. Maybe reposition the update button to accommodate for this?

I believe the buttons on the help bar should be interactable by clicking the text as well as the coloured button. Using a menu with a mouse it's intuitive to attempt to click the text, while the coloured button represents what you'd press on a controller. If the button was grouped with the text and both were clickable, I think that would be a lot more fluent to interact with!
image

jtanan44:
Thanks for the feedback! Kadu's layout had it covering the version string but I can move it down instead! also good call on making it gray scale!
I'll also make the text clickable good call!

PR #151, opened 2023-04-21
Notes Update!

kaduwaengertner:
+ Updated HOPO Model, easier to see now in intense songs
+ Added Tap Note Model


NarrikSynthfox:
After this is merged, I'll sync the taps fork and implement the tap model for them.

kaduwaengertner:
Quick note: In the Note prefab need to change Material ID for Tap from 2 to 1, forgot to change it

PR #162, opened 2023-04-21
Open HO/POs

raphaelgoulart:
This reads open HO/POs, both auto and forced
Pending visuals, currently looks the same as regular opens (I'll most likely need help with this)


kaduwaengertner:
Starting the texturing for this right now, will also rework the overall texture for normal open notes/kicks cause I don't like the current ones

EliteAsian123:
Will wait until graphics are done

PR #171, opened 2023-04-22
Fixed Tap Textures

kaduwaengertner:
+ Fixed Tap materials being applied in the wrong place


kaduwaengertner:
+ Added Kick/Open fret in the .blend file
- Need to remove the current .fbx we're using

PR #174, opened 2023-04-23
lefty flip pro drums red cymbal input fix

### LINKED ISSUE #173: red cymbal input does not work in lefty pro drums
the green cymbal input in lefty pro drums should trigger the red cymbal instead

santorfo:
Fixes #173 

In the previous implementation, the lefty flip red cymbal input was being read as a regular red cymbal, which doesn't exist. 
Pro Drums lefty flip implies the player has moved the Green cymbal above the Red snare, so whilst the chart has flipped, and so have the inputs, that specific input stays the same because the input object was physically moved.

### DIFF
--- a/Assets/Script/PlayMode/DrumsTrack.cs
+++ b/Assets/Script/PlayMode/DrumsTrack.cs
@@ -237,7 +237,11 @@ namespace YARG.PlayMode {
 						drum = kickIndex == 4 ? 1 : 2;
 						break;
 					case 3:
-						drum = kickIndex == 4 ? 0 : 1;
+						// lefty flip on pro drums means physically moving the green cymbal above the red snare
+						// so while the position on the chart has changed, the input object is the same
+						if (!cymbal){
+							drum = kickIndex == 4 ? 0 : 1;
+						}
 						break;
 					case 4:
 						if (kickIndex == 5) {


Assets/Script/PlayMode/DrumsTrack.cs
```diff
@@ -237,7 +237,11 @@ public sealed class DrumsTrack : AbstractTrack {
 						drum = kickIndex == 4 ? 1 : 2;
 						break;
 					case 3:
-						drum = kickIndex == 4 ? 0 : 1;
+						// lefty flip on pro drums means physically moving the green cymbal above the red snare
+						// so while the position on the chart has changed, the input object is the same
+						if (!cymbal){
+							drum = kickIndex == 4 ? 0 : 1;
```
DarkRTA, line 243: have you checked and made sure this doesn't swap the yellow and blue cymbals as well?

DarkRTA (changes requested):

Assets/Script/PlayMode/DrumsTrack.cs
DarkRTA: nvm the diff didn't show the rest of the file
this is correct

santorfo: is it not supposed to? The regular drums left flip has them swapped too...

PR #182, opened 2023-04-23
Fixing Fret.blend intersecting faces again

eckerj:


eckerj:
Also adding back the "using System" for linux builds.

PR #183, opened 2023-04-23
Adds a "Restart" option to the pause menu

zigguy:
like the title says it adds a restart option to the pause menu.
Capture
also this is my first time coding in c# so im sorry if the code is bad.


zigguy:
not sure why the commit says i deleted the entire file and then added all the lines back, i only added a few lines, didnt delete anything

PR #188, opened 2023-04-24
Skip songs with errors loading song.ini when filling the song list

santorfo:
Previous implementation of filling the song list did skip songs without notes.mid files but whilst it did log errors about failure to parse song.ini files, it still added said broken songs to the songlist, rendering the search function unusable.

When a parsing error occurs, an exception is thrown and caught by the code that fills up the songlist and the song is skipped.

I also changed the behaviour when the "song" section isn't found in song.ini, because the song would show up as a blank entry in the songlist and break the search function too.


EliteAsian123:
Thanks for the PR! Please follow styling rules next time!

PR #196, opened 2023-04-27
Turn null artist information in songs.dta into Unknown Artist

santorfo:
This fixes searching by artist being broken when a RB CON doesn't have artist information in its songs.dta file.

### DIFF
--- a/Assets/Script/Serialization/Xbox/XboxSong.cs
+++ b/Assets/Script/Serialization/Xbox/XboxSong.cs
@@ -88,7 +88,11 @@ namespace YARG.Serialization {
 			// song.delay
 			song.drumType = rb ? SongInfo.DrumType.FOUR_LANE : SongInfo.DrumType.FIVE_LANE;
 			// song.hopoFreq
-			song.artistName = songDta.artist;
+			if (songDta.artist != null){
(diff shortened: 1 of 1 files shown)


Assets/Script/Serialization/Xbox/XboxSong.cs
```diff
@@ -88,7 +88,11 @@ public class XboxSong {
 			// song.delay
 			song.drumType = rb ? SongInfo.DrumType.FOUR_LANE : SongInfo.DrumType.FIVE_LANE;
 			// song.hopoFreq
-			song.artistName = songDta.artist;
+			if (songDta.artist != null){
+				song.artistName = songDta.artist;
+			}else {
+				song.artistName = "Unknown Artist";
+			}
```
sonicfind, line 95: It would be better to match how `song.drumType` is set. It uses the ternary ? operator to keep the expression all on one line.

In essence, change your if-else to `song.artistName = songDta.artist != null ? songDta.artist : "Unknown Artist";`

Keeps consistency.

santorfo: I get what you mean, I've changed it to your suggestion and fixed a blank line that wasn't there before

PR #197, opened 2023-04-27
(See changelog) New Kick Flash Animation + bunch of other graphical stuff

berthrage:
Mia's Nightly Changes (04/27) Changelog:
+ New Kick Flash animation when using kicks (thanks Venerabella for the sprites);
+ New particles for Kick Flash;
+ New particles for broken glass on hit effects;
+ Hit effects particles adjusted slightly;
+ Inner green fret color darkened to match the rest;
+ Shadows on the Highway Track adjusted and are now darker on the end;
+ Combo Meter Wing pulled backwards, Sunburst is now on top of it as it should;
+ Sunburst color adjusted slightly; 
+ Combo Meter Base now changes textures based on Multiplier/SP state (Texture credits: Kadu);
+ Sustain Particles should now look more even with the frets;
+ FC Ring emission improved;
+ Overdrive Bar materials improved and should now look more metallic;
+ Kick Camera Shake NaN bug fixed, although it lost some functionality as a result (initial kick is immediate rather than being a really quick lerp, I suggest only trying to fix this when new tweener libraries are in);

To fix: Particles are not being colored properly through function Colorize() from ParticleGroup.cs


EliteAsian123:
Broken PR. Waiting for new (by me!)

PR #199, opened 2023-04-27
Basic .chart support

RileyTheFox:
Adds support for .chart. It currently has minimal features but should work with most charts.

- Beats are now a structure and are generated using an algorithm for .chart songs
- Drums are not supported in .chart yet, only 5 fret guitar game modes.


RileyTheFox:
If anyone can figure out why the star power length is being calculated too long in YargChart.cs in the LoadArray() function I'd appreciate it because I have no clue

RileyTheFox:
> If anyone can figure out why the star power length is being calculated too long in YargChart.cs in the LoadArray() function I'd appreciate it because I have no clue

Fixed this

PR #231, opened 2023-05-01
Unleash the Chipmunks and the Demons

RileyTheFox:
ChanGed **H**ow s**E**mitone sHifting is ca**L**culaTed for Chi**P**munk SpeeDups...


EliteAsian123:
Going to be my favourite patch note

Issue #227, opened 2023-05-01
Solo bonus is not 0 points when solo result is "Messy Solo"


EmperorJacoba:
If you botch a solo, and it gets the "Messy Solo!" result at the end of it, YARG still rewards points when it should not. Clone Hero nor Rock Band have never rewarded points for a messy solo.

PR #234, opened 2023-05-01
Fix bonus solo points being awarded when solo completion is MESSY (50-59%)

muskit:
Adjusted solo bonus scoring minimum to be at 60% instead of 50%.

Addresses #227


EliteAsian123:
Fixes #227

PR #256, opened 2023-05-06
Added toasts

TheFatBastid:
There are 5 public calls that can be made, ToastWarning(), ToastMessage(), ToastError(), ToastInformation(), and ToastGeneral(). Should be used similarly to debug.log. Added an example in UpdateChecker. Added 2 internal checkers as well (devices at start, devices connected/disconnected). Had to move the loading abort button to the other side. Thanks to Kadu for the art!


Assets/Script/UI/ToastManager.cs
```diff
@@ -0,0 +1,208 @@
...
+
+    public enum ToastType{
+        general,
+        success,
+        warning,
+        information,
+        error,
+    }
+
+    public struct toastStruct{ //this is so we can pass 2 things into the queue
```
EliteAsian123, line 74: Please follow C# naming conventions (should be `ToastStruct`, or even better just `Toast`)

Assets/Script/UI/ToastManager.cs
```diff
@@ -0,0 +1,208 @@
...
+        public ToastType type;
+        public string text;
+        public toastStruct(ToastType _type, string _text){
```
EliteAsian123, line 77: You could change these argument names to just `type` and `text`, and convert the two lines below to:
```cs
this.type = type;
this.text = text;
```

Assets/Script/UI/ToastManager.cs
```diff
@@ -0,0 +1,208 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using TMPro;
+using UnityEngine.UI;
+using UnityEngine.InputSystem;
+public class ToastManager : MonoBehaviour
+{
+    /* Currently handles:
```
EliteAsian123, line 9: Please use tabs, not spaces.

EliteAsian123 (changes requested):
Thanks for the PR, I've been wanted to add toasts, but I'm too lazy. Appreciated as always! Please look at review comments, and then I can merge.

EliteAsian123 (approved):

PR #262, opened 2023-05-06
Added some clarification

TheFatBastid:
"Be sure it is added to system path." twice and "greater" for python


EliteAsian123:
Thanks

PR #273, opened 2023-05-07
fix: browse button not working on standalone mac

Pantotone:
Apparently, the last official update for the StandaloneFileBrowser was in 2018, before the adoption from Apple to AppleSilicon in 2021. Because of this, it creates a bug when creating a universal build on Unity (Intel + AppleSilicon) making the "Browse" button from the song folder manager not work anymore.

This replaces the current StandaloneFileBrowser.bundle and uses a .bundle (+ .meta) created by tonidurans that has added support to AppleSilicon

Tested using an Intel-based machine running an Intel + Apple Silicon (universal) build. Still need testing for Apple Silicon.

 tldr: i just grabbed the .bundle + .meta from this pr (https://github.com/gkngkc/UnityStandaloneFileBrowser/pull/116) and put on the project and now it works 👍


EliteAsian123:
Unfortunately, this seems to completely break it on Windows. I'll do some investigation.

EliteAsian123:
Okay, nvm? Seems to be working now. Odd.

PR #278, opened 2023-05-08
Gold star progress graphic tweak

muskit:
Slightly decrease the alpha of the top bar to better blend it in with the GS progress sprite.


EliteAsian123:
This merges to `master`. Please use `dev` now.

PR #284, opened 2023-05-08
drumsfx

YoItsOJ:
Freestyle drums for before and after songs, and, eventually, fills.


RileyTheFox:
This PR should be based on the `dev` branch, not `master`.

Also, please do not pull assets from other games. We do not want to take any assets directly from other games, we create original assets with inspiration from existing material.

YoItsOJ:
> 

They're meant to be temporary, I assumed one of the artists would create some original ones

PR #292, opened 2023-05-09
Dev Watermark

nevespt:
[Added]
- a script to check if is beta version or not

[Changed]
- Elements at persistant Scene

### DIFF
--- a/Assets/Script/UI/DevWatermark.cs
+++ b/Assets/Script/UI/DevWatermark.cs
@@ -1,0 +1,31 @@
+using System.Collections;
+using System.Collections.Generic;
+using TMPro;
+using UnityEngine;
+using YARG;
+using Debug = UnityEngine.Debug;
+
+public class DevWatermark : MonoBehaviour
+{
+
+    public TextMeshProUGUI watermarkText;
+    
+    // Start is called before the first frame update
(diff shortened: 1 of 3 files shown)


Assets/Script/UI/DevWatermark.cs
```diff
@@ -0,0 +1,31 @@
+using System.Collections;
```
EliteAsian123, line 1: Please follow styling (no newline before `{`), and use tabs instead of spaces!

Assets/Script/UI/DevWatermark.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using TMPro;
+using UnityEngine;
+using YARG;
+using Debug = UnityEngine.Debug;
+
+public class DevWatermark : MonoBehaviour
+{
+
+    public TextMeshProUGUI watermarkText;
```
EliteAsian123, line 11: This can be made private and `[SerializeField]` can be added. This allows it to be seen in the editor, while staying private.

EliteAsian123 (changes requested):

EliteAsian123 (approved):

PR #295, opened 2023-05-10
Abstract Track Performance Text (BASS GROOVE, HOT START, etc.)

grishhung:
Some preliminary performance text for abstract tracks. We should really consider making it so that hot start and strong finish require fewer than 25 notes for non-expert difficulties, though, as those charts are generally less populous (check GH for reference).

Edit: On second though, perhaps not. Seems like Guitar Hero makes it 25 notes even for other difficulties. Also just found a slight oversight wherein STRONG FINISH can appear *before* the last note is hit. Will make a check for that in the next commit.


grishhung:
Regarding the X00-NOTE STREAK counter: We most likely want to have this disabled by default. A lot of players have mentioned that it gives them anxiety lol

grishhung:
Alright, most everything is now squared away.

The only thing left to consider (beside making each of the five message types toggleable is whether or not STRONG FINISH should appear after FULL COMBO (which is how it works in GH). When we discussed this, we mostly agreed that it's redundant.

However, if we wanna have a STRONG FINISH counter in the save file for stats and such, we should probably include it. For now, it'll just be one or the other, though. (It's a very easy change.)

PR #312, opened 2023-05-12
Try/catch added to con reading to prevent crashing

TheFatBastid:
Without this, when an "access denied" folder is scanned, it will throw an IO error and end the scanning, and thus no cache is created.


Assets/Script/UI/MusicLibrary/SongSelection.cs
```diff
@@ -60,7 +60,7 @@ public class SongSelection : MonoBehaviour {
 					_selectedIndex -= _songs.Count;
 				}
 
-				if (_songs[value] is SongViewType song) {
+				if (_songs[_selectedIndex] is SongViewType song) {
```
EliteAsian123, line 63: Is this an intentional change?

Assets/Script/Song/Scanning/SongScanThread.cs
```diff
@@ -190,45 +190,48 @@ public class SongScanThread {
...
 					}
 				}
-			}
+				string[] subdirectories = Directory.GetDirectories(subDir);
 
-			string[] subdirectories = Directory.GetDirectories(subDir);
-
-			foreach (string subdirectory in subdirectories) {
-				if(subdirectory != Path.Combine(subDir, "songs_updates"))
+        foreach (string subdirectory in subdirectories) {
```
EliteAsian123, line 229: This line uses spaces as indentation

EliteAsian123 (changes requested):

Assets/Script/UI/MusicLibrary/SongSelection.cs
TheFatBastid: Yes, this fixes an array out of bounds error when scrolling past the upper and lower limits of the song select screen. "Value" is not changed whereas _selectedIndex is clamped correctly when scrolling.

Assets/Script/Song/Scanning/SongScanThread.cs
TheFatBastid: Whoops, okay changed

EliteAsian123 (approved):

EliteAsian123:
Thanks!

PR #319, opened 2023-05-13
Limit song speed to 4 characters

lunalawl:


EliteAsian123:
This should probably be done in Unity via the inspector instead of the script.

EliteAsian123:
Done via inspector in e5ff94897866be1c077b2a51c3aad9a0f100b0b1

PR #328, opened 2023-05-15
Blank sidebar on header selection, add phrase

TheFatBastid:
Blanked the album sidebar info. It was easy to add number of sources but all other info seemed to be harder to get (afaik, meaning i'd have to play with the scanner which I didn't want to do). Also, added a the loadingPhrase text to the helpbar. Shown to Kadu beforehand. 

Apologies for 4bcbc92 (which had some old meta file removed) and its reversion in d990731


Assets/Script/Data/SongSources.cs
```diff
@@ -275,5 +275,9 @@ public static class SongSources {
 
 			return "Unknown Source";
 		}
+
+		public static int GetSourceCount() {
+			return webSourceNames?.Count ?? DEFAULT_SOURCES.Count;
```
EliteAsian123, line 280: Keep in mind that `webSourceNames` is *added* on to `DEFAULT_SOURCES`. If a source isn't in `webSourceNames`, it is instead fetched from `DEFAULT_SOURCES`.

EliteAsian123 (changes requested):

TheFatBastid (commented):
GetSourceCount changed to return just the webSourceNames.Count

EliteAsian123 (approved):

PR #330, opened 2023-05-15
Github Actions for building and packing macOS executables

Pantotone:
Those are Github Actions/Workflows for building and packing a macOS executable inside a Mac environment.
Hopefully it resolves the "The application can’t be opened" messages.

**build-release-mac.yml:** _(for building and packing all remotly inside a Mac environment)_
This action runs manually, needing to go to **Actions > Builds - MacOS (release) > Run workflow** (and putting version and checkmarks for .zip/.tar.bz2/.dmg)
and after building (about 15 minutes/40 for first run-without caching), the results are on the bottom of the summary page of the run, at the Artifacts section ready to be released.

Before triggering the action, it needs to setup the secrets for a Unity Personal License. (probably better use a dummy account just for building stuff to avoid problems)
The steps are here (https://game.ci/docs/github/activation/#converting-into-a-license), and can also call me on Discord if needing help

**fix-mac-released.yml:** _(for TRYING TO recovering past builds and packing with the lost attributes)_
This action also runs manually, with "Fix macOS permissions on latest released build" name. This one will download the latest or a specified version from releases, unzip all, add attributes and permission, and repack in .dmg/.zip/.tar.bz2

!! But this one is kinda flaky, some builds work (I think the v0.10.0 works fine), but others still gave an error (latest one for example), so it's not the best to rely on this one

For both actions it still has the "Apple couldn't verify this file for malware" since we don't have an Apple Developer account and certificates, still needing to Shift + Right Click > Open to bypass the message

If needing help to set this up, feel free to call me on Discord
(also the actions only work on master afaik)


EliteAsian123:
Awesome! Thank you so much!

EliteAsian123:
I actually gotta wait until this gets merged to `master`, so I'll wait until then.

PR #340, opened 2023-05-16
Update FPS Counter with Memory status Now

nevespt:
Replace the Green FPS counter and Memory status with the new one. When in Editor the counter will be Persistent even if you disable from settings

[Changed]
- Update the layout of the components at PersistantScene
- Update the Settings Callback to enable always in Editor to override the settings
- Commented the Old GUIUpdate function
- Added the Memory code in FpsCounter.cs


EliteAsian123 (changes requested):

nevespt:
all done

EliteAsian123 (approved):

PR #344, opened 2023-05-17
Fixed the warnings about font not being readable

TheFatBastid:
If you go to credits or a few other places you will be warning spammed with "Unable to add the requested character to font asset atlas texture. Please make the texture readable." A total unity moment: 1) This can only be set when viewing the font atlas, while running, in the debug inspector, not the regular one. 2) They confuse reading and writing. 3) Why isn't this the default? Whatever, anyways, I set the Barlow Font to IsReadable true


EliteAsian123:
Conflicts :(

TheFatBastid:
Okay there we go

PR #385, opened 2023-05-28
Added navigation buttons "Next section" and "Random song"

SamuelPalma:
* Added navigation button (Blue) "Next section" that selects the next letter with respect to the current song.

* Navigation button (Orange) is added to select a random song.
NextSectionAndRandomSongButtons

### DIFF
--- a/Assets/Script/UI/MusicLibrary/SongSelection.cs
+++ b/Assets/Script/UI/MusicLibrary/SongSelection.cs
@@ -117,12 +117,12 @@ namespace YARG.UI.MusicLibrary {
 				new NavigationScheme.Entry(MenuAction.Shortcut1, "Search Artist", () => {
 					SearchByArtist();
 				}),
-				new NavigationScheme.Entry(MenuAction.Shortcut2, "Next section", () => {
-					SelectNextSection();
-				}),
-				new NavigationScheme.Entry(MenuAction.Shortcut3, "Random song", () => {
+				new NavigationScheme.Entry(MenuAction.Shortcut2, "Random song", () => {
 					ClearSearchBox();
 					SelectRandomSong();
+				}),
+				new NavigationScheme.Entry(MenuAction.Shortcut3, "Next section", () => {
+					SelectNextSection();
 				})
 			}, false));
 
@@ -337,17 +337,17 @@ namespace YARG.UI.MusicLibrary {
 				SelectedIndex = Mathf.Max(1, index);
 			}
 
-			setFirstLetters();
+			SetFirstLetters();
 			UpdateSongViews();
 			UpdateScrollbar();
 		}
 
-		private void setFirstLetters(){
+		private void SetFirstLetters(){
 			songsFirstLetter =
 				_songs
 				.OfType<SongViewType>()
 				.Select(song => song.SongEntry.NameNoParenthesis)
-				.Where(name => name != null && name.Any())
+				.Where(name => !string.IsNullOrEmpty(name))
 				.Select(name => Char.ToUpper(name[0]))
 				.Distinct()
 				.OrderBy(ch => ch)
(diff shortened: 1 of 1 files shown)


Assets/Script/UI/MusicLibrary/SongSelection.cs
```diff
@@ -334,10 +337,23 @@ public class SongSelection : MonoBehaviour {
 				SelectedIndex = Mathf.Max(1, index);
 			}
 
+			setFirstLetters();
 			UpdateSongViews();
 			UpdateScrollbar();
 		}
 
+		private void setFirstLetters(){
```
EliteAsian123, line 345: Follow naming conventions (should be `SetFirstLetters`)

Assets/Script/UI/MusicLibrary/SongSelection.cs
```diff
@@ -334,10 +337,23 @@ public class SongSelection : MonoBehaviour {
...
+			songsFirstLetter =
+				_songs
+				.OfType<SongViewType>()
+				.Select(song => song.SongEntry.NameNoParenthesis)
+				.Where(name => name != null && name.Any())
```
EliteAsian123, line 350: `string.IsNullOrEmpty` may be better here.

Assets/Script/UI/MusicLibrary/SongSelection.cs
```diff
@@ -459,5 +475,53 @@ public class SongSelection : MonoBehaviour {
...
+					return song is SongViewType songType &&
+						songType.SongEntry.NameNoParenthesis.Substring(0, 1) == nextCharacter
+					;
+				});
+
+			SelectedIndex = index;
+		}
+
+		private string GetNextLetterOrNumber(string input){
+			string firstCharacter = input.Substring(0, 1).ToUpper();;
```
EliteAsian123, line 510: This should become `input[0].ToUpper()`. A `string.IsNullOrEmpty` check should also be added above to prevent errors. If an error state is needed (or in this case, `""` is returned), it may be favorable to return `char?` (where `""` would become `null`).

Assets/Script/UI/MusicLibrary/SongSelection.cs
```diff
@@ -459,5 +475,53 @@ public class SongSelection : MonoBehaviour {
...
+
+			int indexOfActualLetter = songsFirstLetter.FindIndex(letter  => {
+				return Char.ToString(letter) == firstCharacter;
```
EliteAsian123, line 513: If above comment occurs, this can become `letter == firstCharacter`

Assets/Script/UI/MusicLibrary/SongSelection.cs
```diff
@@ -459,5 +475,53 @@ public class SongSelection : MonoBehaviour {
...
+			});
+
+			bool isLast = indexOfActualLetter == (songsFirstLetter.Count() - 1);
```
EliteAsian123, line 516: Use `songsFirstLetter.Count` instead of `songsFirstLetter.Count()`.

Assets/Script/UI/MusicLibrary/SongSelection.cs
```diff
@@ -459,5 +475,53 @@ public class SongSelection : MonoBehaviour {
...
+		}
+
+		private void SearchByArtist(){
+			if (_songs[SelectedIndex] is SongViewType view) {
+				searchField.text = $"artist:{view.SongEntry.Artist}";
+			}
+		}
+
+		private void SelectNextSection(){
+			if (_songs[_selectedIndex] is not SongViewType song) {
```
EliteAsian123, line 493: `song` variable is not used.

Assets/Script/UI/MusicLibrary/SongSelection.cs
```diff
@@ -459,5 +475,53 @@ public class SongSelection : MonoBehaviour {
...
+				return;
+			}
+
+			int skip = Mathf.Max(1, _songs.Count - SongContainer.Songs.Count);
+			string nextCharacter = GetNextLetterOrNumber(song.SongEntry.NameNoParenthesis);
+
+			var index = _songs.FindIndex(skip, song => {
+					return song is SongViewType songType &&
```
EliteAsian123, line 501: The lambda can be simplified to `song => [RETURN VALUE]` instead of `song => { return [RETURN VALUE] }`

EliteAsian123 (changes requested):

TheNathannator:
I'd suggest making orange the button to go to the next section, to match how RB/CH do it.

Random song might also be better in a separate menu that contains other less common options as well, but I don't think we have that implemented currently so a button should be fine for now.

SamuelPalma:
Hi @EliteAsian123 , I have already made the requested changes.
Thank you for the review 😄

EliteAsian123 (approved):

PR #411, opened 2023-06-02
Display Settings (Fullscreen & Resolution)

Roystermeat:
Yes the resolution dropdown is in another scene. I did it that way because I couldn't figure out the dropdown EA made so I just used a TMP_Dropdown which works. Fullscreen & Resolution do work as intended.


EliteAsian123:
Merged manually in c1aa5a584f796793f27401040f42aa61502cdb34

PR #427, opened 2023-06-05
Even more smaller fixes

TheNathannator:
Post-release bug reports, yay! Except for the 4th commit, I just happened to realize that was a thing that could happen lol


TheNathannator:
Drafted while I investigate one more pertinent issue, just got more info on it.

EliteAsian123:
You're a legend

PR #449, opened 2023-06-09
Fixed year display for songs from The Beatles Rock Band

EscapeNumber001:
* The Year field for songs from the source `tbrb` or `tbrbdlc` will now be truncated to display only the year, excluding the month and day.


EliteAsian123:
This should probably be using a more general approach that extends to all different date types that isn't dependent on the source. At its current state, if someone manually changes the values in the song.ini, it could crash the caching process.

EscapeNumber001:
I just wrote a new system that checks for four contiguous digits in the Year string and replaces the year with that substring.  This system is source-independent and should work with basically any date format.

PR #489, opened 2023-06-25
player intro callouts (READ BEFORE MERGE)

jnackmclain:
Needs some love for sure. Couple things. For some reason it's really inconsistent. I try to only run it once when the song starts, and that only works sometimes? but when it works it works. Mic players don't have a callout tag. i have no idea how to add in for that. Also the FC animation I did is in here, just completely disabled, to avoid merge conflicts.


jnackmclain:
oh also, when it does work, it starts already animated out
there is a full in and out animation on the callout tag
hoping that can be shown in full, it seems to start animating before the load is complete

TheNathannator:
I think it's about time we let this one go lol

The closed PR will still be here if we want to revisit this, but for now it needs to be redone entirely before merging.

PR #513, opened 2023-07-03
Finish whammy SFX implementation

TheNathannator:
This was not nearly as difficult as I thought it would be, at least if you exclude the hour or two I spent trying to work out a crash lol

I added a new AudioOptions class to hold all of the options for the audio backend, so that instead of having to put them in both the interface and the implementations, they can just be put into the options class and used where needed. Along with that, I added a bunch of new options for the whammy SFX:

- Enable/disable whammy SFX (completely skips creating the FX handles)
- Whammy oversample factor
- Whammy pitch shift amount (defaults to 1 semitone, goes up to a full octave)
- Tried to also allow changing the FFT size, but that crashes BASS_FX, so it's a no-go

I also added whammy support to Pro Guitar, including both the pitch bending and SP gain. As with all the other aspects of Pro Guitar input handling, this is currently hard-coded to only Pro Guitar controllers.

Additional whammy fixes/changes:

- The pitch bend FX handles are only created for channels which *can* be pitch-bent in the first place, which greatly reduces the processing load.
- The default oversample factor is now 8 instead of 32, which is much easier on lower-end hardware. Those that want a higher quality whammy can adjust the setting for it.
- A workaround has been implemented for the stem desync that BASS_FX's pitch bending causes, 'cause apparently it's not smart enough to account for the FFT buffer automatically lol
- The delta value that the guitar input strategy outputs is no longer inverted, it was subtracting the current from the previous instead of the other way around.

Other fixes/changes:
- The localization keys file has been re-sorted alphabetically, since that's just what Unity did when I went to edit it lol
- Pro Bass no longer reverbs the guitar stem instead of the bass stem.


TheNathannator:
Might need to wait a bit longer on this one, something's not quite right with the stems that are getting pitch-bent. They sound all washed out for some reason.

TheNathannator:
I'm thinking it's something inconsistent, just played through the same song again and there were no noticeable quality issues. We'll need more testing on this one lol, I'll leave things here for now.

TheNathannator:
okay this is dumb lol

For some reason, whammiable stems that aren't getting updated constantly will just gradually degrade in quality, manifesting in as little as 20-30 seconds. If I always update them, even if it's just constantly setting the percent to 0 while there's no sustains going on, then it doesn't happen.

PR #521, opened 2023-07-08
Small refactor around loading a song entry's audio

sonicfind:
+ Simplify the stem channel inheritance hierarchy a bit
+ Split the ExConSongEntry stem map into the separate stems - allows easier stem type-base audio removal when specified on audio load.
+ Move MoggInfoGeneration function into the ConEntry class directly


Assets/Script/Audio/Bass/BassMoggStemChannel.cs
```diff
@@ -1,190 +1,18 @@
...
 
 namespace YARG.Audio.BASS
 {
-    public class BassMoggStemChannel : IStemChannel
+    public class BassMoggStemChannel : BassStemChannel
     {
-        public SongStem Stem { get; }
-        public double LengthD { get; private set; }
-        public double Volume { get; private set; }
+        readonly public float left;
```
RileyTheFox, line 9: Proper formatting is access modifiers first, then field modifiers

Assets/Script/Audio/Bass/BassStemMixer.cs
```diff
@@ -215,21 +224,26 @@ public bool RemoveChannel(IStemChannel channel)
...
             return true;
         }
 
-        public IStemChannel GetChannel(SongStem stem)
+        internal static readonly List<IStemChannel> EMPTYCHANNELLIST = new();
+
+        public List<IStemChannel> GetChannels(SongStem stem)
         {
-            return !_channels.ContainsKey(stem) ? null : _channels[stem];
+            return !_channels.ContainsKey(stem) ? EMPTYCHANNELLIST : _channels[stem];
```
RileyTheFox, line 246: Use Array.Empty<T> here, it's cleaner than a static list. Internal is also redundant here as internal only restricts usage to an assembly, not per file/class

Assets/Script/Song/Types/ExConSongEntry.cs
```diff
@@ -527,6 +560,154 @@ public virtual void Update(string folder)
             }
         }
 
+        public void SetAudioValues(DataArray dta, List<DataArray> dta_update_roots = null)
```
RileyTheFox, line 563: Really not a fan of this. It bloats the song entry class when it is mostly intended to be used for storing fields related to caching. This class is just extremely huge now and it doesn't need to be. I'd recommend putting this elsewhere.

RileyTheFox: The LoadAudio functions are fine however

Assets/Script/Song/Scanning/SongScanThread.cs
```diff
@@ -387,7 +387,7 @@ private List<DataArray> UpdateAndUpgradeCon(ExtractedConSongEntry currentSong)
             {
                 _songsScanned++;
                 songsScanned = _songsScanned;
-                MoggBASSInfoGenerator.Generate(currentSong, currentArray.Array("song"), updateValue);
+                currentSong.SetAudioValues(currentArray.Array("song"), updateValue);
```
RileyTheFox, line 390: I kind of preferred the old system for this because the ExConSongEntry is just extremely bloated now

RileyTheFox (changes requested):
Mostly happy with the changes, I just think that ExConSongEntry is extremely bloated now as the purpose of the SongEntry classes were for storing song info to play the song or to cache it

Assets/Script/Song/Types/ExConSongEntry.cs
sonicfind: To be fair, this isn’t exactly how I want to it be like to start with. However, how I want to work would require completely redoing our current DTA loading system entirely. I’m fine with reverting to the generator for now (though it’s 3am, so that’ll have to wait, lol)

Assets/Script/Audio/Bass/BassMoggStemChannel.cs
sonicfind: Idk how I did that

RileyTheFox (approved):

PR #527, opened 2023-07-16
Fix 404 not found error

Crafty-The-Fox:
Fixes 404 not found error when trying to test play in unity when it checks for game updates. If you open the old URL in a browser, itl also do an error 404 not found. Not entirely sure why this hasnt been changed (?)
image


EliteAsian123:
Good catch, thanks!

PR #528, opened 2023-07-16
Fix 404 not found error when trying to update (rewrite branch)

Crafty-The-Fox:
fixes 404 not found error in rewrite branch

### DIFF
--- a/Assets/Script/ThirdParty/UpdateChecker.cs
+++ b/Assets/Script/ThirdParty/UpdateChecker.cs
@@ -42,7 +42,7 @@ namespace YARG
                 // Setup request
                 var request =
                     (HttpWebRequest) WebRequest.Create(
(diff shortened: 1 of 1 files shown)


RileyTheFox:
I'll merge this but in the short term future for the `rewrite` branch I'd like to avoid doing very small fixes to things which really only affect the `dev` and `master` branches, as most things will be getting changed regardless. We're mostly trying to do large system changes right now.

Thanks though!

Crafty-The-Fox:
> I'll merge this but in the short term future for the `rewrite` branch I'd like to avoid doing very small fixes to things which really only affect the `dev` and `master` branches, as most things will be getting changed regardless. We're mostly trying to do large system changes right now.
> 
> Thanks though!

ooohhh okay all good!

PR #530, opened 2023-07-21
Fix GitHub links leading to the wrong page

Supersonicboss1:
While normally the links would be redirected automatically, since EliteAsian123/YARG does exist now, the redirect no longer works.


EliteAsian123:
Good point, thanks!

PR #531, opened 2023-07-23
Add audio preview file support

sonicfind:


Assets/Script/Audio/PreviewContext.cs
```diff
@@ -76,7 +76,7 @@ public async UniTask PlayPreview(SongEntry song, float volume, CancellationToken
                 }
 
                 // Load the song
-                await UniTask.RunOnThreadPool(() => song.LoadAudio(_manager, 1f, SongStem.Crowd));
+                bool usesPreviewFile = await song.LoadPreviewAudio(_manager, 1f);
```
EliteAsian123, line 79: Awaiting this would do nothing because none of the overrides are asynchronous. Consider changing return type to just `bool`, or use `UniTask.RunOnThreadPool` within the overrides.

EliteAsian123 (changes requested):

sonicfind:
*burp*

EliteAsian123 (approved):

PR #534, opened 2023-07-26
Extra OnSongEnd call removed

TheFatBastid:
I can't think of reason it needs two here, there is one just 2 lines above this one


EliteAsian123:
Closing as removed in `rewrite` anyways

PR #541, opened 2023-08-12
Update Dev and Alpha Watermark

nevespt:


RileyTheFox:
I'm not sure I understand the change here. Why does this need changing?

nevespt:
bcz it wasnt showing before also Pre-release is diferent then dev builds

RileyTheFox:
We don't really differentiate what the build type is. It's either a stable release or not (prerelease, dev, whatever you want to call it)

nevespt:
so keep just the Pre-Release text whatever the version is 1.23.1b or .1.23.1-a12?

RileyTheFox:
Yeah the text doesn't really need changing. The version itself more or less denotes what the build type is.

PR #557, opened 2023-09-01
New Layout for Debug Text if more then 1 player

nevespt:
I added the code to change the Debug Text Position, size and anchor if more then 1 player is present, else stays as before

### DIFF
--- a/Assets/Script/Gameplay/GameManager.cs
+++ b/Assets/Script/Gameplay/GameManager.cs
@@ -134,6 +134,19 @@ namespace YARG.Gameplay
             Navigator.Instance.PopAllSchemes();
             GameStateFetcher.SetSongMetadata(Song);
 
+            // Change Debug Text Layout if more than 1 player
+            if (_yargPlayers.Count > 1)
+            {
+                _debugText.alignment = TextAlignmentOptions.TopLeft;
+                _debugText.fontSize = 25;
+                _debugText.font = Resources.Load<TMP_FontAsset>("Fonts & Materials/Barlow SDF");
+                _debugText.rectTransform.anchorMin = Vector2.up;
+                _debugText.rectTransform.anchorMax = Vector2.up;
+                _debugText.rectTransform.pivot = Vector2.up;
+                _debugText.rectTransform.anchoredPosition = new Vector2(25, -60);
+            }
+
+            // If no song, go back to menu
             if (Song is null)
             {
                 Debug.LogError("Null song set when loading gameplay!");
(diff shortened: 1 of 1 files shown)


nevespt:
**Exemple In-Game**
1 Player (Default):
image

More then 1 Player (New):
image

Assets/Script/Gameplay/GameManager.cs
```diff
@@ -134,6 +134,19 @@ private void Awake()
...
+            if (_yargPlayers.Count > 1)
+            {
+                _debugText.alignment = TextAlignmentOptions.TopLeft;
+                _debugText.fontSize = 25;
+                _debugText.font = Resources.Load<TMP_FontAsset>("Fonts & Materials/Barlow SDF");
+                _debugText.rectTransform.anchorMin = Vector2.up;
+                _debugText.rectTransform.anchorMax = Vector2.up;
+                _debugText.rectTransform.pivot = Vector2.up;
+                _debugText.rectTransform.anchoredPosition = new Vector2(25, -60);
+            }
```
EliteAsian123, line 147: I'm not sure if this whole part is necessary; these kinds of things should be configured in the inspector instead, and moving the debug test based on position is a bit of an over complication, especially for *debug* text. The user isn't really supposed to see it anyway. If you just move the text to the top left in the inspector, and commit that, I'll merge this.

Assets/Script/Gameplay/GameManager.cs
```diff
@@ -236,27 +249,28 @@ private void Update()
                     double starPower = fiveFretPlayer.Engine.EngineStats.StarPowerAmount;
 
                     _debugText.text =
-                        $"Note index: {noteIndex}\n" +
-                        $"Buttons: {buttonMask}\n" +
-                        $"Star Power: {starPower:0.0000}\n" +
-                        $"TicksPerEight: {ticksPerEight}\n";
+                        $"<b>Player:</b> {fiveFretPlayer.Player.Profile.Name}\n" +
```
EliteAsian123, line 252: The “Player” field only effects guitar players here, as the drum player one below has not been modified.

EliteAsian123 (changes requested):

Assets/Script/Gameplay/GameManager.cs
nevespt: the idea was to not change the default, but if more the 1 i can see the board and not the taxt overlay it, thats why i change the layout in code, but just do it at the start

Assets/Script/Gameplay/GameManager.cs
nevespt: var PlayerName = _players[0].Player.Profile.Name;
this should work ig

Assets/Script/Gameplay/GameManager.cs
RileyTheFox: Yeah I have to agree, its debug text only meant for developer use or getting additional info from a player when reporting a bug. It's not meant to look pretty or do anything special, it serves one function and that is to provide debug info during gameplay.

PR #570, opened 2023-10-06
Fixed race condition on beat subscribe

TheFatBastid:
Previously, when subscribing to BeatEventManger after the song started, the action would fire a billion times in a second to catch up to the current song tick, now it calculates and adds whatever was missed before subscribing.


EliteAsian123 (changes requested):

TheFatBastid (commented):
Removed all the Subscribe stuff and just changed the main loop

TheFatBastid (commented):
back to !

EliteAsian123 (approved):

PR #579, opened 2023-10-11
Stagekit implementation

TheFatBastid:
This is a port of the old engine code, not a re-write, so I welcome people to check it out to see if there are better ways to do things that I do not have knowledge of.

### DIFF
--- a/Assets/Prefabs/Gameplay/Visual/TrackElements/DrumsNote.prefab
+++ b/Assets/Prefabs/Gameplay/Visual/TrackElements/DrumsNote.prefab
@@ -510,6 +510,11 @@ PrefabInstance:
       propertyPath: m_Materials.Array.data[1]
       value: 
       objectReference: {fileID: 2100000, guid: 0e410b84cdc264544a6ba5b805c80fb3, type: 2}
+    - target: {fileID: 7107746757724140747, guid: c63965b45533b034bbced6668bd723e0,
+        type: 3}
+      propertyPath: m_Materials.Array.data[2]
+      value: 
+      objectReference: {fileID: 2100000, guid: e0a3414a16b0adc41846b4583868d30d, type: 2}
     - target: {fileID: 7289053058223276050, guid: c63965b45533b034bbced6668bd723e0,
         type: 3}
       propertyPath: m_Layer


Assets/Script/Venue/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,296 @@
...
+        private int _vocalsIndex;
+        private int _drumIndex;
+        private bool _onPause = false;
+        private StageKitLightingController _controller;
+        public event Action<BeatlineType> HandleBeatline;
+        public event Action<int> HandleDrums;
+        public event Action<LightingType> HandleLighting;
+        public event Action<double> HandleVocals;
+
+        public static StageKitGameplay Instance { get; private set; }
```
EliteAsian123, line 31: Static instance variables on `GameplayBehaviour` types (classes that only exist at play time) are no longer recommended. It's now advised to have an instance of the variable in `GameManager` instead. [See `PracticeManager` as an example](https://github.com/YARC-Official/YARG/blob/19adab6d630a49469aefc991b18eeb1b4bb1b938/Assets/Script/Gameplay/GameManager.cs#L114).

Assets/Script/Venue/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,296 @@
...
+using UnityEngine;
+using YARG.Core;
+using YARG.Core.Chart;
+using YARG.Gameplay;
+using System;
+using Random = UnityEngine.Random;
+
+namespace YARG
+{
+    public class StageKitGameplay :  GameplayBehaviour
```
EliteAsian123, line 11: It's probably a good idea to format these files (there are a couple formatting inconsistencies such as the double space here).

Assets/Script/Venue/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,334 @@
...
+            Fastest
+        }
+
+        public enum FogState
+        {
+            Off = 0,
+            On = 1,
+        }
+
+        public static StageKitLightingController Instance { get; private set; }
```
EliteAsian123, line 165: Assuming this call is present in the persistent scene, this is fine, however, the `MonoSingleton` class should be used instead. [See `DiscordController` as an example](https://github.com/YARC-Official/YARG/blob/19adab6d630a49469aefc991b18eeb1b4bb1b938/Assets/Script/Integration/DiscordController.cs#L11).

Assets/Script/Venue/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,334 @@
...
+        {
+            _isSendingCommands = true;
+            var things = CurrentLightingCue;
+            while (_commandQueue.Count > 0)
+            {
+                var curCommand = _commandQueue.Dequeue();
+
+                switch (curCommand.Item1)
+                {
+                    case 0:
```
EliteAsian123, line 237: Consider using an enum here instead of just using 0, 1, 2, 3, 4... etc.

Assets/Script/Venue/StageKit/StageKitLighting.Parents.cs
```diff
@@ -0,0 +1,105 @@
...
+            StageKitGameplay.Instance.HandleBeatline += HandleBeatlineEvent;
+            StageKitGameplay.Instance.HandleDrums += HandleDrumEvent;
+            StageKitGameplay.Instance.HandleLighting += HandleLightingEvent;
+            StageKitGameplay.Instance.HandleVocals += HandleVocalEvent;
+        }
+
+        public void Dispose()
+        {
+            CancellationTokenSource?.Cancel();
+            StageKitGameplay.Instance.gameManger.BeatEventManager.Unsubscribe(OnBeat);
```
RileyTheFox, line 72: GameManager should almost never be a publicly accessible instance to another class.

Assets/Script/Venue/StageKit/StageKitGameplay.cs
RileyTheFox: Public variables should be named using PascalCase. For example largeVenue should be LargeVenue. I also recommend sorting fields the fields in a class so that public fields are above private fields. Easier to find what you're looking for if you know the accessor (which is denoted by the underscore in a _privateField or such as PublicField)

Assets/Script/Venue/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,334 @@
...
+            On = 1,
+        }
+
+        public static StageKitLightingController Instance { get; private set; }
+
+        public FogState currentFogState = FogState.Off;
+        public FogState previousFogState = FogState.Off;
+
+        public StrobeSpeed currentStrobeState = StrobeSpeed.Off;
+        public StrobeSpeed previousStrobeState = StrobeSpeed.Off;
```
RileyTheFox, line 171: All these fields should be PascalCase

Assets/Script/Venue/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,334 @@
...
+
+        public List<IStageKitHaptics> StageKits = new();
+
+        public StageKitLightingCues CurrentLightingCue = null;
+
+        // Stuff for the actual command sending to the unit
+        private bool _isSendingCommands;
+        private readonly Queue<(int, byte)> _commandQueue = new();
```
RileyTheFox, line 179: I suggest naming the tuple variables here so its easy to understand what they do.

You can do Queue<(int anInt, byte aByte)>, and then the tuple will have named fields.

Assets/Script/Venue/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,334 @@
...
+
+        private async UniTask SendCommands()
+        {
+            _isSendingCommands = true;
+            var things = CurrentLightingCue;
+            while (_commandQueue.Count > 0)
+            {
+                var curCommand = _commandQueue.Dequeue();
+
+                switch (curCommand.Item1)
```
RileyTheFox, line 235: Refer to my comment regarding the _commandQueue tuple, naming the fields of the tuple will make it more descriptive than `Item1` and `Item2`.

RileyTheFox (changes requested):

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,340 @@
...
+            _drumIndex = 0;
+
+            _controller.CurrentLightingCue = null;
+            StageKitLightingController.Instance.StageKits.ForEach(kit => kit.ResetHaptics());
+        }
+
+        private void Update()
+        {
+            //On Pause, turn off the fog and strobe so people don't die, but leave the leds on, looks nice.
+            //Is there a OnPause/OnResume event? that would simplify this.
```
EliteAsian123, line 52: No, but you can just modify the pause method to call a method within this class. If that's too much work though, leaving it like this is fine.

Assets/Script/Integration/StageKit/StageKitScore.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            KillCue();
+        }
+
+        private void OnApplicationQuit()
+        {
+            KillCue();
+        }
+
+        private void KillCue() //The only cue used on the score screen is timed, no need to have all the action and token stuff here.
```
EliteAsian123, line 26: It may be beneficial to make this a method of `_controller`.

EliteAsian123 (commented):
I made some quick comments below; no action necessary. I'll make a proper review when I'm home (yes I did this on mobile).

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,340 @@
+using System.Collections.Generic;
+using UnityEngine;
+using YARG.Core;
+using YARG.Core.Chart;
+using YARG.Gameplay;
+using System;
+using Random = UnityEngine.Random;
+
+namespace YARG
```
EliteAsian123, line 9: The namespace should match the folder: `YARG.Integration.StageKit`

Assets/Script/Integration/StageKit/StageKitLighting.Primitives.cs
```diff
@@ -0,0 +1,167 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using Cysharp.Threading.Tasks;
+using YARG.Core.Chart;
+using YARG.Gameplay;
+using Object = UnityEngine.Object;
+
+namespace YARG {
+	internal class BeatPattern : StageKitLighting
```
EliteAsian123, line 10: Since everything is in the same assembly, `internal` has no effect compared to `public` here, and can probably be replaced with `public`.

Assets/Script/Integration/StageKit/StageKitLighting.Cues.cs
```diff
@@ -0,0 +1,746 @@
...
+namespace YARG {
+    internal class BigRockEnding : StageKitLightingCue
+    {
+        public BigRockEnding()
+        {
+            StageKitLightingController.Instance.SetLed(RED, ALL);
+            StageKitLightingController.Instance.SetLed(BLUE, ALL);
+            StageKitLightingController.Instance.SetLed(GREEN, ALL);
+            StageKitLightingController.Instance.SetLed(YELLOW, ALL);
+            var patternList1 = new List<(int, byte)>
```
EliteAsian123, line 15: Since these are (I believe) never added to after their creation, you can convert these into arrays instead of lists. You can then optimize it by only creating the array once and providing the same array:
```cs
private static readonly (int, byte)[] PATTERN_1 =
{
  (RED, ALL),
  (RED, NONE),
  (RED, NONE),
  (RED, NONE),
};

public BigRockEnding()
{
  // ...

  CuePrimitives.Add(new BeatPattern(PATTERN_1, true, 8.0f));

  // ...
}
```
This can be repeated for all cues.

Assets/Script/Integration/StageKit/StageKitLighting.Cues.cs
```diff
@@ -0,0 +1,746 @@
...
+                (YELLOW, FIVE),
+                (YELLOW, FOUR),
+                (YELLOW, THREE),
+            };
+
+            CuePrimitives.Add( new BeatPattern(patternList1));
+            CuePrimitives.Add( new BeatPattern(patternList2, true, 0.5f));
+        }
+    }
+    internal class LoopCool : StageKitLightingCue
```
EliteAsian123, line 81: (Formatting) put empty lines between classes

Assets/Script/Integration/StageKit/StageKitLighting.Cues.cs
```diff
@@ -0,0 +1,746 @@
...
+                (YELLOW, FOUR),
+                (YELLOW, THREE),
+            };
+
+            CuePrimitives.Add(new BeatPattern(patternList2, true, 0.5f));
+            // I thought it listens to the next but it doesn't seem to. I'll save this for funky fresh mode
+            //new ListenPattern(new List<(int, byte)>(), StageKitLightingPrimitives.ListenTypes.Next);
+        }
+	}
+	internal class ManualCool : StageKitLightingCue{
```
EliteAsian123, line 503: Formatting (`{`)

Assets/Script/Integration/StageKit/StageKitLighting.Cues.cs
```diff
@@ -0,0 +1,746 @@
...
+            {
+				(BLUE, ZERO),
+				(BLUE, ONE),
+				(BLUE, TWO),
+				(BLUE, THREE),
+				(BLUE, FOUR),
+				(BLUE, FIVE),
+				(BLUE, SIX),
+				(BLUE, SEVEN),
+			} ;
```
EliteAsian123, line 697: Formatting (space before `;`). It's probably a good idea to just let Rider format this file.

EliteAsian123 (changes requested):
Hey, I did a few more comments. It's probably a good idea to just auto-format all of these files, and disregard all of the formatting reviews. After that, it should be good! Thanks!

Also, your branch is outdated (from both YARG and YARG.Core). It's probably a good idea to update them, so I can merge without conflicts!

PR #585, opened 2023-10-22
Tweaked implementation of stagekit

TheFatBastid:
Files now formatted by Rider, wow!


Assets/Script/Integration/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,390 @@
...
+
+        public bool LargeVenue; //Doesn't get set until the chart is loaded.
+
+        // Stuff for the actual command sending to the unit
+        private bool _isSendingCommands;
+        private readonly Queue<(int command, byte data)> _commandQueue = new();
+        private byte[] _currentLedState = { 0x00, 0x00, 0x00, 0x00 }; //blue, green, yellow, red
+        private byte[]
+            _previousLedState =
+                { 0x00, 0x00, 0x00, 0x00 }; //this is only for the SendCommands() command to limit swamping the kit.
```
EliteAsian123, line 212: Try to avoid comments on the end of lines because you'll get some weird wrapping like this.

Assets/Script/Integration/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,390 @@
...
+        private const float
+            SEND_DELAY =
+                0.001f; //necessary to prevent the stage kit from getting overwhelmed and dropping commands. In seconds. 0.001 is the minimum. Preliminary testing indicated that 7ms was needed to prevent dropped commands, but it seems that most songs are slow enough to allow 1ms.
```
EliteAsian123, line 215: Same here

Assets/Script/Integration/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,390 @@
...
+
+        private void
+            OnDeviceChange(InputDevice device,
+                InputDeviceChange change) // Listen for new stage kits being added or removed at any time.
```
EliteAsian123, line 219: And here

Assets/Script/Integration/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,390 @@
...
+
+                    default:
+                        Debug.LogWarning("Unknown Stagekit command: " + curCommand.command);
+                        break;
+                }
+
+                if (
+                    things != CurrentLightingCue &&
+                    _commandQueue.Count >
+                    0.05f / SEND_DELAY) //If there is more 1/20th of a second in commands left in the queue when the cue changes, clear it. Really fast songs can build up a queue in the thousands while in BRE or Frenzy. 1/20th of a second is said to be the blink of an eye.
```
EliteAsian123, line 315: And here. Should be like this instead. Try to keep characters on the left of the gray line:
image

Assets/Script/Integration/StageKit/StageKitLightingController.cs
```diff
@@ -0,0 +1,390 @@
...
+                                large venue
+                                2 yellow at 180 to each other clock 2 second, 2 red at 180 counter clock 1 second
+
+        RE Menu lighting        1 blue@0 rotates counter clock every two seconds. Made by me. Not in the original game.
+ */
+namespace YARG.Integration.StageKit
+{
+    public class StageKitLightingController : MonoSingleton<StageKitLightingController>
+    {
+        public enum StrobeSpeed
```
EliteAsian123, line 147: Can this just be removed and replaced with the `StageKitStrobeSpeed` type instead? It would simplify the `SetStrobeSpeed` method for example.

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,338 @@
...
+using Random = UnityEngine.Random;
+
+namespace YARG.Integration.StageKit
+{
+    public class StageKitGameplay : GameplayBehaviour
+    {
+        public event Action<BeatlineType> HandleBeatline;
+        public event Action<int> HandleDrums;
+        public event Action<LightingType> HandleLighting;
+        public event Action<double> HandleVocals;
```
EliteAsian123, line 16: Making these events is very inefficient, as you're constantly doing `+=` and `-=` on them (all yellow lines are new allocations).
image

Instead, you should make these methods. See comment below as to how.

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,338 @@
...
+
+        private void SetupCue(StageKitLightingCue cue)
+        {
+            _controller.CurrentLightingCue = cue;
+            _controller.CurrentLightingCue.LargeVenue = _controller.LargeVenue;
+
+            HandleBeatline += _controller.CurrentLightingCue.HandleBeatlineEvent;
+            HandleDrums += _controller.CurrentLightingCue.HandleDrumEvent;
+            HandleLighting += _controller.CurrentLightingCue.HandleLightingEvent;
+            HandleVocals += _controller.CurrentLightingCue.HandleVocalEvent;
```
EliteAsian123, line 286: This can be easily converted to, for example:
```cs
private void HandleBeatline() {
  if (_controller.CurrentLightingCue == null) return;

  _controller.CurrentLightingCue.HandleBeatlineEvent();

  foreach (var primitive in _controller.CurrentLightingCue.CuePrimitives) {
    primitive.HandleBeatlineEvent();
  }
}
```
or something since the `CurrentLightingCue` is set above.

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,338 @@
...
+                HandleBeatline += primitive.HandleBeatlineEvent;
+                HandleDrums += primitive.HandleDrumEvent;
+                HandleLighting += primitive.HandleLightingEvent;
+                HandleVocals += primitive.HandleVocalEvent;
+            }
+        }
+
+        private void KillPrimitive(StageKitLighting primitive)
+        {
+            GameManager.BeatEventManager.Unsubscribe(primitive.OnBeat);
```
EliteAsian123, line 299: The `KillPrimitive` method can be removed after the refactor above, however, when is the `primitive.OnBeat` subscribed? It isn't above. Unless it is in the constructor? If so, a comment should be put here to state that.

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,338 @@
...
+            if (_controller.CurrentLightingCue == null)
+            {
+                return;
+            }
+
+            HandleBeatline -= _controller.CurrentLightingCue.HandleBeatlineEvent;
+            HandleDrums -= _controller.CurrentLightingCue.HandleDrumEvent;
+            HandleLighting -= _controller.CurrentLightingCue.HandleLightingEvent;
+            HandleVocals -= _controller.CurrentLightingCue.HandleVocalEvent;
+
```
EliteAsian123, line 320: After the refactor, all of this event stuff can be removed.

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -0,0 +1,338 @@
...
+            _eventIndex = 0;
+            _drumIndex = 0;
+
+            _controller.CurrentLightingCue = null;
+            StageKitLightingController.Instance.StageKits.ForEach(kit => kit.ResetHaptics());
+        }
+
+        private void Update()
+        {
+            //On Pause, turn off the fog and strobe so people don't die, but leave the leds on, looks nice.
```
EliteAsian123, line 51: This whole update method is large, which is fine, but is running in whole if there's no stage kits attached. It may be a good idea to skip it all if there's nothing attached.

Assets/Scenes/PersistentScene.unity
EliteAsian123, line 1: The "Stage Kit Lighting Controller" component should be moved to the "Integrations" object to make things more consistent.

EliteAsian123 (changes requested):
Hello, last review, I promise. After all of these things are addressed, we should be good to go! We just gotta wait for #584 to be merged, and this PR updated for that, and I can merge this ASAP in time for the next alpha. Thanks again for your hard work!

Assets/Script/Integration/StageKit/StageKitLightingController.cs
TheFatBastid: TIL of the gray line

Assets/Script/Integration/StageKit/StageKitLightingController.cs
TheFatBastid: Can't remember why I did that in the old code. Not needed and removed.

TheFatBastid:
Okay changes made. Waiting for #584. Review and comments always welcome!

EliteAsian123 (approved):

PR #586, opened 2023-10-24
Improve handling of song-specific video backgrounds

TheNathannator:
Was very satisfying to load up TWO-TORIAL and have its video background start right on the drop 40 seconds in lol

Notable changes:

- The struct returned by the venue loader has been refactored to both have a better name and indicate where a venue was sourced from, as we don't want to apply video start/end times to a global video.

- BackgroundManager now only `Update()`s when a video is being played, and its update loop has been refactored to implement video start/end time. When it no longer needs to update, it disables itself.

- Song videos will loop if they are notably shorter than the song itself. The threshold is 85% of the song's length: below that it will loop, equal to or above it won't. All other videos will always loop.

- Song videos now follow the song speed and are seeked when setting song position. Music videos and modcharts will now match up in practice mode.

### DIFF
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -15,6 +15,8 @@ namespace YARG.Gameplay
         [SerializeField]
         private RawImage _backgroundImage;
 
+        private VenueInfo _venueInfo;
+
         private bool _videoShouldBeStarted;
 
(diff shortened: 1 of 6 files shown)


TheNathannator:
TODO: implement seeking of song-specific video backgrounds, waiting for #584 to be merged first

TheNathannator:
All good to go now!

PR #588, opened 2023-11-03
Auto-generate venue lighting incase it doesn't exist

raphaelgoulart:
This code reads a JSON file containing info on which lighting presets/post-procs to use for which sections, and generates a lighting track based on that info, if the song currently being loaded doesn't have it.
Implementation loosely based on MagmaC3's, detailed in http://docs.c3universe.com/rbndocs/index.php?title=Venue_Autogeneration
This needs implementation on the YARG side as well, but the game should still behave the same without it (i.e. songs loading normally, without venue autogeneration).

Example auto-generation preset JSON file, for testing purposes: [DefaultPreset.json](https://github.com/YARC-Official/YARG.Core/files/13253806/DefaultPreset.json)
Demo video: https://www.youtube.com/watch?v=jd588FNhSOE (for any `[next]` counter that resets to 0 for no reason -- that's the blend-in lighting event, not a `[first]` keyframe)
The demo video is a bit outdated, as it lacks spotlight events for solos.

TODO:
- ~~Figure out better way to organize the lookup dictionaries~~ Pending https://github.com/YARC-Official/YARG.Core/pull/109;
- Implement selection in Settings screen. Implementation suggestion:
  - Have built-in presets in some game folder, like `StreamingAssets/autogen` or similar;
  - Read custom presets from `PersistentDataPath/autogen` or similar;
  - List them in a dropdown in the settings, either near to other venue-related settings, or on the Presets tab;
  - Pass the selected preset as a parameter for the Song.LoadChart() function.
- (Very low priority) Add singalong events based on HARM2 phrases if any (add to the first two performers that are available in this order: guitar/bass/keys/drums)


Assets/Script/Venue/Lights/VenueAutogenerationPreset.cs
```diff
@@ -0,0 +1,383 @@
...
+                   lighting == LightingType.Chorus ||
+                   lighting == LightingType.Cool_Manual ||
+                   lighting == LightingType.Stomp ||
+                   lighting == LightingType.Verse ||
+                   lighting == LightingType.Warm_Manual;
+        }
+
+        private AutogenerationSectionPreset JObjectToSectionPreset(JObject o)
+        {
+            AutogenerationSectionPreset sectionPreset = new AutogenerationSectionPreset();
```
EliteAsian123, line 252: Instead of doing all of this, I'd highly recommend serializing into the `AutogenerationSectionPreset` instead using `JsonConvert.DeserializeObject<AutogenerationSectionPreset>(<text>)` instead. In fact, you can make it *even easier* on yourself by making a class like this:
```cs
public class AutogenerationSectionFile {
   public string CameraPacing;
   public Dictionary<string, AutogenerationSectionPreset> SectionPresets;
}
```
and doing this on `ReadPresetFromFile`:
```cs
JsonConvert.DeserializeObject<AutogenerationSectionFile>(File.ReadAllText(path))
```

This has several pros:
* Type checking on everything automatically
* No need for string constants and stuff everywhere (bad practice)
* Code that is a lot easier to understand
* And also the `.json` fields will be in spec with YARG's preferred `.json` field naming (PascalCase)

Any value checking can be done after the fact by looking at the values of the output file.

Assets/Script/Venue/Lights/VenueAutogenerationPreset.cs
```diff
@@ -0,0 +1,383 @@
...
+    /// Possible camera pacing values.
+    /// </summary>
+    public enum CameraPacing
+    {
+        Minimal,
+        Slow,
+        Medium,
+        Fast,
+        Crazy
+    }
```
EliteAsian123, line 382: Are these only used within the `VenueAutogenerationPreset` class? If so, they should be private and at the top and within the said class.

Assets/StreamingAssets/autogen/DefaultPreset.json
EliteAsian123, line 1: Placing default presets for this in StreamingAssets is fine, however other users putting things in here could be problematic because of how the YARC Launcher works (files here will be deleted).

EliteAsian123 (changes requested):
Hi thanks for the PR! Mostly everything is pretty good, however there are a few things with the serialization I recommended.

As for styling, if you could, please separate blocks of code with empty lines. For example, the `GenerateLightingEvents` is just one big block of code and is kinda hard to look through. Thanks!

Assets/StreamingAssets/autogen/DefaultPreset.json
raphaelgoulart: Yeah, these are meant to be only default presets, customs should be placed elsewhere when I get to implementing it in settings

Assets/Script/Gameplay/HUD/LyricBar.cs
```diff
@@ -137,7 +137,7 @@ private string GetDisplayTextWithSpace(string lyricText)
             // Special replacements
             lyricText = lyricText.Replace('=', '-');
             lyricText = lyricText.Replace('_', ' ');
-            lyricText = lyricText.Replace('§', '‿');
+            lyricText = lyricText.Replace('§', ' ');
```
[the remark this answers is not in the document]
raphaelgoulart: No, but it fixes a bug with the synalepha character showing as an unknown character. Fixed it on my end, guess I forgot to remove this from the commit.
If you want I can undo it

Assets/Script/Venue/Lights/VenueAutogenerationPreset.cs
raphaelgoulart: Thanks for the advice! I have zero familiarity with Newtonsoft.JSON so I knew my implementation had lots of room for improvement (to say the least lmao), this'll make my life easier for sure

Assets/Script/Venue/Lights/VenueAutogenerationPreset.cs
```diff
@@ -0,0 +1,383 @@
...
+        private bool LightingIsManual(LightingType lighting)
+        {
+            return lighting == LightingType.Default ||
+                   lighting == LightingType.Dischord ||
+                   lighting == LightingType.Chorus ||
+                   lighting == LightingType.Cool_Manual ||
+                   lighting == LightingType.Stomp ||
+                   lighting == LightingType.Verse ||
+                   lighting == LightingType.Warm_Manual;
+        }
```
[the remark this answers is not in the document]
raphaelgoulart: Fixed in https://github.com/YARC-Official/YARG/pull/588/commits/c1e8058104ce2f0a7274b44b99dbe8e80091c8ca

Assets/Script/Venue/Lights/VenueAutogenerationPreset.cs
raphaelgoulart: Fixed in https://github.com/YARC-Official/YARG/pull/588/commits/c1e8058104ce2f0a7274b44b99dbe8e80091c8ca and https://github.com/YARC-Official/YARG/pull/588/commits/8e1200bb560f0db6652dda52b86f430738354b35

Assets/Script/Gameplay/HUD/LyricBar.cs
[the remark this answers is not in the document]
EliteAsian123: Yeah, if you can undo it that'll be great (since it doesn't have anything to do with the PR)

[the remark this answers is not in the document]
raphaelgoulart: Fixed in https://github.com/YARC-Official/YARG/pull/588/commits/7661dd5b0b5213e7f63bd02c7da6e170750b28f0

Assets/Script/Venue/Lights/VenueAutogenerationPreset.cs
raphaelgoulart: Fixed in https://github.com/YARC-Official/YARG/pull/588/commits/7661dd5b0b5213e7f63bd02c7da6e170750b28f0

EliteAsian123:
Merged in 3d62b5f55ae46cc283241ae1f1308dd4d5c79516 (PR did not close for some reason)

PR #603, opened 2023-11-23
Moved loading phrase from song select to difficulty screen

TheFatBastid:
It has been a long time coming but since things have settled down in terms of design, it seemed the right time to move the loading phrase from the cramped song select help bar location to the wide open difficulty screen


Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
```diff
@@ -97,6 +101,8 @@ private void OnEnable()
             // ChangePlayer(0) will update for the current player
             _playerIndex = 0;
             ChangePlayer(0);
+
+            SetInfoText(RichTextUtils.StripRichTextTagsExclude(GlobalVariables.Instance.CurrentSong.LoadingPhrase, RichTextUtils.GOOD_TAGS));
```
RileyTheFox, line 105: Would prefer to use `StripRichTextTags(phrase, RichTextUtils.BAD_TAGS)` over the exclude function with the `GOOD_TAGS` constant. Just a little more clearer on what is being removed versus it being a double negative situation (stripping all except good tags vs stripping bad tags)

Not sure why it was written this way originally in the first place.

I think extracting the stripped text into a variable before passing it in would be a little easier to read as well (`string stripped = RichTextUtils.strip etc...` a line above). Also the `SetInfoText` function is redundant as setting the loading phrase text is only done once. Probably best to remove it.

Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
```diff
@@ -499,5 +505,10 @@ public void SongSpeedEndEdit(string text)
             speed = Math.Clamp(speed, 10, 4995);
             _speedInput.SetTextWithoutNotify($"{speed}%");
         }
+
+        public void SetInfoText(string str)
```
RileyTheFox, line 509: I don't think the function is necessary as the loading phrase only ever gets set from one place in `OnEnable`. Would be fine to just set the `_loadingPhrase.text` field directly.

If for some reason the loading phrase were to be set from elsewhere in the future then a function would make sense.

RileyTheFox (changes requested):

Assets/Prefabs/Menu/DifficultySelect/DifficultySelectMenu.prefab
EliteAsian123, line 1: image

Anchor points are not set correctly so it does not align on different resolutions.

It's also recommended that you don't adjust these margins unless you absolutely have to:
image
Instead, choose one of the stretch options:
image

EliteAsian123 (changes requested):

EliteAsian123:
Merged in 65171ab99e8533f62164293f10ff7fbd45b0818b (had to do it manually)

PR #605, opened 2023-11-27
Refactor how gameplay behaviors are handled

TheNathannator:
`GameplayBehaviour`s are now managed directly by `GameManager`, which allows for a number of improvements:
- Initialization methods can now be async where needed, and will be awaited properly while `GameManager` loads.
  - The `SongLoaded`/`ChartLoaded`/etc. events have been removed, as they are no longer necessary after this change.
- Everything is now consistently updated after `GameManager` is, which ensures nothing is potentially a frame behind.
- New hooks for seeking, speed changes, and pausing have been added, which means objects that need these don't have to rely on `GameManager` calling them directly, and don't have to constantly check each update to poll their status.

This fixes the song starting despite loading not actually being finished yet, and should make the system a bit more flexible.

The player classes have also been reworked to *not* use `GameplayBehaviour` (alongside other cleanups), as they are already handled specially by `GameManager`. It also prevents them from being updated twice, since they won't be part of the list of GameplayBehaviour`s.

### DIFF
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -1,26 +1,24 @@
 ﻿using System;
-using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using Cysharp.Threading.Tasks;
 using UnityEngine;
 using UnityEngine.UI;
 using UnityEngine.Video;
-using YARG.Core.Extensions;
 using YARG.Core.Venue;
-using YARG.Helpers;
 using YARG.Venue;
 
 namespace YARG.Gameplay
 {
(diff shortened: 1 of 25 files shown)


TheNathannator:
Needs testing on practice and replays, forgot to before heading off for the day.

TheNathannator:
Drafted while I figure out a fix for an initialization exception caused by the theme manager creating an instance of the note prefab without initializing it in the way it expects.

Also want to clean up the commits a bit, some of them aren't entirely related to this PR. I just kinda went on a spree and did a bunch of fixes lol

EliteAsian123:
I'm going to close this for now as it's like 7 months out of date lol

PR #609, opened 2023-12-02
Added separate tracks for each harmony.

JSanchez78657:
Added separate tracks for each harmony.
Extended vocal track elements and moved practice mode elements to fit the larger vocal track. 
Removed desync of harmony lyrics since they can now be displayed with their correct timing.


Assets/Art/Textures/Gameplay/Track/VocalTrackHarmony.png
kaduwaengertner, line 1: I'm going to redo this image so it matches the style, the extra lyric addition have a different contrast, I'll make the change and I'll send to Elite to add

Assets/Art/Textures/Gameplay/Track/VocalTrackNormal.png
kaduwaengertner, line 1: Same for this, will edit the file to keep the style used, some of the effects got removed in the edit

Assets/Script/Gameplay/Player/Vocals/VocalLyricContainer.cs
```diff
@@ -25,13 +25,6 @@ public bool TrySpawnLyric(TextEvent lyric, VocalNote probableNotePair, bool isSt
             int i = harmIndex;
```
EliteAsian123, line 25: From your changes, `i` is now a useless variable and can be removed. There is one use case below which is incorrect.

Assets/Script/Gameplay/Player/Vocals/VocalLyricContainer.cs
```diff
@@ -25,13 +25,6 @@ public bool TrySpawnLyric(TextEvent lyric, VocalNote probableNotePair, bool isSt
...
 
-            // Never show HARM3's lyrics unless they're different from HARM2's lyric
-            if (harmIndex == 2 && _lastSecondHarmonyLyric == lyric.Text)
-            {
-                _lastSecondHarmonyLyric = null;
-                return true;
-            }
-
             // Skip this frame if the pool is full
             if (!_pools[i].CanSpawnAmount(1))
```
EliteAsian123, line 29: Since `harmIndex` is being used to actually spawn in the lyrics, and `i` is not equal to `harmIndex` when on harmony 3, this could cause errors. Replacing `i` with `harmIndex` would fix this.

EliteAsian123 (changes requested):
Hi, thanks so much for your PR! I did a quick code review, so if you could fix those issues, that'll be great. I'll also do a review on the scene changes and that when I am able to. Thanks!

Assets/Script/Gameplay/Player/Vocals/VocalLyricContainer.cs
JSanchez78657: Replaced i with harmIndex

kaduwaengertner:
Sent the new files on Discord DM, make sure to adjust the tracks on Unity, noticed you cropped the top border on the edit you did, new ones have the right size on both border so might have to adjust a bit

PR #615, opened 2023-12-09
Refactor dropdown settings for cleanup/extensibility

TheNathannator:
This allows the resolution setting to be handled as just another dropdown setting, rather than having to specially handle it in the settings menu. It should also make it easier to create additional specialized dropdowns in the future.


jgoyvaerts:
I was looking for a good way to refresh the TrackPreview on the right side when switching resolution/full screen mode (just like the preset tab does) because when switching from a really wide resolution to a normal resolution the trackpreview is squished if you don't refresh it.

However, I couldn't find a good place/method for refreshing it, could you add that in this PR as well?

EliteAsian123:
> However, I couldn't find a good place/method for refreshing it, could you add that in this PR as well?

I already did some reworking of that system in the `engine-presets` branch, so I'll do it there while I'm at it!

PR #616, opened 2023-12-09
Improve resolution dropdown by sorting from high to low

jgoyvaerts:


EliteAsian123:
Since the dropdown was changed in #615, I committed 1fefdbe1d4979b8042bfad664791753e093f3d87 which does the same thing for the new system. Thanks for your PR!

PR #621, opened 2023-12-18
Add setting to disable StarPower colored notes

jgoyvaerts:


EliteAsian123:
This should be a color profile option, so it's per player. Technically you can already achieve this as well with color profiles, by setting the starpower colors as the same.

PR #624, opened 2023-12-22
Clean up navigation group selection handling and fix various issues

TheNathannator:
Navigatables are now selected using an index instead of a direct reference, which helps to ensure various things about how selections are handled. Deselections are also now reported via the selection changed event, and more error checking/reporting has been added to detect issues such as out-of-range indexes and duplicate entries.

In addition, this PR fixes the song path manager throwing an out-of-range exception when refreshing the settings menu, as well as the unintentional wraparound that would happen in the difficulty menu when scrolling down on Ready.

### DIFF
--- a/Assets/Prefabs/Menu/DifficultySelect/DifficultySelectMenu.prefab
+++ b/Assets/Prefabs/Menu/DifficultySelect/DifficultySelectMenu.prefab
@@ -1809,8 +1809,8 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   _defaultGroup: 1
   _canBeCurrent: 1
-  _addAllChildrenOnAwake: 1
-  _selectFirst: 1
+  _addAllChildrenOnAwake: 0
+  _selectFirst: 0
 --- !u!224 &7507340441076923383 stripped
 RectTransform:
   m_CorrespondingSourceObject: {fileID: 375337769934742335, guid: 81863a63e0d96cf459c263ee0ea6f98e,
--- a/Assets/Script/Helpers/LocaleHelper.cs
+++ b/Assets/Script/Helpers/LocaleHelper.cs
@@ -8,6 +8,8 @@ namespace YARG.Helpers
 {
     public static class LocaleHelper
     {
+        public static readonly LocalizedString EmptyString = new();
+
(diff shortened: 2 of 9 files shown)


jgoyvaerts:
Some thoughts: 
- Using selectedIndex with default value -1 isn't very clean, we have nullables, why not use them? 
- This still has a method SetSelectedFromNavigatable, which is called from child to parent. Ideally this should be an event listener or something equivalent, to get rid of the circular reference. 
- The NavigatableBehaviour SelectedBehaviour currently modifies SelectedIndex and validates it. A getter shouldn't modify other properties/variables, and the validation shouldn't happen in this location, but in the place where selectedindex is originally modified

TheNathannator:
- IMO nullables don't really add much benefit here aside from making intention more clear, ultimately it just results in another safety check that has to be made, as negative values still need to be checked for.
- That would probably be better, yeah, it just wasn't part of my focus as that wasn't what was causing exceptions lol
- The clamp is largely just a safety mechanism to help prevent things from getting stuck in invalid states, validation is already applied elsewhere and the case where the clamp actually does something should never get hit.

jgoyvaerts:
> * IMO nullables don't really add much benefit here aside from making intention more clear, ultimately it just results in another safety check that has to be made, as negative values still need to be checked for.

Negative values are not checked. Check the method SelectAt method, it doesn't check if SelectedIndex is negative. 

Code that should never get hit shouldn't be written, it can hide bugs if it actually does get hit.

Using -1 as a magic value to indicate a missing value is outdated and not as readable or maintainable as a nullable value. Seeing as this is an open source project, those things should also be taken into account.

TheNathannator:
It doesn't check for negative to allow passing in negative for deselection, e.g. in `SettingsMenu.RefreshAndKeepPosition` where it has to store the current index and restore it after refreshing. If nothing was selected beforehand (i.e. the selected index were -1), it would end up throwing and breaking stuff.

image

The negative check is elided into the null checks on `SelectedBehaviour`, since that will return null if the value is negative.

I suppose this stuff isn't very obvious at first glance lol, but using a nullable value instead wouldn't change that much. But I'm also probably overthinking things and making it way deeper than it should be.

PR #625, opened 2023-12-23
Main menu colored text on hover.

Insomnautika:
Added text color to main menu on hover using the NavigationTextColorizer script and changed selection highlight PNG to white so that the color can be selected in the unity editor allowing for multiple colored items.

First PR so not sure if I did the commit correctly.

[PREVIEW]
image
image
image


EliteAsian123:
Hi, thanks for the PR! Since this is an art change, I'm going to need @kaduwaengertner's opinion on it.

As for the PR, you did everything correctly, however I highly recommend making a new branch for each PR. That way, you can work on other things while this one is being reviewed. 

Thanks!

kaduwaengertner:
I don't think we need hover effects on it, cause the left gradient detail + yellow coloring is already the hover effect for the main menu, don't think having a hover effect for a selection that is already "active" makes sense talking about UI/UX, I do like the Exit having a red coloring, just not a fan of having the blue options, just the regular yellow we use its enough for this case

TheNathannator:
@Insomnautika You still up for this PR? It seems there was some miscommunication on what exactly Kadu was referring to, he thought the colored text was already a thing lol

The only issue was the blue coloring, I'm not the best at summarizing the intent so I'll leave that to him to clarify.

Very sorry about this situation, we feel kinda bad lol

PR #626, opened 2023-12-24
Added support for some Santroller LEDs

TheFatBastid:
Santroller devices have LEDs on them. This will send commands to them for star power fill, star power activate, multiplier change, and solo start/end


Assets/Script/Gameplay/Player/BasePlayer.cs
```diff
@@ -244,6 +274,15 @@ protected virtual void OnStarPowerStatus(bool status)
                     ? SfxSample.StarPowerDeploy
                     : SfxSample.StarPowerRelease);
             }
+
+            if (status)
```
RileyTheFox, line 278: This if statement can be removed and just pass the status variable in like `kit.SetStarPowerActive(status);`

RileyTheFox (changes requested):
Overall the PR is good besides a couple things I'd change which I've outlined in some comments.

It's also worth thinking if this sort of thing could be abstracted into a dedicated player haptics class, which handles adding and removing devices as well as calling device specific code (for example santroller, or things like razer, logitech etc)

Let me know if you think that would be something we should do.

Assets/Script/Gameplay/Player/BasePlayer.cs
```diff
@@ -199,11 +205,29 @@ protected virtual void UpdateInputs(double time)
         private void SubscribeToInputEvents()
         {
             Player.Bindings.SubscribeToGameplayInputs(Player.Profile.GameMode, OnGameInput);
+            Player.Bindings.DeviceAdded += OnDeviceAdded;
```
EliteAsian123, line 208: In my opinion, subscribing to these events here is not the best. Instead, I think it would be better if the bindings contained a list of `ISantrollerHaptics` instead, which the player class can then access and call to. This will be especially useful in the future if we plan to add customization to this, so it can be configured in the binds menu or something of the like.

EliteAsian123 (changes requested):

Assets/Script/Gameplay/Player/BasePlayer.cs
```diff
@@ -244,6 +274,15 @@ protected virtual void OnStarPowerStatus(bool status)
                     ? SfxSample.StarPowerDeploy
                     : SfxSample.StarPowerRelease);
             }
+
+            if (status)
+            {
+                SantrollerHaptics.ForEach(kit => kit.SetStarPowerActive(true));
```
TheNathannator, line 280: I'd generally recommend not using the `ForEach` method on lists, it's just as easy to do a regular `foreach` loop and it avoids the delegate allocations that happen as part of calling `ForEach`. Here it's not super important though, as these aren't going to be called often enough for it to matter in the long run.

Assets/Script/Gameplay/Player/BasePlayer.cs
TheNathannator: For now I think this approach works fine, it would probably be better to abstract things out once we know what we want to do for customization, other devices, etc. The only real customization options I can think of would be to enable/disable specific commands though, which could also be configured in the firmware itself instead of in-game.

Just want to make sure we avoid unnecessary complexity that we don't actually end up benefitting from.

TheFatBastid: I didn't change this for now.

EliteAsian123 (approved):

PR #627, opened 2023-12-24
Add "New Highscore" text notification while playing

jgoyvaerts:


Assets/Script/Gameplay/HUD/TextNotifications.cs
```diff
@@ -13,8 +14,8 @@ public class TextNotifications : MonoBehaviour
         private int _streak;
         private int _nextStreakCount;
 
-        private string _nextNotification;
-        private bool   _notificationPending;
+        private Queue<string> _notificationQueue = new();
```
RileyTheFox, line 17: Using a queue was discussed before and it was dismissed because it causes notifications to get severely backed up in some situations, here's an example:
image
Note index is 14,000 but the streak text is still at 4000.

This is caused by the high nps but it could also happen with other text such as "Star Power Ready", "High Score", "x Note Streak" all occurring very quickly, causing them to get delayed. The proposed solution was a priority system which would definitely be needed in the case of adding a high score notification (would in my opinion would take highest priority)

To get the same behaviour as of the existing code, a stack can be used instead. Take the latest notification and clear the rest. But it should be reworked into a weighted/prioritised system

Assets/Script/Gameplay/GameManager.Loading.cs
```diff
@@ -306,7 +308,8 @@ private void CreatePlayers()
                     // Setup player
                     var trackPlayer = playerObject.GetComponent<TrackPlayer>();
                     var trackView = _trackViewManager.CreateTrackView(trackPlayer, player);
-                    trackPlayer.Initialize(index, player, Chart, trackView);
+                    var currentHighScore = ScoreContainer.GetHighScore(Song.Hash)?.Score;
+                    trackPlayer.Initialize(index, player, Chart, trackView, currentHighScore);
```
RileyTheFox, line 312: I think instead of passing this to the player, it should be global and displayed somewhere that's for all players, rather than per track. So for example, it could display above/below the score box

RileyTheFox (changes requested):
Good PR, I just don't want the core behaviour of the notification system changing yet until we work on a weighted system. Other than that everything looks alright

Assets/Script/Gameplay/GameManager.Loading.cs
jgoyvaerts: The way I understand it is that the highscore is also per player (look at the query). It's not a band score but an individual score. So the scoring/query system would have to change too then

Assets/Script/Gameplay/HUD/TextNotifications.cs
jgoyvaerts: How about a custom Queue with only one message per type? 

```
_queue.Enqueue(NotificationType.NoteStreak, "100-Note Streak")
```
Would then replace the previously notestreak message if there was still one on the queue?

RileyTheFox: Yeah this is exactly the idea we had in mind! The idea for the weighted system would take the newest version for each type, as well as prioritising certain types over others. For example a note streak notification would have less weight than a new personal best.

I think it was decided that if a notification gets out-classed by another then it is dropped entirely instead of just delaying, as notifications stay for some time so it would be kind of pointless to display something like "100 note streak" when you've already hit another 30 notes because something else was in the way.

Assets/Script/Gameplay/GameManager.Loading.cs
RileyTheFox: Personally I don't know how personal and band scores are handled. All I know is that all of them are saved and that they are searchable per player. Probably needs more discussion as to how it should work as I don't think its been decided.

Assets/Script/Gameplay/HUD/TextNotifications.cs
jgoyvaerts: Made a second pass at it, let me know what you think of this approach

Assets/Script/Gameplay/HUD/TextNotificationQueue.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public void Clear()
+        {
+            _notificationQueue.Clear();
+        }
+
+        public int Count => _notificationQueue.Count;
+    }
+
+    public class TextNotification
```
EliteAsian123, line 60: This can be converted into a readonly struct I believe

EliteAsian123 (changes requested):

Assets/Script/Gameplay/GameManager.Loading.cs
EliteAsian123: They are saved as individual scores, however, in this case I think it'd be best to make a new query that gets the high score for this *specific* part, rather than the highest score of every part.

Assets/Script/Gameplay/HUD/TextNotificationQueue.cs
sonicfind: You can

jgoyvaerts: Fixed

Assets/Script/Gameplay/GameManager.Loading.cs
jgoyvaerts: Score is per instrument now!

RileyTheFox (approved):

EliteAsian123 (approved):

PR #631, opened 2023-12-26
Main menu colored text highlight.

Insomnautika:
Added text color to main menu on selection using the NavigationTextColorizer script and changed selection highlight PNG to white so that the color can be selected in the unity editor.

Updated PR:
- Changed history color back to yellow.
- Updated all colors to follow YARG design standards.

[PREVIEW IMAGES]
image
image
image


EliteAsian123:
My only complaint is that the "New" badge colour does not match the selection color. That can be changed here: 
image

Insomnautika:
Alright. The menu badge is now the same color as the highlight.

PR #632, opened 2023-12-26
Take two: Clean up navigation group selection handling and fix various issues

TheNathannator:
The *one* time I don't check my changes before pushing them, it immediately comes back to bite me lol

Re-do of #624 that should actually work this time, there were a number of issues I didn't come across initially which should now be fixed.


EliteAsian123:
Not sure if you know, but, the “More Options” popup in the music library throws errors.

TheNathannator:
Fixed, also went and made sure everything else was good and added a warning/auto-disable against the thing that causes it.

PR #637, opened 2023-12-27
Implement collective debounce across all controls on a button binding, along with displays for collective value on bindings

TheNathannator:
The new debounce behaves the same as the one for individual controls: when a value change occurs, it will lock to that new state for the configured amount of time, keeping track of any further state changes during that time. Once that time expires, it will switch back to the tracked state, and when the next input comes in it repeats the process.

All binding types also now have displays for their current value, for parity with the individual bindings and to help with figuring out possible input issues.

image


jgoyvaerts:
I would put the debounce treshold to a default value of 1/2/5ms as discussed. 
Discussion starts here: https://discord.com/channels/1086048856678084609/1095149759691493386/1189550037622075392

EliteAsian123:
With this PR, I'm unable to add new bindings

TheNathannator:
Binding addition should be fixed now. As for the shadows, tbh I can hardly even notice it in the editor lol, if the menu background was more transparent I'd be able to see it.

PR #638, opened 2023-12-28
Adds "Album Art as Background" settings

yoitsOJ:
Adds a setting that uses the album artwork and metadata of a chart to create a background venue.

TODO:
- Learn to use the transform function on cover + text so there's not 3 different sets of objects to account for multiplayer situations
- Find/make a blur + darken shader for the background image that's compliant with LGPL (SnapshotShaders will have to be replaced according to EA)
- Make it so enabling DisablePerSongBackground disables AlbumArtBackground and vice-versa

### DIFF
--- a/Assets/Art/Reference/oneplayer.png
+++ b/Assets/Art/Reference/oneplayer.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:452cc0191f400a535843226397c5f1e0097787af68eb5165ef5ddd7cb5fc524d
+size 507476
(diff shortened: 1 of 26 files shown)


yoitsOJ:
Requesting the help of someone else to make a blur + darkness shader or script because I can't understand of this and the Unity documentation just looks like gibberish to me lmao.

Other than that though this is pretty much done!
Features like the blur sliders and a reorderable list rather than the dropdown are very far out of my skillset, and this already took enough hand holding from Nate, EA, and Riley (thanks a ton by the way guys) as it is, so if this does get added then it'd have to get handed off to someone else for those features.

PR #642, opened 2024-01-01
Box Highlight Under Help Bar Buttons when Mouse Hovering

Insomnautika:
- Colored help bar highlights.

[PREVIEW IMAGES]
hover1
hover2
hover3


kaduwaengertner:
I would just remove the box coloring on the background, just keeping the text in yellow for the hover is enough, maybe we can also lower the opacity of the icons on default and change to 100% when hover also

Insomnautika:
The background box coloring is the only point of this pull request. Was trying to make it more clear the button was being hovered with the mouse besides just text color sense there is no change in the button icon when hovering. Makes it feel like you need to click on the text directly instead of it being clear the whole panel is a button. If you don't like this idea you can close the pull request.

EliteAsian123:
I do agree that it does make it more clear of why you are supposed to click, and the fact that you can click it. Perhaps a different box coloring color could help?

kaduwaengertner:
Did some changes on figma, if we can change to something like this it would fit the style, subtle gradient instead of a full solid box
image


Ideally the footer would be a gradient also, dont remember why we ended up using a solid footer but this was the original concept:
image

Insomnautika:
I can make it look like that. Give me a bit to update it.

EliteAsian123:
The transparency gradient on the help bar didn't work because Unity's transparency quality is terrible on UI (thanks Unity for always being a pain for UI stuff, lol). I didn't spend too long on it, but if you could find a way to make it work without tons of banding, that'll be great.

Assets/Script/Menu/Persistent/HelpBarButton.cs
```diff
@@ -1,36 +1,76 @@
...
+
+        [SerializeField]
+        private TextMeshProUGUI _buttonLabel;
+
+        [SerializeField]
+        private TextMeshProUGUI _buttonText;
 
         private NavigationScheme.Entry? _entry;
 
+        private Color32 ButtonBackgroundColor;
```
EliteAsian123, line 26: Private fields should be name `_likeThis`. This field should be named `_buttonBackgroundColor`.
Along with this, this field should be of type `Color` instead of `Color32` (`Color32`s can be implicitly converted into the simpler `Color`).

Assets/Script/Menu/Persistent/HelpBarButton.cs
```diff
@@ -1,36 +1,76 @@
...
 
-            // Set the icon
-            var icons = MenuData.NavigationIcons;
+            _buttonText.color = Color.white;
+
+            // Set the icon color
             _buttonImage.sprite = icons.GetIcon(entry.Action);
-            _buttonImage.color = icons.GetColor(entry.Action);
+            _buttonImage.color = ButtonBackgroundColor;
+            _buttonBackground.color = new Color32(255, 255, 255, 0);
```
EliteAsian123, line 42: `Color.clear` should be used in place of `new Color32(255, 255, 255, 0)` (unless it causes lerping issues)

Assets/Script/Menu/Persistent/HelpBarButton.cs
```diff
@@ -1,36 +1,76 @@
...
+        {
+            _buttonBackground.color = ButtonBackgroundColor;
+            _buttonImage.color = ButtonBackgroundColor;
+            _buttonLabel.color = Color.white;
+            _buttonText.color = Color.white;
+        }
+
+        public void OnPointerExit(PointerEventData eventData)
+        {
+            _buttonBackground.color = new Color32(255, 255, 255, 0);
```
EliteAsian123, line 55: Same here

Assets/Art/Menu/Backgrounds/HelpBarHighlight.png
EliteAsian123, line 1: Try to fix this issue if you can
image

EliteAsian123 (changes requested):
Good work overall! I just requested a few small code changes.

Insomnautika:
Alright, I can make those changes. I'll let you know when its ready.

EliteAsian123 (approved):

PR #648, opened 2024-01-14
Move more lyric flag handling to YARG.Core, and implement handling of vocals range shifts

TheNathannator:
Depends on https://github.com/YARC-Official/YARG.Core/pull/135 and needs a YARG.Core submodule update before merging.

### DIFF
--- a/Assets/Script/Gameplay/Visuals/VocalElements/VocalNoteElement.cs
+++ b/Assets/Script/Gameplay/Visuals/VocalElements/VocalNoteElement.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using UnityEngine;
 using YARG.Core.Chart;
 
@@ -32,7 +32,7 @@ namespace YARG.Gameplay.Visuals
             UpdateLinePoints();
         }
 
-        private void UpdateLinePoints()
+        public void UpdateLinePoints()
         {
             // Create points
             _points.Clear();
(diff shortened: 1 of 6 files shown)


Assets/Script/Gameplay/Visuals/VocalElements/VocalNoteElement.cs
```diff
@@ -52,8 +52,11 @@ private void UpdateLinePoints()
             }
 
             // Set line info
+            float width = VocalTrack.CurrentNoteWidth;
             foreach (var line in _lineRenderers)
             {
+                line.startWidth = width;
+                line.endWidth = width;
```
EliteAsian123, line 59: This doesn't seem to work with the note overlap system, since it sets every line to the same width.
Old:
image
New:
image

For reference on how this works: the notes are a stack of line renderers of different widths so that they overlap in such a way to create the proper patterns.
image
I'm honestly not sure if this is the best way of doing it, but it's the only way I could think of, lol

EliteAsian123 (changes requested):

Assets/Script/Gameplay/Visuals/VocalElements/VocalNoteElement.cs
TheNathannator: Ah, so that's what that's for lol, I'll try to get that fixed. Can probably just set a width multiplier for each layer.

It might be better to handle the overlapping more explicitly in code if we can in the future, we could use `widthCurve` to set the width for the overlap areas. There isn't currently any way of the notes knowing when they're overlapping with each other though, so that would be more work.

EliteAsian123 (approved):

PR #650, opened 2024-01-16
Fix CurrentNavigationGroup not updating correctly when leaving the SettingsMenu

jgoyvaerts:


EliteAsian123:
This should work for now, ideally we should move the settings menu into the menu scene, though.

PR #656, opened 2024-01-23
Fix frets not displaying correctly when pressed/released when the game is paused

jgoyvaerts:
Fixes https://yarg.youtrack.cloud/issue/YARG-114


EliteAsian123:
Getting a ton of `Engine was forced to move an input time!` on vocals

PR #659, opened 2024-01-28
Display actual name of source in Music Library when sorted by source

poliwhirlge:
e.g. rb4 -> Rock Band 4

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -104,7 +104,7 @@ namespace YARG.Menu.MusicLibrary
             _subHeader.text = LibraryMode switch
             {
                 MusicLibraryMode.QuickPlay => "Quickplay",
-                MusicLibraryMode.Practice  => "Practice",
+                MusicLibraryMode.Practice => "Practice",
                 _ => throw new Exception("Unreachable.")
             };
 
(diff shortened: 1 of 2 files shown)


Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -161,7 +161,8 @@ protected override List<ViewType> CreateViewList()
             foreach (var section in _sortedSongs)
             {
                 // Create header
-                list.Add(new SortHeaderViewType(section.Key, section.Value.Count));
+                string key = Sort == SongAttribute.Source ? SongSources.SourceToGameName(section.Key) : section.Key;
+                list.Add(new SortHeaderViewType(key, section.Value.Count));
```
sonicfind, line 165: There's a small flaw that comes with this.
While this works well with sources defined in our database, anything that isn't appears as Custom/Unknown.

Which leads to stuff like this
image

The only way to address that issue in an efficient manner would be to have a second version of `SongSources.SourceToGameName` return whether or not a name was found for the given source alongside said name. 

If one was not found, then we could just append the inner source name to the end of `Custom/Unknown`.

I do agree with the basic idea though.

sonicfind (changes requested):

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
sonicfind: Making a second version of SongSources.SourceToGameName (preferably like `public static bool TrySourceToGameName(string id, out string name)`), will also require an accompanying `SongSources.GetSource` as well.

poliwhirlge:
If ```GetSource(id)``` is already returning the backup source wouldn't that mean that we can't identify the source and thus can only append the id to the end of ```Custom/Unknown```? Or is there a field I've missed that we can use to fill in this information?

sonicfind:
That's why I mentioned you need an accompanying `SongSources.TryGetSource` as well

sonicfind (approved):

PR #661, opened 2024-01-31
Display songs sorted by album if artist sort is selected in MusicLibrary

poliwhirlge:
Under each artist header, songs will be displayed starting from those with only 1 song in the album, then the albums grouped under subheadings, sorted in the album track order.


poliwhirlge:
gonna refactor in YARG.Core instead for efficiency

PR #670, opened 2024-02-08
DMX support, "Advanced" tab, Stagekit toggle

TheFatBastid:
Added scan (dmx over ethernet) support. Added an advanced settings tab. Added ability to turn off the stage kit.
Haukcode.sACN is now MIT licensed.


Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace YARG.Integration.Sacn
+{
+    public class SacnController : MonoBehaviour
+    {
+        //A 128-bit (16 byte) UUID that translates to "KEEP PLAYING YARG!"
+        private const float TARGET_FPS = 44f; //DMX spec says 44 updates per second is the max
+        private const string ACN_SOURCE_NAME = "YARG";
+        private const int UNIVERSE_SIZE = 512; //Each universe supports up to 512 channels
+        private const float TIME_BETWEEN_CALLS = 1f / TARGET_FPS;
```
EliteAsian123, line 16: Small nitpick, but I would group this these consts and readonlys like this

```cs
// DMX spec says 44 updates per second is the max
private const float TARGET_FPS = 44f; 
private const float TIME_BETWEEN_CALLS = 1f / TARGET_FPS;

// Each universe supports up to 512 channels
private const int UNIVERSE_SIZE = 512;

private const string ACN_SOURCE_NAME = "YARG";

// A 128-bit (16 byte) UUID that translates to "KEEP PLAYING YARG!"
private static readonly Guid AcnSourceId = new Guid("{4B454550-504C-4159-494E-475941524721}");

// THEN the rest of the fields...
```

TheFatBastid:
> Seems like sACN _does_ have a NuGet package _with_ a license. Can you confirm that this is the same package, and if so, attempt to use it via NuGetForUnity instead of importing the files?
> 
> https://www.nuget.org/packages/Haukcode.sACN

Same thing. He just added the license today when I e-mailed him. :) The difference being that I removed all the reading code, as YARG won't be receiving anything so about 10k lines, if that weight doesn't matter I'll try the package.

Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -0,0 +1,139 @@
...
+
+                case StageKitLedColor.All:
+                    SetChannelValues(_yellowChannels, ledIsSet);
+                    SetChannelValues(_greenChannels, ledIsSet);
+                    SetChannelValues(_blueChannels, ledIsSet);
+                    SetChannelValues(_redChannels, ledIsSet);
+                    break;
+
+                default:
+                    Debug.Log("(Sacn) Unknown color: " + color);
```
EliteAsian123, line 117: Should probably be a warning or error

Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -0,0 +1,139 @@
...
+        private byte[] _dataPacket = new byte[UNIVERSE_SIZE];
+
+        private static readonly Guid AcnSourceId = new Guid("{4B454550-504C-4159-494E-475941524721}");
+        private static SACNClient _sendClient;
+        //DMX channels - 8 per color to match the stageKit layout. Default channels, the user must change them in settings.
+        private static int[] _dimmerChannels;
+        private static int[] _redChannels;
+        private static int[] _greenChannels;
+        private static int[] _blueChannels;
+        private static int[] _yellowChannels;
```
EliteAsian123, line 27: If static variables are used like this, it should probably be a `MonoSingleton` instead.

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -44,9 +45,9 @@ protected override void OnChartLoaded(SongChart chart)
 
         private void Update()
         {
-            if (StageKitLightingController.Instance.StageKits.Count == 0)
+            if ( (StageKitLightingController.Instance.StageKits.Count == 0 || SettingsManager.Settings.StageKitEnabled.Value == false) && SettingsManager.Settings.DMXEnabled.Value == false)
```
EliteAsian123, line 48: `x == false` or `x == true` is redundant. Use `!x` or `x` respectively.

Assets/Script/Settings/Types/DMXChannelsSetting.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public DMXChannelsSetting(int[] value, Action<int[]> onChange = null) : base(onChange)
+        {
+            //DMX channels are 8-bit, so the range is 0-255
+            Min = 0;
+            Max = 255;
+
+            _value = value;
+        }
+        public override bool ValueEquals(int[] value)
```
EliteAsian123, line 21: Separate these methods with empty lines

Assets/Script/Settings/Types/DMXChannelsSetting.cs
```diff
@@ -0,0 +1,26 @@
...
+        public override string AddressableName => "Setting/DMXChannels";
+
+        public int Min { get; }
+        public int Max { get; }
+
+        public DMXChannelsSetting(int[] value, Action<int[]> onChange = null) : base(onChange)
+        {
+            //DMX channels are 8-bit, so the range is 0-255
+            Min = 0;
+            Max = 255;
```
EliteAsian123, line 17: Are these actually used? Better yet, make all of the `int[]` into `byte[]` instead!

Assets/Settings/Localization/Settings Shared Data.asset
```diff
@@ -915,6 +915,58 @@ MonoBehaviour:
     m_Key: Tab.Sound
     m_Metadata:
       m_Items: []
+  - m_Id: 113150520376582144
+    m_Key: Tab.Advanced
+    m_Metadata:
+      m_Items: []
+  - m_Id: 113150787738296320
+    m_Key: Header.Lighting General
```
EliteAsian123, line 923: Localization keys should not have spaces in them. Should be `Header.LightingGeneral` instead.

Assets/packages.config
```diff
@@ -9,4 +9,5 @@
   <package id="sqlite-net" version="1.6.292" />
   <package id="System.Diagnostics.Tracing" version="4.3.0" />
   <package id="System.Runtime.CompilerServices.Unsafe" version="6.0.0" />
+  <package id="UniRx" version="5.4.1.0" />
```
EliteAsian123, line 12: Is this package actually used in this PR?

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -162,8 +164,44 @@ public void OpenCalibrator()
 
             #endregion
 
+            #region Lighting
+            public ToggleSetting StageKitEnabled { get; } = new(true);
+            public ToggleSetting DMXEnabled { get; } = new(false);
```
EliteAsian123, line 169: Empty line after the region and also align the `{ get; }`s together like above

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -162,8 +164,44 @@ public void OpenCalibrator()
...
+
+            public DMXChannelsSetting DimmerChannels { get; } = new(
+                new int[]{ 0, 8, 16, 24, 32, 40, 48, 56 },
+                DmxCallback
+                );
```
EliteAsian123, line 174: The `);` should be on the same line as `DmxCallback`

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -18,6 +18,8 @@ public static partial class SettingsManager
     {
         public class SettingContainer
         {
+            public static event System.Action<int[]> OnDMXChannelsChanged;
```
EliteAsian123, line 21: This is a no, no. See my comment on the callback for more info.

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -162,8 +164,44 @@ public void OpenCalibrator()
...
+
+            #endregion
+
             #region Callbacks
 
+
+            private static void DmxCallback(int[] value)
+            {
+                //Because SacnController doesn't exist before this, I'm using an action here.
+                OnDMXChannelsChanged?.Invoke(value);
```
EliteAsian123, line 200: What *should* be happening instead if that the `SacnController` should be a `MonoSingleton` and this should call a method within there like `SacnController.Instance.UpdateDMXChannels()`

Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            //Many DMX fixtures have a 'Master dimmer' channel that controls the overall brightness of the fixture.
+            //Got to turn those on.
+            for (int i = 0; i < _dimmerChannels.Length; i++)
+            {
+                _dataPacket[_dimmerChannels[i]] = 255;
+            }
+        }
+
+        private void HandleValueChanged(int[] value)
```
EliteAsian123, line 52: This argument can be removed

Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -0,0 +1,139 @@
...
+            Debug.Log("Starting SacnController...");
+
+            SettingsManager.SettingContainer.OnDMXChannelsChanged += HandleValueChanged;
+
+            HandleValueChanged(null);
+
+            StageKitLightingController.Instance.OnLedSet += HandleEvent;
+
+            _sendClient = new SACNClient(senderId: AcnSourceId, senderName: ACN_SOURCE_NAME,
+                localAddress: SACNCommon.GetFirstBindAddress().IPAddress);
```
EliteAsian123, line 40: I'm not really sure how SACN works, but it *really really* should not be initialized *unless* the user specifically enables the settings, especially since we're dealing with network stuff and sending packets.

The DMX setting should use a callback to call a method within this class, to then initialize/destroy it. To be honest, we really should be doing that for the Discord integration as well.

Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            InvokeRepeating(nameof(Sender), 0, TIME_BETWEEN_CALLS);
```
EliteAsian123, line 42: `InvokeRepeating` is unoptimized if I remember correctly, and it's annoying to deal with, you should probably just use a timer within the `Update` method

Assets/Script/Integration/StageKit/StageKitLightingController.cs
```diff
@@ -221,6 +219,9 @@ public enum LedColor
         //Gets set by StageKitGameplay but this is just a fail-safe.
         public bool LargeVenue = false;
 
+        //This is used to send events to any other light controllers that might be in the scene (aka dmx lights)
+        public Action<StageKitLedColor, StageKitLed> OnLedSet;
```
EliteAsian123, line 223: Make this into an `event`.

Assets/Prefabs/Menu/Common/InputField.prefab
```diff
@@ -139,7 +139,21 @@ MonoBehaviour:
...
+        m_Mode: 1
+        m_Arguments:
+          m_ObjectArgument: {fileID: 0}
+          m_ObjectArgumentAssemblyTypeName: UnityEngine.Object, UnityEngine
+          m_IntArgument: 0
+          m_FloatArgument: 0
+          m_StringArgument: 
+          m_BoolArgument: 0
+        m_CallState: 2
   m_OnSubmit:
```
EliteAsian123, line 157: Haven't looked at this in Unity yet, so it's hard to tell; is this change intentional?

EliteAsian123 (changes requested):
Thanks, this is pretty awesome! Just a few styling and one important thing and this should be good!

EliteAsian123:
> Same thing. He just added the license today when I e-mailed him. :) The difference being that I removed all the reading code, as YARG won't be receiving anything so about 10k lines, if that weight doesn't matter I'll try the package.

Overall, I'd recommend packages when possible, so things aren't a pain to update.

TheFatBastid:
> > Same thing. He just added the license today when I e-mailed him. :) The difference being that I removed all the reading code, as YARG won't be receiving anything so about 10k lines, if that weight doesn't matter I'll try the package.
> 
> Overall, I'd recommend packages when possible, so things aren't a pain to update.

image
:(

EliteAsian123:
Damn okay, in which case we don't have much of a choice

Assets/Prefabs/Menu/Common/InputField.prefab
TheFatBastid: Probably not. I don't even know what kind of change that is. Only thing I did with InputField is added 8  of them to the DMX prefab.

Assets/Script/Integration/Sacn/SacnController.cs
TheFatBastid: I attempted a few different things for this, unitask, timers, etc and they all seemed to be more erratic than InvokeRepeating. The sacn converter that receives these messages (from YARG) had an incoming fluctuating fps  27 - 46, whereas the InvokeRepeating gave it a rock solid 44. I'll try other methods again.

RileyTheFox:
You could ask if the package's framework targets can be updated. It should have no issues being listed for .NET Standard 2.1 if it's set to target .NET 7.

https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1

EliteAsian123:
Yeah if you could do that that'll be great; I'd rather not include a bunch of library files unless *absolutely necessary*.

EliteAsian123 (approved):

PR #671, opened 2024-02-09
Bugfixes and improvements

jgoyvaerts:
Ideally don't squash the commits when merging so the changelog is updated correctly


Assets/Script/Gameplay/Player/BasePlayer.cs
```diff
@@ -277,6 +278,14 @@ protected void OnGameInput(ref GameInput input)
...
             if (GameManager.Paused)
             {
+                if (Player.Profile.GameMode == GameMode.Vocals ||
+                    Player.Profile.GameMode == GameMode.FourLaneDrums ||
+                    Player.Profile.GameMode == GameMode.FiveLaneDrums)
+                {
+                    // We don't need to track frets being held down in the pause menu for these game modes
+                    return;
+                }
+
```
EliteAsian123, line 288: This should probably be an abstract property in `BasePlayer` that is overridden in each instrument class instead

EliteAsian123 (changes requested):

Assets/Script/Gameplay/Player/BasePlayer.cs
jgoyvaerts: Good point! I was looking for a cleaner way to do it but couldn't immediatly come up with one. Fixed

PR #679, opened 2024-02-11
Add sorting by PlayCount, refactor filtering and sorting a bit

jgoyvaerts:
Just opening a draft pull request to get some feedback.

Had to change some stuff because a lot of it was in YARG.Core but core is not aware of playcounts, and it shouldn't be either.

There was also no distinction between filterable options and sortable options, so I tried making that.

In an ideal world, the whole search gets rewritten with unit tests, and a distinction between sort options and filter options, but that's a lot more work.

Open to feedback!

### DIFF
--- a/Assets/Script/Scores/ScoreContainer.cs
+++ b/Assets/Script/Scores/ScoreContainer.cs
@@ -200,6 +200,47 @@ namespace YARG.Scores
             return new List<SongMetadata>();
         }
 
+        public static Dictionary<int, List<SongMetadata>> GetSongsWithPlayCount()
+        {
+            try
+            {
+                var query =
+                    $"SELECT SongChecksum, COUNT(SongChecksum) AS `Count` FROM GameRecords GROUP BY SongChecksum";
+                var playCounts = _db.Query<PlayCountRecord>(query);
+
+                var allSongs = GlobalVariables.Instance.SongContainer.SongsByHash.ToDictionary(x => x.Key, x => x.Value.Pick());
+                var songs = new Dictionary<int, List<SongMetadata>>();
+                foreach (var record in playCounts)
+                {
+                    var hash = new HashWrapper(record.SongChecksum);
+                    if (allSongs.TryGetValue(hash, out var list))
+                    {
+                        if (songs.TryGetValue(record.Count, out var songList))
+                        {
+                            songList.Add(list);
+                        }
+                        else
+                        {
+                            songs[record.Count] = new List<SongMetadata> { list };
+                        }
+
+                        allSongs.Remove(hash);
+                    }
+                }
+
+                songs[0] = allSongs.Values.ToList();
+
+                return songs;
+            }
+            catch (Exception e)
+            {
+                Debug.LogError("Failed to load songs with play count from database. See error below for more details.");
+                Debug.LogException(e);
+            }
+
+            return new Dictionary<int, List<SongMetadata>>();
+        }
+
         public static void Destroy()
         {
             _db.Dispose();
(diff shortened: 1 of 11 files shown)


sonicfind:
Some concerns:

1. I don't see the reason to have a separate enum for sort & filter (especially when they're really just separate pieces of the same pie). If you have to have a function to convert from one to another with barely any difference in the values they hold, then that's just an unnecessary step.

2. Those changes to how instrument filtering works unfortunately are gonna be unpopular. This was close to how it worked after I changed some searching functionality, but people complained that clicking an instrument icon no longer sorted the songs by intensity level. The only compromise that can made for that would require adding a separate sort option for each and every instrument type and having those icons map to changing the sort. Regardless, the **sort by intensity** behavior must be prioritized first before trying to change how the filtering works for instruments.

3. The SongContainer type is meant to an unchanging object. Once the cache is given to its constructor, that's it. None of the public facing functions should be able to alter its internals. It's why everything is returned through ReadOnly interfaces. Anything that involves values that change irrespective of rescanning the music library needs to be handled externally. In this situation, that means the playcount sort needs to run and setup by SongSearching. Preferably as well, with how few people use that sort in question, and with how it requires accessing a file, I'd like it if that viewlist was only set up when the sort is selected or when returning from the results screen with that sort already selected. Otherwise, you force that reconstruction every time you enter that menu regardless of where you're entering from. Quit a song early? refresh list; Change a profile selection? refresh list; Edit settings? refresh list; etc.

sonicfind:
Update: I'll take back the first point if we do go the route of having separate sorts for each instrument type.

Unless we decide to allow something like: "FiveFretGuitar: 5" for intensity based filtering, which may be too far.

Assets/Script/Scores/ScoreContainer.cs
```diff
@@ -200,6 +200,47 @@ public static List<SongMetadata> GetMostPlayedSongs(int maxCount)
...
+            }
+            catch (Exception e)
+            {
+                Debug.LogError("Failed to load songs with play count from database. See error below for more details.");
+                Debug.LogException(e);
+            }
+
+            return new Dictionary<int, List<SongMetadata>>();
+        }
+
```
sonicfind, line 243: I found a simpler way to handle this function (though the branch needs a rebase to use it)

```
public static SortedDictionary<int, List<SongEntry>> GetSongsWithPlayCount()
{
    var songs = new SortedDictionary<int, List<SongEntry>>();
    try
    {
        var query = $"SELECT SongChecksum, COUNT(SongChecksum) AS `Count` FROM GameRecords GROUP BY SongChecksum";
        var playCounts = _db.Query<PlayCountRecord>(query).ToDictionary(record => new HashWrapper(record.SongChecksum));
        
        foreach (var song in GlobalVariables.Instance.SongContainer.SongsByHash)
        {
            int count = playCounts.TryGetValue(song.Key, out var record) ? record.Count : 0;
            if (!songs.TryGetValue(count, out var list))
            {
                songs.Add(count, list = new List<SongEntry>());
            }
            list.AddRange(song.Value);
        }
    }
    catch (Exception e)
    {
        Debug.LogError("Failed to load songs with play count from database. See error below for more details.");
        Debug.LogException(e);
    }
    return songs;
}
```

jgoyvaerts: Cool! I'll look at it in the weekend, been swamped with work

sonicfind:
Update: disregard point 3 to a degree, While everything else should remain "const", considering the volatility of this sort in particular, there's not much of an easy choice.

For example, I'm having to do something similar with implementing a "currently playable" sort that changes based on the player instruments currently active.

EliteAsian123:
Unfortunately, this is quite out of date at this point, so I'll be closing the PR. Please feel free to open a new one if you come back to the changes!

PR #680, opened 2024-02-11
Fix vocals players after the first not being forced to sit out if there are no vocals parts

Supersonicboss1:
This just adds a check to make sure that the vocals part actually exists, as well as being the correct style.
Fixes YARG-149

### DIFF
--- a/Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
+++ b/Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
@@ -537,7 +537,7 @@ namespace YARG.Menu.DifficultySelect
                     var playerInstrument = player.Profile.CurrentInstrument;
                     if (playerInstrument is Instrument.Vocals or Instrument.Harmony)
                     {
-                        return playerInstrument == instrument;
+                        return playerInstrument == instrument && parts.HasInstrument(instrument);
                     }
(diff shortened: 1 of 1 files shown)


Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
```diff
@@ -537,7 +537,7 @@ private bool HasPlayableInstrument(AvailableParts parts, Instrument instrument)
                     var playerInstrument = player.Profile.CurrentInstrument;
                     if (playerInstrument is Instrument.Vocals or Instrument.Harmony)
                     {
-                        return playerInstrument == instrument;
```
EliteAsian123, line 540: It may be better to add a check *before* the loop instead, so you don't have to loop through every player first. Something like this:
```cs
if (!parts.HasInstrument(instrument)) {
    return false;
}
```

EliteAsian123 (changes requested):
Thanks for your PR! I've requested some changes below. Have a good one!

Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
Supersonicboss1: Fixed, thanks!

PR #686, opened 2024-02-16
Added search toggle buttons to music library search bar

JayDiddyThaGOAT:
# Summary

I got the concept from the YARG design library and added functionality & UI for the search toggles. I also used the dialog system for create a dialog for filtering the Instruments & Years. Also, I added a new menu color for the deactivated toggles.

# Code Details

Instead of having the search query be typed in the search bar and it's code in the music library, the search bar code has been organized into it's own class and the query is stored in a private variable. The Search function has been refactored to parse out the filters in the query using Regex and only show the toggled filters in the music library.

# Demo

https://github.com/YARC-Official/YARG/assets/25489150/b12d11dc-5b91-4e07-9ba3-e83e72fd8213


JayDiddyThaGOAT:
Kadu says he is doing another update on the search bar UI, but I think since the functionality is done, it should be merged as the way it is. When Kadu finishes with the work, I'll update the UI until then.

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
```diff
@@ -0,0 +1,416 @@
...
+            return _searchContext.Refresh(sort);
+        }
+
+        public IReadOnlyList<SongCategory> Search(SongAttribute sort)
+        {
+            if (_currentSearchFilter == SongAttribute.Unspecified)
+            {
+                if (Regex.IsMatch(_searchField.text, @":"))
+                {
+                    var match = Regex.Match(_searchField.text, @"^(.*?):");
```
RileyTheFox, line 113: Would recommend moving all applicable regex strings into some constant strings at the top of the class, with some comments describing what it does (regex isn't easy to understand at first glance)

sonicfind: Example: I have no idea what those mean LOL

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
```diff
@@ -0,0 +1,416 @@
...
+                        {
+                            _fullSearchQuery = $"{filter}:";
+                        }
+                        else
+                        {
+                            _fullSearchQuery += $";{filter}:";
+                        }
+                    }
+                }
+            }
```
sonicfind, line 174: Just syntactical: would be a good spot for `else if` over nested `if`s

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
```diff
@@ -0,0 +1,438 @@
...
+using TMPro;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using YARG.Core.Song;
+using YARG.Menu.Dialogs;
+using YARG.Menu.Navigation;
+using YARG.Menu.Persistent;
+using YARG.Song;
+
+namespace YARG.Menu.SongSearching
```
EliteAsian123, line 13: Namespace should be `YARG.Menu.MusicLibrary`

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
JayDiddyThaGOAT: Updated those patterns with comments in 25b1985. Had to watch a long ass tutorial to understand Regex, so I should've took that into consideration.

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
```diff
@@ -0,0 +1,438 @@
...
+        private const string COLON_AFTER_WORD = @"^(.*?):";
+
+        /// <summary>
+        /// Regex pattern to match a whole word
+        /// - \b: Asserts a word boundary, matching the position between a word character (i.e., a letter,
+        ///       digit, or underscore) and a non-word character (or vice versa).
+        /// - \w+: Matches one or more word characters (i.e., letters, digits, or underscores).
+        /// - \b: Asserts another word boundary.
+        /// </summary>
+        private const string WHOLE_WORD_PATTERN = @"\b\w+\b";
```
EliteAsian123, line 69: Constants should be ordered at the top, however, regexes should be compiled when able.
[Example](https://github.com/YARC-Official/YARG/blob/edfef1b179b98931a04315e419a7bdfc8666bc11/Assets/Script/Replays/ReplayEntry.cs#L12).

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
JayDiddyThaGOAT: Unnested those if statements in 2ab33b8

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
```diff
@@ -0,0 +1,438 @@
...
+        private bool _searchNavPushed;
+        private bool _wasSearchFieldFocused;
+
+        public bool IsSearching => !string.IsNullOrEmpty(_fullSearchQuery);
+        public bool IsCurrentSearchInField => _searchQueries[_currentSearchFilter] == _searchField.text;
+        public bool IsUpdatedSearchLonger => _searchField.text.Length > _currentSearchText.Length;
+        public bool IsUnspecified => _searchContext.IsUnspecified();
+
+        public delegate void OnSearchQueryUpdated(bool forceUpdate);
+        public OnSearchQueryUpdated SearchQueryUpdated;
```
EliteAsian123, line 35: Should be an event.

EliteAsian123 (changes requested):
Hi, thanks for your PR!

I've pointed out some simple things below, however, I think the main problem I'm having right now is that the way this works is quite confusing.

https://github.com/YARC-Official/YARG/assets/29520859/5f1d584e-d7a9-4cdc-bafa-5b07dc084934

I had to look at the code to actually understand what was going on here, which in theory is fine. However, it's very, very confusing for the end-user I feel like (some opinions from other people would be nice as well, I may just be dumb).

For me, what would make more sense, is that you type in a query, and the button group on the side decides *what* you're searching for. So, for example, if you searched “Metal” and you selected the genre tag, it'll show results for *only* songs with the “Metal” genre. If you then select the artist tag, it'll show results *only* artists with the word “Metal” in it (e.g., Metallica). Right now, it seems you are able to type in a query for *each tag,* which gives some nice flexibility, but the way it's setup is very annoying to work with. Also, it's kind of weird that it lets you type something in when you haven't selected anything with this system.

The “More Filters” button is *even more confusing* because there's no clear indication that there's actually anything selected, unless you go back into the menu. For this, I understand that there's no UI for it yet, but, for now, I think that this should be left out in this case as it isn't the greatest UX as well.

tl;dr: For now, it's probably best to disable the “More Filters” button and also make it so you can only use one search query at a time instead of multiple, unless some UI changes are made.

Thanks!

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
JayDiddyThaGOAT: Removed the constants, but compiled the Regexes in 1abbc1e

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
JayDiddyThaGOAT: Change namespace to YARG.Menu.MusicLibrary in 47a35f5. Thus deleted the references to the old namespace in 3bffcad

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
```diff
@@ -95,9 +102,9 @@ public void SetSearchInput(SongAttribute attribute, string input)
                 }
                 else
                 {
-                    var currentQuery = $"{filter}:{_searchQueries[attribute]}";
-                    _fullSearchQuery = Regex.Replace(_fullSearchQuery, currentQuery, updatedQuery,
-                        RegexOptions.IgnoreCase);
+                    // Regex pattern: The filter specified and the search query tagged with that filter
+                    var currentQueryRegex = new Regex($"{filter}:{_searchQueries[attribute]}", DefaultOptions);
```
RileyTheFox, line 106: Don't compile regular expressions if they are going to be frequently created (such as when using string interpolation), it's actually slower than just interpreting them. Use Regex.Match or Regex.Replace in place in these spots.

Expressions are cached by the runtime briefly so if the same expression is used it doesn't need to be interpreted again.

Reference: https://blog.codinghorror.com/to-compile-or-not-to-compile/

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
```diff
@@ -161,23 +168,24 @@ public IReadOnlyList<SongCategory> Search(SongAttribute sort)
...
                 // Regex pattern representing a word boundary around the filter value
-                string filterFoundPattern = $@"\b{filter}\b";
+                var filterFoundRegex = new Regex($@"\b{filter}\b", DefaultOptions);
 
-                if (Regex.IsMatch(_fullSearchQuery, filterFoundPattern))
+                if (filterFoundRegex.IsMatch(_fullSearchQuery))
                 {
-                    _fullSearchQuery = Regex.Replace(_fullSearchQuery, currentQuery, updatedQuery, RegexOptions.IgnoreCase);
+                    // Regex pattern: The filter specified and the search query tagged with that filter
+                    var currentQueryRegex = new Regex($"{filter}:{_searchQueries[_currentSearchFilter]}", DefaultOptions);
```
RileyTheFox, line 180: Same as the other regex comment

RileyTheFox (changes requested):

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
JayDiddyThaGOAT: Thanks for the heads up! Fixed in ab1a412

Assets/Script/Menu/MusicLibrary/SongSearchingField.cs
JayDiddyThaGOAT: Thanks for the heads up! Fixed in ab1a412

JayDiddyThaGOAT:
I replaced the More Filters button with the Instruments & Years filter toggles. Also, I found an alternative to the UI flow by darkening the toggles previously selected. 

### UI Flow
- Pressing a deactivated toggle once activates it (adds to the query)
- Pressing the same toggle deactivates it (removes it out of the query)
- Pressing a new toggle activates it while darkening the previously selected toggle
- Pressing a previously selected toggle activates it while darkening the previously activated toggle

### Demo
https://github.com/YARC-Official/YARG/assets/25489150/c9a8fee4-7715-42b1-865c-5d17dbacb28f

JayDiddyThaGOAT:
I found a bug where you can't press the colon. I'll fix that right now

JayDiddyThaGOAT:
Fixed not being able to type in a colon in 15fa5f9. I had a feature where you could type in a filter and then a colon and that'll automatically add the filter to the query, but it wasn't executed well so I just reverted the change to my initial plan.

Under no filter, you can search for songs the original way

EliteAsian123 (changes requested):
Sorry for the late update! Definitely *a lot* better, but there are still some small things that should be changed. 

For example, you can still search stuff with *nothing* highlighted, which doesn't make much sense. "Track" or something of the like should be selected by default.

Another thing is that the UI doesn't really make it obvious that you need to click twice in order to actually get rid of the filter/search. Instead, there should be a little X you gotta click or something. Maybe @kaduwaengertner can help with that.

Thanks! Sorry for all of the requested changes, haha.

sonicfind:
> For example, you can still search stuff with _nothing_ highlighted, which doesn't make much sense. "Track" or something of the like should be selected by default.

We've always had it so that searching with no filters attached resulted in checking through both song name & artist for the same string, and displaying the results of both. It's actually a nice feature since those are definitely the most common filters people use. It'd be cumbersome to have to switch between those two in particular.

JayDiddyThaGOAT:
> Another thing is that the UI doesn't really make it obvious that you need to click twice in order to actually get rid of the filter/search. Instead, there should be a little X you gotta click or something. Maybe @kaduwaengertner can help with that.

My suggestion for this is to make the toggle the Cancel color when you hover over an active toggle. I'll commit this if you or anyone else approves this. @EliteAsian123 

https://github.com/YARC-Official/YARG/assets/25489150/87192843-e7d9-46ff-bc13-658941552f31

JayDiddyThaGOAT:
> For example, you can still search stuff with nothing highlighted, which doesn't make much sense. "Track" or something of the like should be selected by default.

Searching stuff with nothing highlighted is the Unspecified search, which by default checks through the song name & artist in the same string like @sonicfind says.

EliteAsian123:
> We've always had it so that searching with no filters attached resulted in checking through both song name & artist for the same string, and displaying the results of both. It's actually a nice feature since people usually search through those filters interchangeably most of the time. It'd be cumbersome to have to manually switch between those two in particular.

If that's the case then I suppose you are right.

Using colours to represent whether a filter will be added/removed is a good idea, so it's fine with me. @kaduwaengertner may wanna change that, but we don't have a concept, so it's perfect for now.

JayDiddyThaGOAT:
Awesome! Just pushed the hovering colors in 1a418ee

EliteAsian123 (approved):

PR #690, opened 2024-02-17
YARG-176 Add button to go to currently-playing song in music player

vinaykapadia:
https://yarg.youtrack.cloud/issue/YARG-176/Add-button-to-go-to-currently-playing-song-in-music-player

### DIFF
--- a/Assets/Script/Helpers/Extensions/SongMetadataExtensions.cs
+++ b/Assets/Script/Helpers/Extensions/SongMetadataExtensions.cs
@@ -82,6 +82,8 @@ namespace YARG.Helpers.Extensions
 
         public static void LoadAudio(this SongMetadata song, IAudioManager manager, float speed, params SongStem[] ignoreStems)
         {
+            GlobalVariables.Instance.CurrentSong = song;
+
             if (song.IniData != null)
             {
                 LoadIniAudio(song.IniData, manager, speed, ignoreStems);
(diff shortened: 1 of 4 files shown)


Assets/Script/Helpers/Extensions/SongMetadataExtensions.cs
```diff
@@ -82,6 +82,8 @@ public static class SongMetadataExtensions
 
         public static void LoadAudio(this SongMetadata song, IAudioManager manager, float speed, params SongStem[] ignoreStems)
         {
+            GlobalVariables.Instance.CurrentSong = song;
```
EliteAsian123, line 85: Setting this value probably isn't the best idea as it could have unintended side effects (if I'm going to be honest, this whole system isn't great in the first place). Instead, it should be a value accessible in the `MusicPlayer` itself, or a method in the `MusicPlayer` that takes you to the song.

EliteAsian123 (changes requested):
Hi, sorry for the late review! Just one small change and this should be good to merge.

Assets/Script/Helpers/Extensions/SongMetadataExtensions.cs
vinaykapadia: Cool I'll take care of that on Monday.

vinaykapadia: Ok, done!

PR #691, opened 2024-02-18
Add remaining text notifications and setting for note streak frequency

grishhung:
Added the following text notifications:

- `BASS GROOVE` (reach a 6x multiplier; automatically removes note streak from queue)
- `HOT START` (reach a 4x multiplier without missing any notes or overhitting)
- `OVERDRIVE READY` (marked as `StarPowerReady` internally)
- `STRONG FINISH` (finish the song with at least a 4x multiplier)

Changed `%d-NOTE STREAK` to be configurable with the following milestones:

- `Frequent` (matches Guitar Hero {50, 100, 200, 300, ...})
- `Sparse` (matches ScoreSpy {50, 100, 250, 500, 750, ...})
- `Disabled`

### DIFF
--- a/Assets/Script/Gameplay/HUD/New Variables Group Asset.asset
+++ b/Assets/Script/Gameplay/HUD/New Variables Group Asset.asset
@@ -1,0 +1,18 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!114 &11400000
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: b1e3e22ff9634e441a479f01f905a2f7, type: 3}
+  m_Name: New Variables Group Asset
+  m_EditorClassIdentifier: 
+  m_Variables: []
+  references:
+    version: 2
+    RefIds: []
--- a/Assets/Script/Gameplay/HUD/TextNotifications.cs
+++ b/Assets/Script/Gameplay/HUD/TextNotifications.cs
@@ -5,6 +5,13 @@ using YARG.Settings;
 
 namespace YARG.Gameplay.HUD
 {
+    public enum NoteStreakFrequencyMode
+    {
+        Frequent,
+        Sparse,
+        Disabled
+    }
+
     public class TextNotifications : MonoBehaviour
     {
         [SerializeField]
@@ -46,6 +53,26 @@ namespace YARG.Gameplay.HUD
             _notificationQueue.Enqueue(new TextNotification(TextNotificationType.FullCombo, "FULL COMBO"));
         }
 
+        public void ShowHotStart()
+        {
+            _notificationQueue.Enqueue(new TextNotification(TextNotificationType.HotStart, "HOT START"));
+        }
+
+        public void ShowBassGroove()
+        {
+            _notificationQueue.Enqueue(new TextNotification(TextNotificationType.BassGroove, "BASS GROOVE"));
+        }
+
+        public void ShowStarPowerReady()
+        {
+            _notificationQueue.Enqueue(new TextNotification(TextNotificationType.StarPowerReady, "OVERDRIVE READY"));
+        }
+
+        public void ShowStrongFinish()
+        {
+            _notificationQueue.Enqueue(new TextNotification(TextNotificationType.StrongFinish, "STRONG FINISH"));
+        }
+
         public void UpdateNoteStreak(int streak)
         {
             // Don't build up notifications during a solo
@@ -67,7 +94,11 @@ namespace YARG.Gameplay.HUD
             // Queue the note streak notification
             if (_streak >= _nextStreakCount)
             {
-                _notificationQueue.Enqueue(new TextNotification(TextNotificationType.NoteStreak, $"{_nextStreakCount}-NOTE STREAK"));
+                if (!SettingsManager.Settings.NoteStreakFrequency.ValueEquals(NoteStreakFrequencyMode.Sparse))
+                {
+                    _notificationQueue.Enqueue(new TextNotification(TextNotificationType.NoteStreak,
+                        $"{_nextStreakCount}-NOTE STREAK"));
+                }
                 NextNoteStreakNotification();
             }
         }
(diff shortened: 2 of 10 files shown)


Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -245,6 +250,34 @@ protected void UpdateBaseVisuals(BaseStats stats, BaseEngineParameters enginePar
...
+            }
+
+            bool currentBassGrooveState = stats.ScoreMultiplier == 6;
+
+            if (!PreviousBassGrooveState && currentBassGrooveState)
+            {
+                TrackView.ShowBassGroove();
+            }
+
+            PreviousBassGrooveState = currentBassGrooveState;
```
EliteAsian123, line 270: `BaseParameters.MaxMultiplier` should probably be used here to make things more future-proof. As for the bass groove, the instrument can be checked with `Player.Profile.CurrentInstrument`.

Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -338,9 +371,17 @@ protected virtual void OnNoteHit(int index, TNote note)
...
             }
 
-            if (index == Notes.Count - 1 && IsFc && note.ParentOrSelf.WasFullyHit())
+            if (index == Notes.Count - 1 && note.ParentOrSelf.WasFullyHit())
             {
-                TrackView.ShowFullCombo();
+                if (Combo >= 30)
+                {
+                    TrackView.ShowStrongFinish();
+                }
```
EliteAsian123, line 379: The `30` here should probably be commented with an explanation of where the number comes from, or should be based on `BaseParameters.MaxMultiplier` (depending on where the number comes from).

Assets/Script/Gameplay/HUD/New Variables Group Asset.asset
EliteAsian123, line 1: Was this file meant to be created?

EliteAsian123 (changes requested):
Thanks for the PR! Overall very nice! Just a few minor tweaks and this should be good to go.

Assets/Script/Gameplay/Player/TrackPlayer.cs
grishhung: Thanks! I'll update this accordingly.

Assets/Script/Gameplay/Player/TrackPlayer.cs
grishhung: That's a good point -- the idea is to have both `HOT START` and `STRONG FINISH` be marked by being at 4x multiplier (or higher).

Assets/Script/Gameplay/HUD/New Variables Group Asset.asset
grishhung: Whoops! No, it was not. Removing~

Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -91,6 +98,12 @@ public abstract class TrackPlayer : BasePlayer
             // Move the HUD location based on the highway length
             var change = ZeroFadePosition - DEFAULT_ZERO_FADE_POS;
             _hudLocation.position = _hudLocation.position.AddZ(change);
+
+            // Determine if a track is bass or not for the BASS GROOVE text notification
+            IsBass = Player.Profile.CurrentInstrument.Equals(Instrument.FiveFretBass)
+                || Player.Profile.CurrentInstrument.Equals(Instrument.SixFretBass)
+                || Player.Profile.CurrentInstrument.Equals(Instrument.ProBass_17Fret)
+                || Player.Profile.CurrentInstrument.Equals(Instrument.ProBass_22Fret);
```
EliteAsian123, line 106: Don't use `.Equals`, do something like `Player.Profile.CurrentInstrument is Instrument.FiveFretBass or Instrument.SixFretBass or ...` etc.

EliteAsian123 (changes requested):

Assets/Script/Gameplay/Player/TrackPlayer.cs
grishhung: Changed this to use the `or` operator instead. Thanks!

Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -338,9 +380,18 @@ protected virtual void OnNoteHit(int index, TNote note)
                 }
             }
 
-            if (index == Notes.Count - 1 && IsFc && note.ParentOrSelf.WasFullyHit())
+            if (index == Notes.Count - 1 && note.ParentOrSelf.WasFullyHit())
```
RileyTheFox, line 383: Would recommend doing `>=` in any event the note index jumps higher than the note count (should never happen but it acts as a fail-safe and would still trigger the notification)

Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -338,9 +380,18 @@ protected virtual void OnNoteHit(int index, TNote note)
...
             {
-                TrackView.ShowFullCombo();
+                // 30 to coincide with 4x multiplier (including on bass)
+                if (Combo >= 30)
```
RileyTheFox, line 386: We had some discussion in Discord a while back about not displaying both "Strong Finish" and "Full Combo" within the same song. Only one should be shown, Full Combo obviously taking priority.

So for example

```
if (IsFc)
{
    // show full combo
} else if(Combo >= 30)
{
    // show strong finish
}
```

Assets/Script/Gameplay/HUD/TextNotifications.cs
```diff
@@ -46,6 +53,26 @@ public void ShowFullCombo()
...
+        }
+
+        public void ShowBassGroove()
+        {
+            _notificationQueue.Enqueue(new TextNotification(TextNotificationType.BassGroove, "BASS GROOVE"));
+        }
+
+        public void ShowStarPowerReady()
+        {
+            _notificationQueue.Enqueue(new TextNotification(TextNotificationType.StarPowerReady, "OVERDRIVE READY"));
```
RileyTheFox, line 68: I know this is a personal preference over GH/RB but the entire game, front end and back end refer to it as Star Power so this should be changed to "STAR POWER READY" too

RileyTheFox (changes requested):

Assets/Script/Gameplay/HUD/TextNotifications.cs
```diff
@@ -67,7 +94,11 @@ public void UpdateNoteStreak(int streak)
             // Queue the note streak notification
             if (_streak >= _nextStreakCount)
             {
-                _notificationQueue.Enqueue(new TextNotification(TextNotificationType.NoteStreak, $"{_nextStreakCount}-NOTE STREAK"));
+                if (!SettingsManager.Settings.NoteStreakFrequency.ValueEquals(NoteStreakFrequencyMode.Sparse))
```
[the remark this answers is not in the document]
EliteAsian123: I didn't catch this the first time around, but I also don't recommend using `ValueEquals` unless it's for generics, etc. Use `NoteStreakFrequency.Value == NoteStreakFrequencyMode.Whatever` instead.

Assets/Script/Gameplay/HUD/TextNotifications.cs
grishhung: Ahhh. I thought we were referring to it as "overdrive" so I changed it. Will change it to `STAR POWER READY` to maintain consistency. Thanks!

Assets/Script/Gameplay/Player/TrackPlayer.cs
grishhung: Gotcha. Will change this accordingly.

I do agree that (as many have pointed out) a `FULL COMBO` is the strongest finish possible. `STRONG FINISH` coming after the `FULL COMBO` is consistent with how Guitar Hero does it, but it's also definitely redundant.

I think something that someone floated was the idea of having a "lifetime stats" thing that would keep track of the number of times that you've gotten `FULL COMBO`s and `STRONG FINISH`es and the like, but even if we did do that, we could just count a `FULL COMBO` as a `STRONG FINISH` or something.

Assets/Script/Gameplay/Player/TrackPlayer.cs
grishhung: Good idea. Changing this accordingly.

Assets/Script/Gameplay/HUD/TextNotifications.cs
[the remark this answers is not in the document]
grishhung: Oops, big miss here. >_< Will address both.

Assets/Script/Gameplay/HUD/TextNotifications.cs
```diff
@@ -105,20 +136,42 @@ private IEnumerator ShowNextNotification(string notificationText)
 
         private void NextNoteStreakNotification()
         {
-            switch (_nextStreakCount)
+            if (SettingsManager.Settings.NoteStreakFrequency.ValueEquals(NoteStreakFrequencyMode.Frequent))
```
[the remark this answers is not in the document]
grishhung: Sure, I'll condense with the `when` keyword.

RileyTheFox (approved):

EliteAsian123 (approved):

PR #692, opened 2024-02-19
Added basic Logitech RGB support for keyboards

TheFatBastid:
Maps the Stage Kit calls 1:1 to keyboard keys. Just the first step, next would be making the lighting cues more custom to take advantage of the full keyboard. Can be disabled in settings.


EliteAsian123:
This is really cool, but I personally see an issue with this feature, or rather this set of features:
image
That is **a lot of required libraries, DLLs, etc.** that we need to verify on different platforms, and that we need to include to the project. This can cause (in my opinion) bloat. However, that is just my stance.

An alternative and better solution would be to support one library that works for everything, or almost everything on different platforms. An example of this would be [OpenRGB](https://openrgb.org), however, I'm not sure if that's the best thing to use (I'm sure there are others as well that have pros and cons).

Point is, is that we **want to avoid using a bunch of libraries when possible** because they can cause compatibility issues, and potentially other issues that'll have to be fixed if we choose to move to IL2CPP, and probably more as well. Using one singular library, would solve this problem (or at least lessen it) while also keeping compatibility with devices.

Again, this is only my stance, so some opinions from @RileyTheFox @TheNathannator @sonicfind (sorry for the mass ping) would be nice as well.

Thanks!

TheFatBastid:
I don't disagree with any of that. Not only that, the manufactures libraries tend to incomplete and outdated as well. The logitech dll and sdk is from 2018. That is asking for trouble in the future. Implementing one of the 'all in one' platforms is fine, but just like that old [xkcd](https://xkcd.com/927/) there are a bunch of them as well. If we want to have just one library for each of the following: home (like wled), pro (the dmx stuff), peripherals (openRGB or similar), and native (the stagekit), Then have maybe like a plugins library that could be maintained separately if needed? That way we can still be 'modless' but not everyone needs a pack-in to support the rgb on their Coleco speak-n-spell.

RileyTheFox:
Fully agree with @EliteAsian123 here. We **will** be moving to IL2CPP building in the future and having all these different dependencies will be a headache.

This sort of stuff is much better off being supported by external applications by sending events over WebSocket to applications. I fear that wanting to support all these things is going to bloat the game massively when only a handful of people would ever use it.

TheFatBastid:
Going to try openRGB, first going to refactor the stage kit stuff.

PR #697, opened 2024-02-20
Fix sACN startup oversight

TheFatBastid:
Oops, I didn't add anything to check the on/off of the setting at startup, this fixes that.


EliteAsian123:
Hi, thanks! I don't think this is actually the case however, I'm going the log without this PR
image
All callbacks are automatically called at the start of the game.

PR #699, opened 2024-02-20
Starpower SFX only on multitrack songs.

vinaykapadia:
Changes the "Use Starpower FX" toggle to a drop down. "Off" and "Always" do what the toggle used to do. "Multitrack Only" reverbs only if you're using an instrument specific stem, will not reverb the song stem.

### DIFF
--- a/Assets/Script/Settings/SettingsManager.Settings.cs
+++ b/Assets/Script/Settings/SettingsManager.Settings.cs
@@ -1,6 +1,7 @@
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.InputSystem;
+using YARG.Audio;
 using YARG.Core.Audio;
 using YARG.Gameplay.HUD;
 using YARG.Helpers;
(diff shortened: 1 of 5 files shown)


Assets/Settings/Localization/Settings_en-US.asset
```diff
@@ -1107,6 +1107,18 @@ MonoBehaviour:
...
+  - m_Id: 120322696970141696
+    m_Localized: Off
+    m_Metadata:
+      m_Items: []
+  - m_Id: 120322832085450752
+    m_Localized: Multitrack Only
+    m_Metadata:
+      m_Items: []
+  - m_Id: 120322882958163968
+    m_Localized: Always
```
RileyTheFox, line 1119: Should rename this to "On" in my opinion. "Always" gives the impression that it will be enabled all of the time, regardless of star power status (and is a feature people have requested) so it should be differentiated.

Please change this in all areas where "Always" is used (as well as in the code), I'm just commenting on this one instance

RileyTheFox (changes requested):

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -354,7 +360,7 @@ private static void MusicPlayerVolumeCallback(float volume)
                 HelpBar.Instance.MusicPlayer.UpdateVolume();
             }
 
-            private static void UseStarpowerFxChange(bool value)
+            private static void UseStarpowerFxChange(StarPowerFxMode value)
```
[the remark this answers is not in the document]
vinaykapadia: Cool, will do.

Assets/Settings/Localization/Settings_en-US.asset
vinaykapadia: Sounds good.

Assets/Script/Settings/SettingsManager.Settings.cs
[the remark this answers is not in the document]
vinaykapadia: Done!

Assets/Settings/Localization/Settings_en-US.asset
vinaykapadia: Done!

RileyTheFox (approved):

PR #701, opened 2024-02-22
Play Previous Song button.

vinaykapadia:
Stores pretty much an unlimited number of songs. Will go previous until the first song, and then will just start that song over. For the button visual, I just used the same icon as Next Song, and did a flip transform.

### DIFF
--- a/Assets/Script/Menu/Persistent/MusicPlayer.cs
+++ b/Assets/Script/Menu/Persistent/MusicPlayer.cs
@@ -1,9 +1,9 @@
-﻿using System.Diagnostics.CodeAnalysis;
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using Cysharp.Threading.Tasks;
 using TMPro;
 using UnityEngine;
 using UnityEngine.UI;
-using YARG.Audio;
 using YARG.Settings;
 using YARG.Helpers.Extensions;
 using YARG.Core.Audio;
@@ -13,7 +13,9 @@ namespace YARG.Menu.Persistent
 {
     public class MusicPlayer : MonoBehaviour
     {
-        public static SongEntry NowPlaying = null;
+        public static SongEntry NowPlaying => nowPlayingPointer >= 0 ? NowPlayingList[nowPlayingPointer] : null;
+        private static List<SongEntry> NowPlayingList = new();
+        private static int nowPlayingPointer = -1;
 
         [SerializeField]
         private Image _playPauseButton;
@@ -57,8 +59,14 @@ namespace YARG.Menu.Persistent
 
         private async UniTask NextSong()
         {
-            var song = GlobalVariables.Instance.SongContainer.GetRandomSong();
-            NowPlaying = song;
+            nowPlayingPointer++;
+
+            if (nowPlayingPointer == NowPlayingList.Count)
+            {
+                NowPlayingList.Add(GlobalVariables.Instance.SongContainer.GetRandomSong());
+            }
+            
+            var song = NowPlayingList[nowPlayingPointer];
             await UniTask.RunOnThreadPool(() => song.LoadAudio(GlobalVariables.AudioManager, 1f, SongStem.Crowd));
 
             // Set song title text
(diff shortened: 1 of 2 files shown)


Assets/Script/Menu/Persistent/MusicPlayer.cs
```diff
@@ -57,8 +59,14 @@ private void OnDisable()
...
-            var song = GlobalVariables.Instance.SongContainer.GetRandomSong();
-            NowPlaying = song;
+            nowPlayingPointer++;
+
+            if (nowPlayingPointer == NowPlayingList.Count)
+            {
+                NowPlayingList.Add(GlobalVariables.Instance.SongContainer.GetRandomSong());
+            }
+            
+            var song = NowPlayingList[nowPlayingPointer];
```
sonicfind, line 69: Big problem here: with no reset builtin, the music player will forever hold onto SongEntry objects that are no longer present in a users library. You can not just hold onto every song you ever get. It should reset every time you scan the library and every time the music player is stopped.

Also, as Jaden already mentioned, he'll just be able to spam that button and bloat memory usage. You should *never* allow a user to purposefully harm their system. Three should be the max.

sonicfind (changes requested):

vinaykapadia:
Added the limit. Will add the reset on scan in the next day or so.

vinaykapadia:
Ok @sonicfind I've added the reset.

Assets/Script/Menu/Persistent/MusicPlayer.cs
```diff
@@ -13,7 +13,10 @@ namespace YARG.Menu.Persistent
 {
     public class MusicPlayer : MonoBehaviour
     {
-        public static SongEntry NowPlaying = null;
+        public static SongEntry NowPlaying => nowPlayingPointer >= 0 ? NowPlayingList[nowPlayingPointer] : null;
+        private static List<SongEntry> NowPlayingList = new();
+        private static int nowPlayingPointer = -1;
+        private const int PREVIOUS_PLAYING_LIMIT = 3;
```
sonicfind, line 19: Firstly, explicitly construct the list with a value of `3`. That ensures the capacity stays at 3 max - instead of it's normal behavior of growing to 16 on the first insert.

Assets/Script/Menu/Persistent/MusicPlayer.cs
```diff
@@ -57,8 +60,20 @@ private void OnDisable()
...
+                if (NowPlayingList.Count > PREVIOUS_PLAYING_LIMIT)
+                {
+                    nowPlayingPointer--;
+                    NowPlayingList.RemoveAt(0);
+                }
+
+                NowPlayingList.Add(GlobalVariables.Instance.SongContainer.GetRandomSong());
+            }
+            
+            var song = NowPlayingList[nowPlayingPointer];
```
sonicfind, line 76: Secondly, start the index at 0. I never recommend using logic with the index is in an invalid state (outside of enumerators).

With that, this is where the logic flows:
```
if `index == 3`: `list[0] = list[1]; list[1] = list[2]; _nowPlaying = list[2] = random;`
else
{
     if `index == list.count`: `list.add(random);`
     `_nowPlaying = list[index++]`
}
```

Assets/Script/Menu/Persistent/MusicPlayer.cs
```diff
@@ -68,6 +83,14 @@ private async UniTask NextSong()
...
         }
 
+        private void PreviousSong()
+        {
+            nowPlayingPointer -= 2;
+            if (nowPlayingPointer < -1) nowPlayingPointer = -1;
+
+            OnSongEnd();
+        }
+
```
sonicfind, line 93: Then, you just decrement by one and proceed unless index was already 0. Easy to just early return on `index==0`

Assets/Script/Menu/Persistent/MusicPlayer.cs
sonicfind: Additionally, add a private `_nowPlaying` variable to hold the current song instead of indexing into the list over and over.

sonicfind (changes requested):

PR #703, opened 2024-02-27
Fixes DMX not working when enabled at YARG start up.

TheFatBastid:
The callback timing was happening before variables were set, this moves things after, to start() and update()

### DIFF
--- a/Assets/Script/Integration/Sacn/SacnController.cs
+++ b/Assets/Script/Integration/Sacn/SacnController.cs
@@ -33,40 +33,61 @@ namespace YARG.Integration.Sacn
 
         private readonly byte[] _dataPacket = new byte[UNIVERSE_SIZE];
 
-        public void HandleEnabledChanged(bool enabled)
+        private bool _enabled;
+
+        private void Start()
         {
-            if (enabled)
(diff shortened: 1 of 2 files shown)


Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -33,40 +33,61 @@ public class SacnController : MonoSingleton<SacnController>
...
+        private void Start()
+        {
+            //Start and Update are here instead of being a setting callback because that caused problems with when
+            //variables are set.
+            if (!SettingsManager.Settings.DMXEnabled.Value) return;
+            Init();
+            _enabled = true;
+        }
+
+        public void Update()
```
EliteAsian123, line 47: Should be private

EliteAsian123 (changes requested):
Hi, just a few styling changes and this should be good.

Something I remembered while reviewing this was the existence of the `IsLoading` static boolean in `SettingContainer`. Perhaps this can be used to solve the issues with the old way? Return early in `DMXEnabledCallback`, and check manually in `Start` and it *may* fix the problem.

Thanks!

PR #706, opened 2024-02-29
Added DMX Cue Change channel

TheFatBastid:
Whenever the lighting cue changes, a DMX channel will be set to a value. Useful for external programs that make light shows from incoming triggers, such as Light Jamz. Also, this adds the DMX info panel when going to the light settings and fixes a few small bugs with DMX.

This will be the last time the stagekit code will be used like this, next I'll redo how the cues are handled and uncouple that from the stagekit code directly so anything can listen to it in a better way.


TheFatBastid:
Also, this adds the DMX info panel when going to the light settings and fixes a few small bugs with DMX.

Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -30,31 +71,133 @@ public class SacnController : MonoSingleton<SacnController>
...
+        {
+            switch (scene.name)
+            {
+                case "Gameplay":
+                    StageKitGameplay.OnStageEffectChange -= HandleStageEffectChange;
+                    StageKitGameplay.OnLightingTypeChange -= HandleLightingTypeChange;
+                    break;
+
+                default:
+                    break;
```
EliteAsian123, line 153: Use `switch (scene.buildIndex)` instead with the `(int) SceneIndex.Gameplay` for example. Don't use strings when you can avoid it!

Assets/Script/Integration/Sacn/SacnController.cs
```diff
@@ -30,31 +71,133 @@ public class SacnController : MonoSingleton<SacnController>
...
+                case "ScoreScreen":
+                    _cueValue = (byte)CueEnum.Score;
+                    break;
+
+                case "MenuScene":
+                    _cueValue = (byte)CueEnum.Menu;
+                    break;
+
+                default:
+                    break;
```
EliteAsian123, line 175: Same as above

Assets/Script/Integration/StageKit/StageKitGameplay.cs
```diff
@@ -23,6 +24,8 @@ public class StageKitGameplay : GameplayBehaviour
         private int _drumIndex;
         private StageKitLightingController _controller;
 
+        public static event Action<LightingType> OnLightingTypeChange;
+        public static event Action<StageEffect> OnStageEffectChange;
         protected override void OnChartLoaded(SongChart chart)
```
EliteAsian123, line 29: New line between method and field

Assets/Script/Settings/Metadata/PreviewBuilders/DMXInformationPanelBuilder.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public async UniTask BuildPreviewUI(Transform uiContainer)
+        {
+            var go = Object.Instantiate(Information, uiContainer);
+
+            // Enable and wait for layouts to rebuild
+            await UniTask.WaitForEndOfFrame(SettingsMenu.Instance);
+
+            // Skip the game object was somehow destroyed
+            if (go == null) return;
```
EliteAsian123, line 36: These lines can just be removed.
```cs
Object.Instantiate(Information, uiContainer);
return UniTask.CompletedTask;
```

Assets/Script/Settings/Metadata/PreviewBuilders/DMXInformationPanelBuilder.cs
```diff
@@ -0,0 +1,40 @@
...
+using UnityEngine.AddressableAssets;
+using YARG.Core;
+using YARG.Menu.Settings;
+
+namespace YARG.Settings.Metadata
+{
+    public class DMXInformationPanelBuilder : IPreviewBuilder
+    {
+        // Prefabs needed for this tab type
+        private static readonly GameObject Information = Addressables
```
EliteAsian123, line 12: `private` fields should be `_camelCase`

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -219,6 +219,8 @@ public void OpenExecutablePath()
 
             public IntSetting DMXStrobeChannel { get; } = new(07, 1, 512);
 
+            public IntSetting DMXCueChangeChannel { get; } = new(08, 1, 512);
+
```
EliteAsian123, line 223: The leading zeros here are not necessary, they are only in the arrays above for alignment purposes.

Assets/Script/Settings/Types/DMXInformationPanelSetting.cs
EliteAsian123, line 1: What is this file? Is it used? Same with all of the changes below this

EliteAsian123 (changes requested):
Hi, sorry for the late review! Just a few things and this should be good. The `CueEnum` enum is not ideal (something like a dictionary or array would be better so everything doesn't have to be redefined), but you mentioned you'd like to redo it so it's fine for now.

Assets/Script/Settings/Types/DMXInformationPanelSetting.cs
TheFatBastid: Thought I needed that for localization setting to show up. Not needed, will remove.

EliteAsian123:
Did the `OnSceneLoaded` changes get forgotten, or did it not work out?

TheFatBastid:
Oops, I was lumping that in with "well, it works for now will redo with rework", the change has been made.

PR #707, opened 2024-03-01
Improve UX for Go To menu.

vinaykapadia:
For Artist, Album, or Artist-Album, the Go To... list had way too many items. This makes it so that you only get letters in the Go To menu, just like how Song Name works now, and it goes to the first item in the list with that letter.

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/PopupMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/PopupMenu.cs
@@ -1,4 +1,5 @@
-﻿using TMPro;
+﻿using System.Linq;
+using TMPro;
 using UnityEngine;
 using UnityEngine.Events;
 using YARG.Core.Extensions;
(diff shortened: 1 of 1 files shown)


EliteAsian123:
I think it may be better to make this a setting or something. @sonicfind thoughts?

vinaykapadia:
I personally think it works better without a setting, makes the UX more consistent, but I'm not opposed to it either way.

Assets/Script/Menu/MusicLibrary/PopupMenu.cs
```diff
@@ -163,6 +164,23 @@ private void CreateGoToSection()
...
+                    .Select(g => g.First()))
+                {
+                    CreateItem(((SortHeaderViewType) header).HeaderText[0].ToString(), () =>
+                    {
+                        _musicLibrary.SelectedIndex = index;
+                        gameObject.SetActive(false);
+                    });
+                }
+            }
+
```
sonicfind, line 183: Unfortunately, you forgot the `else`. As it is, you'll get `just by first letter` followed by everything coming afterwards anyways.
Also, the grouping is currently case sensitive, so `A` would be in a separate group from `a`.

sonicfind (changes requested):

sonicfind (approved):

PR #711, opened 2024-03-08
YARG-209 Show Bad Songs count on Scan

vinaykapadia:
If there is a bad songs file, it will read the first line and extract the number of errors. A button will be shown with the number of errors. Clicking on the button opens the file.

https://yarg.youtrack.cloud/issue/YARG-209/Show-error-count-after-scanning-songs-if-any-bad-songs-were-encountered

### DIFF
--- a/Assets/Script/Menu/Settings/SongManager/SongManagerHeader.cs
+++ b/Assets/Script/Menu/Settings/SongManager/SongManagerHeader.cs
@@ -9,7 +9,7 @@ namespace YARG.Menu.Settings
     public class SongManagerHeader : MonoBehaviour
     {
         [SerializeField]
-        private ColoredButton badSongsButton;
+        private ColoredButton _badSongsButton;
 
         private void Awake()
         {
@@ -31,14 +31,14 @@ namespace YARG.Menu.Settings
 
         private void CheckBadSongsFile()
         {
-            badSongsButton.gameObject.SetActive(File.Exists(PathHelper.BadSongsPath));
+            _badSongsButton.gameObject.SetActive(File.Exists(PathHelper.BadSongsPath));
             
             var numErrors = CacheHandler.Progress.BadSongCount;
 
             if (numErrors > 0)
             {
                 var errors = numErrors == 1 ? "ERROR" : "ERRORS";
-                badSongsButton.Text.text = $"{numErrors} {errors} FOUND";
+                _badSongsButton.Text.text = $"{numErrors} {errors} FOUND";
             }
         }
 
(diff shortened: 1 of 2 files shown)


Assets/Script/Menu/Settings/SongManager/SongManagerHeader.cs
```diff
@@ -17,5 +29,30 @@ public async void RefreshSongs()
...
+                        var errors = numErrors == 1 ? "ERROR" : "ERRORS";
+                        badSongsButton.Text.text = $"{numErrors} {errors}";
+                    }
+                }
+            }
+        }
+
+        public void OpenBadSongs()
+        {
+            FileExplorerHelper.OpenFolder(PathHelper.BadSongsPath);
```
vinaykapadia, line 55: I used the OpenFolder method, since there is no OpenFile method. It works right on Windows, but I couldn't test the OSX implementation. Someone should probably do that...

EliteAsian123:
Hi, thanks for the PR!

To me, this looks like more of a hack, which we'd like to avoid.

What should happen is the song scanner should return some value that indicates the error(s), and shows a popup after scanning. The popup dialog can have like an “Open badsongs.txt” button for more info or something like that.

FWIW I do think the "Open badsongs.txt" button in the settings menu is a good idea, but it should probably be made to open the actual file instead of opening the folder it's in.

EliteAsian123:
> The button does open the file. The OpenFolder() method doesn't technically need to be a folder, it can take in a file as well. At least on Windows, it would need to be tested on Mac.

Ah okay, I see. According to the documentation, it should work!
image

EliteAsian123 (changes requested):
Forgot to do this on my comment above, whoops

vinaykapadia:
Ok it's no longer hack-y. Didn't realize there was already a place where the bad songs count was stored, I'm reading that from the ChaceHandler now.

EliteAsian123:
Sorry for the late response! A few more things:
image
The “Errors Found” button should follow the same colour scheme as the delete button down there. Use that shade of red with white text,

Upon restarting the game, the button turns into this:
image
The text doesn't fit the button too well. It should probably just state "Errors Found" without the number.

Also `badSongsButton` should be `_badSongsButton`.

I would still prefer a popup right after the scan, but this is fine for now.

Thanks so much, and sorry once again!

vinaykapadia:
> The “Errors Found” button should follow the same colour scheme as the delete button down there. Use that shade of red with white text,

Yeah I'm terribly with visual stuff. Your idea is much better. Will do.

> The text doesn't fit the button too well. It should probably just state "Errors Found" without the number.

Sounds good, will do.

> Also badSongsButton should be _badSongsButton.

Sounds good, will do.

> I would still prefer a popup right after the scan, but this is fine for now.

Yeah a popup would have been nicer, but this is a good v1. Would need a UI person to make a popup though, you really don't want me handling that (see my work with just this button for an example why lol.

vinaykapadia:
Ok, made the changes.

PR #717, opened 2024-03-14
Rework Loading Screen Management

sonicfind:


Assets/Script/Persistent/LoadingScreen.cs
```diff
@@ -0,0 +1,122 @@
...
+
+namespace YARG
+{
+    public class LoadingScreen : MonoSingleton<LoadingScreen>
+    {
+        [SerializeField]
+        public TextMeshProUGUI loadingPhrase;
+
+        [SerializeField]
+        public TextMeshProUGUI subPhrase;
```
EliteAsian123, line 17: Should be pascal case

Assets/Script/Persistent/LoadingScreen.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+
+            // Fast scan (cache read) on startup
+            await SongContainer.RunRefresh(true, context);
+        }
+    }
+
+    public sealed class LoadingContext : IDisposable
+    {
+        private bool disposedValue;
```
EliteAsian123, line 41: Should be `_disposedValue`

EliteAsian123 (changes requested):
Sorry for the late review, I was busy with engine stuff. Just a few naming convention things and this is good to merge. Thanks!

PR #722, opened 2024-03-18
Replace YargTrace with a much more proper logging system

TheNathannator:
Continuation of https://github.com/YARC-Official/YARG.Core/pull/156

## Notes for Future Usage

Copying from the YARG.Core PR since it's an important note.

String interpolation should be avoided whenever possible in future logging code to avoid up-front formatting costs. Use the Format versions of the logging methods, which stores all of the arguments for later and delays formatting until it reaches the logging thread.

Importantly, note the "stores all of the arguments for later" part. If an argument cannot be safely transferred across threads or otherwise stored off of the stack, an up-front formatting will be required for that argument. In particular, `IDisposable` items likely should not be given as an argument.


TheNathannator:
Another note: This PR may cause build failures locally if the changes are pulled before Unity is started, which results in the YARG.Core build buttons being unavailable. Workaround is to either build YARG.Core manually and move the built .dll/.pdb into the Plugins > YARG.Core folder, or ensure that Unity is started before the changes are pulled.

Need to look into why the Plugins > Editor scripts aren't being loaded correctly, they're supposed to still work if compile errors like this happen.

RileyTheFox (approved):

PR #723, opened 2024-03-21
Rework Audio Handling To Allow Song Entires To Directly Create Mixes (Plz test)

sonicfind:
Audio Managers become a factory of mixers. Instead of mapping all control of a mixer to the instance of a manager, you queue managers to create mixers from songs that you then control in your own context. This allows more freedom and more easily ties explicit lifetimes to mixers (instead them being possibly global state).

This also means you can have more than 1 mixer active at a time - which makes scrolling the music library a lot more interesting.

The structure style utilized strictly enforces thread-safe behavior (and believe me, I tried other ways to get the same result).

+ Requires Core update before any sort of merge,


EliteAsian123:
1. When I first start a song in practice, I get these errors (not sure if it's connected to the current practice bug, if so ignore):
image

2. It appears that songs with only a guitar stem will mute on miss, when it doesn't on the dev branch.

Other than that, haven't noticed any other problems from my limited testing. Thanks!

sonicfind:
1. yea, that's the practice bug. Bass will throw that if the speed value is out of range (which for the practice bug, was obviously quite large).
2. That should be handled now. Though mute on miss will still activate to a degree, it'll be hard limited to a 50% minimum (applied on top of the stem's settings manager volume).

sonicfind:
Now truly in the testing stage. Please go hard on this shit to find any behaviors I missed or small things to change.

Note: calibrations may require adjustments for this update. However, at least for me, I got a lower value.

EliteAsian123:
I've been testing this a little bit for the past couple of days and I haven't noticed any issues so far! ...except for one small one

If you spam the skip button on the main menu music player, multiple songs start playing at the same time, and they don't stop playing even if you go into a song.

sonicfind:
ah ok. lemme look into that

EliteAsian123:
Oop something else: if you click the little mute button on the Unity editor while the game is playing the state doesn't update. It does work fine if you mute/unmute before you go into the game though.

EliteAsian123:
Not sure if you know this, apologies if you do, but when the song ends, it freezes for a few seconds and adds like 500 high score entries:
image

sonicfind:
Hopefully addressed that. Don't personally see anything else weird in that regard.

sonicfind:
Nevermind. Still some things to figure out

sonicfind:
Please, God. Let that be it. lol

PR #736, opened 2024-04-07
Fixed errant duplicate Add bug 

TheFatBastid:
Event handlers aren't added/removed in the order requested so having two lists that first do Add then Remove causes anomalies. Changing to one list fixes that.

### DIFF
--- a/Assets/Script/Gameplay/GameManager.cs
+++ b/Assets/Script/Gameplay/GameManager.cs
@@ -282,6 +282,7 @@ namespace YARG.Gameplay
                     text.AppendFormat("Note index: {0}\n", state.NoteIndex);
                 }
 
+                text.AppendFormat("Device audio latency: {0}ms\n", GlobalAudioHandler.PlaybackLatency);
                 text.AppendFormat("Song time: {0:0.000000}\n", _songRunner.SongTime);
                 text.AppendFormat("Audio time: {0:0.000000}\n", _songRunner.AudioTime);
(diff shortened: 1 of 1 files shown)


Assets/Script/Playback/BeatEventHandler.cs
```diff
@@ -175,8 +172,9 @@ public void Update(double songTime, SyncTrack sync)
         private readonly Dictionary<Delegate, IBeatAction> _states = new();
 
         // Necessary to allow adding or removing subscriptions within event handlers
-        private readonly List<(Delegate, IBeatAction)> _addStates = new();
-        private readonly List<Delegate> _removeStates = new();
+        private readonly List<(bool, Delegate, IBeatAction)> _changeStates = new();
+        //private readonly List<(Delegate, IBeatAction)> _addStates = new();
+        //private readonly List<Delegate> _removeStates = new();
```
EliteAsian123, line 177: All of the old stuff can just be removed instead of commented out

Assets/Script/Playback/BeatEventHandler.cs
```diff
@@ -153,8 +151,7 @@ public void Update(double songTime, SyncTrack sync)
             {
                 // Apply offset up-front
                 songTime -= _offset;
-                if (songTime < 0)
-                    return;
+                if (songTime < 0) return;
```
EliteAsian123, line 154: For this change (and all of the above), just add braces instead

EliteAsian123 (commented):
Few minor styling things, and also it looks like you forgot to update your dev branch before making this, so there are some faulty changes in this PR

PR #739, opened 2024-04-09
Fixed venue count on song restart

TheFatBastid:
Added count to debug overlay. formatting.

### DIFF
--- a/Assets/Script/Integration/MasterLightingGameplayMonitor.cs
+++ b/Assets/Script/Integration/MasterLightingGameplayMonitor.cs
@@ -11,8 +11,8 @@ namespace YARG
 {
     public class MasterLightingGameplayMonitor : GameplayBehaviour
     {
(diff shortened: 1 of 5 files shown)


Assets/Script/Integration/MasterLightingGameplayMonitor.cs
```diff
@@ -11,24 +11,32 @@ namespace YARG
 {
     public class MasterLightingGameplayMonitor : GameplayBehaviour
     {
-        private VenueTrack _venue;
+        public static VenueTrack _venue;
+        public static int _lightingIndex;
```
EliteAsian123, line 15: These should probably be `{ get; private set; }` properties with `PascalCase` names

EliteAsian123 (changes requested):

EliteAsian123 (approved):

PR #743, opened 2024-04-21
RB3E limited data stream implemention

TheFatBastid:
This adds an opt-in data stream subset of the one used in RB3E, this one is only for stage kit cue calls so that hardware like StageKitPied can be used.

This also includes the new setting prefab "IPv4" to set IP addresses


Assets/Script/Integration/RB3E/RB3EHardware.cs
```diff
@@ -0,0 +1,161 @@
...
+                KillRB3E();
+            }
+        }
+
+        public void HandleBroadcastIPChanged()
+        {
+            _ipAddress = new IPAddress(SettingsManager.Settings.RB3EBroadcastIP.Value);
+        }
+
+        private void KillRB3E()
```
EliteAsian123, line 65: The send client should probably be closed when the game does just to be safe. Putting this in `OnApplicationQuit` should be good.`

EliteAsian123 (changes requested):
Looks good to be pretty much, once the conflict is fixed this should be good to merge. Thanks!

PR #744, opened 2024-04-21
separate labels for 5-lane drum bindings

W-Livi:
5-lane drums bindings previously reused the 4-lane localization strings, meaning the numbers were weird and Orange Cymbal was just blank. giving 5L its own strings means we can label orange and uniquely number all five lanes.


TheNathannator (approved):
I'd suggest prefixing the 4-lane labels as well to match, but otherwise this should be good!

PR #746, opened 2024-04-24
Setting to Auto Connect Profile

vinaykapadia:
Not sure if this is helpful to everyone, but it does help me. This will attempt to automatically connect to the first profile in the list. Saves having to select it manually, and allows for control without mouse.

There is a setting to turn this on or off.

### DIFF
--- a/Assets/Script/Persistent/LoadingScreen.cs
+++ b/Assets/Script/Persistent/LoadingScreen.cs
@@ -1,11 +1,15 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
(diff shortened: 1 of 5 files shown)


kaduwaengertner:
My main problem with this is that my main profile isnt the first one lol, maybe an option in the profile to set that to auto connect? so any profile can be set, if lets say Ive been playing drums more lately I can set my drum profile, or guitar etc

vinaykapadia:
> My main problem with this is that my main profile isnt the first one lol, maybe an option in the profile to set that to auto connect? so any profile can be set, if lets say Ive been playing drums more lately I can set my drum profile, or guitar etc

I did consider that. Probably a better idea. I'll do that.

PR #750, opened 2024-04-25
Allow Profile to Auto Connect.

vinaykapadia:
Each profile has a toggle that can tell YARG to attempt to auto connect it on load. If it fails for any reason, a toast is shown.

This is not intended to be a final solution, when we revamp the whole profile connect system this will be dropped.

I haven't been able to test this with the drum e-kit thing that needs a button press or something before YARG can pick it up, since I don't have one. Works great with keyboard input and my xbox one guitar input (via rb4instrumentmapper) though.


EliteAsian123:
I think it may be better if it just tries to reconnect the profiles that we last used when closing the game instead of adding a checkbox. It makes it a bit more convenient that way, and also more user-friendly. What do you think?

vinaykapadia:
Cool I can do that.

vinaykapadia:
@EliteAsian123 ok done, it now connects to whichever profile was last connected. There is a setting to allow this or not.

EliteAsian123:
If you could fix the conflicts by merging with `dev`, that'd be great.

Also, it looks like there aren't any translations pushed for the new setting.

vinaykapadia:
@EliteAsian123 done and done!

PR #751, opened 2024-04-26
Status Bar Improvements

vinaykapadia:
Moves Status information to the right and Build information to the left. Adds Time and Battery info.

I set the battery thresholds at 10, 40, and 90% for now. They're set as constants so they're easy to tweak any time.


vinaykapadia:
Don't merge yet. Adding toggles for Time and Battery display.

vinaykapadia:
Ok, toggles added. Can be merged now.

vinaykapadia:
Added an option to show the memory stats, may be interesting to some people.

Also organized it so all status bar options are available in one place:
image

The FPS Counter option is still available under Graphics settings as well.

vinaykapadia:
Here's how it looks with all options enabled:
image

vinaykapadia:
Added the memory icon. Here is an updated screen shot (without battery showing):
image

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -62,6 +62,10 @@ public void OpenVenueFolder()
             public ToggleSetting DisableGlobalBackgrounds  { get; } = new(false);
             public ToggleSetting DisablePerSongBackgrounds { get; } = new(false);
 
+            public ToggleSetting ShowBattery { get; } = new(false);
```
EliteAsian123, line 65: Align the `{ get; }` with the other ones to make it look neater.

EliteAsian123 (changes requested):
Very nice, love it! Just a few minor tweaks and this should be good to go:

The spacing between the left and the right content is not the same if the battery icon isn't shown:
image

It may be worth combining all of these new sprites into a singular sprite sheet (see `Art/Menu/Common/Icons` for some examples). Not needed, but it would tidy things up a bit.

Assets/Script/Settings/SettingsManager.Settings.cs
vinaykapadia: Cool will do both.

vinaykapadia:
@EliteAsian123 done.

PR #752, opened 2024-04-26
Add fallback font for fullwidth characters (code page FF01-FFE3)

rk-exxec:
Setlist with special characters in the title are not rendered properly.   
They show up as boxes:
image

From this playlist:
https://customsongscentral.com/club_mix-3/

image

This text is using Unicode characters with Codes FF01-FFE3 (Fullwidth https://unicode-explorer.com/b/FF00 ).
Liberation Sans does not support this range, so I added a new open source font [Noto Sans from Google Fonts](https://fonts.google.com/noto/specimen/Noto+Sans) and built the atlas texture for that block specifically.

This fixes the issue.

image

The font is licensed with the open font license. https://openfontlicense.org/

> If you bundle a font under the OFL with your mobile app you must comply with the terms of the license. At a minimum you must include the copyright statement, the license notice and the license text. A mention of this information in your About box or Changelog, with a link to where the font package is from, is good practice, and the extra space needed to carry these items is very small. You do not, however, need to include the full contents of the font package - only the fonts you use and the copyright and license that apply to them. For example, if you only use the regular weight in your app, you do not need to include the italic and bold versions.

### DIFF
--- a/Assets/TextMesh Pro/Resources/Fonts & Materials/LiberationSans SDF.asset
+++ b/Assets/TextMesh Pro/Resources/Fonts & Materials/LiberationSans SDF.asset
@@ -7714,6 +7714,7 @@ MonoBehaviour:
   fallbackFontAssets: []
   m_FallbackFontAssetTable:
   - {fileID: 11400000, guid: 2e498d1c8094910479dc3e1b768306a4, type: 2}
+  - {fileID: 11400000, guid: bcb9bdf7d6ecff043a3921396b481c87, type: 2}
   m_CreationSettings:
     sourceFontFileName: 
     sourceFontFileGUID: e3265ab4bf004d28a9537516768c1c75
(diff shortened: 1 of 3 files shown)


EliteAsian123 (changes requested):
Hi, thanks for the PR! New fonts should be added to `Art/Fonts` instead, as `TextMesh Pro` should only contain default assets from TMP. Along with this, if the original `.ttf` file could be included so it can be updated in the future if needed, that'd be great (it just shows up as missing right now).

Thanks!

image

rk-exxec:
Ah sure thing, I also removed the fallback from the liberation sans and added it to the barlow font. Additionally I added all the different font weights as well. I hope this is now correct.

EliteAsian123:
Hello, thanks so much!

From the looks of it, it doesn't look like you actually added the fallback settings in 667ccb1f7b94c3a3cc02825d4b147a45b29976dd (unless I'm dumb and I'm just seeing it).

Also, if you could do me a favour, and track `.ttf` files with LFS since the Japanese ones are quite large? This can be done easily by just running `git lfs track "*.ttf"`. I did just make some LFS changes so I recommend you update this branch from `dev` before doing this.

rk-exxec:
I thought I did, sorry. My git client didn't show any other changes, 
I pushed them now!

EliteAsian123 (approved):

EliteAsian123:
No problem, thanks!

PR #754, opened 2024-04-29
Wider text field for playlist category header text

### LINKED ISSUE #753: Longer section titles in music menu
Hey, I was wondering if you could allow for longer section titles to be viewed properly.
Right now they are cut short relatively early.

image

I tinkered a bit with the SongView Prefab, to maybe find an easy solution but it didn't work out that well:
I added flexible width for the primary text and minimum text width for the other elements, so these fields keep their original width.  
When used as section header, the other fields not enabled and the primary text can expand to whole width, allowing for longer playlist titles.
This comes at the cost, that the artist is now more to the right, as you can see here.

image

rk-exxec:
for your long playlist name needs, fixes #753
image

- Added a new group in the SongView prefab with a second primary text as suggested by EliteAsian
- The swap happens depending on which background is chosen

Feedback is appreciated. Not quite familiar with the codebase layout yet.

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/SongView.cs
+++ b/Assets/Script/Menu/MusicLibrary/SongView.cs
@@ -46,7 +46,7 @@ namespace YARG.Menu.MusicLibrary
             base.Show(selected, viewType);
 
             // if used as section header, use different primary text for wider text field
-            if (viewType.Background == BaseViewType.BackgroundType.Category)
+            if(viewType.GetType() == typeof(SortHeaderViewType))
             {
                 _songNameContainer.SetActive(false);
                 _categoryNameContainer.SetActive(true);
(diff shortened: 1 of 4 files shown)


EliteAsian123:
Instead, perhaps the `ViewType` should have a virtual property called something like `UseWidePrimaryText`, which is checked for in  `Show`.

rk-exxec:
Right, thats a better idea.

Assets/Script/Menu/MusicLibrary/SongView.cs
```diff
@@ -33,10 +35,29 @@ public class SongView : ViewObject<ViewType>
...
         public override void Show(bool selected, ViewType viewType)
         {
             base.Show(selected, viewType);
 
+            // use different primary text for wider text field, e.g. when used as section header
+            if(viewType.UseWiderPrimaryText)
+            {
+                _songNameContainer.SetActive(false);
+                _categoryNameContainer.SetActive(true);
+                _categoryText.text = viewType.GetPrimaryText(selected);
```
EliteAsian123, line 53: Instead of doing this, the `_categoryText` can just be added here and it should handle everything for you:
image

EliteAsian123 (changes requested):
One more thing and this should be good to go!

rk-exxec:
Done!

Issue #755, opened 2024-04-30
Feature Request: Library Navigation Suggestion (Jumping through headers)


Angstdiscord:
Navigating large libraries is kinda slow.

When sorting by something like artist, it would be nice if you could quickly jump between the different artist headers.

This way I could have an artist with a large amount of songs but skip that artist with a single tap.

Angstdiscord:
Actually, seems like this feature already exists here 
https://docs.yarg.in/en/songs

However I don't have a yellow fret option in the menu here. nor does my yarg give me the option to hold the orange fret to get anything other than the other options menu

poliwhirlge:
> Actually, seems like this feature already exists here https://docs.yarg.in/en/songs
> 
> However I don't have a yellow fret option in the menu here. nor does my yarg give me the option to hold the orange fret to get anything other than the other options menu

The screenshot on that page is either old or a concept, right now to go between artists you can hit orange to go to the menu, then select "go to section" and scroll down the list. Menus/navigation aren't final and improvements are planned.

Angstdiscord:
Yeah, it looks like this feature existed in version 11., but was changed in 12. I see newer wip builds on twitter that seemingly brought the feature back, so I'll be patient and wait for the update. Thank you for the response.

Angstdiscord:
This was fixed in #756

PR #756, opened 2024-05-02
Re-implement holding down the orange fret for section navigation in MusicLibrary

poliwhirlge:
Hold orange + strum up and down to go to next/previous sections, seems like a capability that was in a previous version as discussed here: https://github.com/YARC-Official/YARG/issues/755

https://github.com/YARC-Official/YARG/assets/141366895/3a81e04a-73d9-4605-a4fb-eb02ac788916

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -5,10 +5,8 @@ using System.Threading;
 using Cysharp.Threading.Tasks;
 using TMPro;
 using UnityEngine;
-using YARG.Audio;
 using YARG.Core.Audio;
 using YARG.Core.Input;
-using YARG.Core.Logging;
 using YARG.Core.Song;
 using YARG.Menu.ListMenu;
 using YARG.Menu.Navigation;
@@ -82,6 +80,7 @@ namespace YARG.Menu.MusicLibrary
         private SongEntry _currentSong;
 
         private List<int> _sectionHeaderIndices = new();
+        private List<HoldContext> _heldInputs = new();
 
         protected override void Awake()
         {
@@ -91,27 +90,22 @@ namespace YARG.Menu.MusicLibrary
             _sidebar.Initialize(this, _searchField);
         }
 
-        private List<HoldContext> _heldInputs = new();
-
-        private bool IsButtonHeldByPlayer(YargPlayer player, MenuAction button)
-        {
-            return _heldInputs.Any(i => i.Context.Player == player && i.Context.Action == button);
-        }
-
         private void OnEnable()
         {
             // Set navigation scheme
             Navigator.Instance.PushScheme(new NavigationScheme(new()
             {
                 new NavigationScheme.Entry(MenuAction.Up, "Up",
-                    (NavigationContext ctx) => {
+                    ctx =>
+                    {
                         if (IsButtonHeldByPlayer(ctx.Player, MenuAction.Orange))
                             GoToPreviousSection();
                         else
                             SelectedIndex--;
                     }),
                 new NavigationScheme.Entry(MenuAction.Down, "Down",
-                    (NavigationContext ctx) => {
+                    ctx =>
+                    {
                         if (IsButtonHeldByPlayer(ctx.Player, MenuAction.Orange))
                             GoToNextSection();
                         else
@@ -442,8 +436,8 @@ namespace YARG.Menu.MusicLibrary
 
         protected override void Update()
         {
-            foreach (var _heldInput in _heldInputs)
-                _heldInput.Timer -= Time.unscaledDeltaTime;
+            foreach (var heldInput in _heldInputs)
+                heldInput.Timer -= Time.unscaledDeltaTime;
 
             base.Update();
         }
@@ -513,13 +507,17 @@ namespace YARG.Menu.MusicLibrary
 
         private void OnButtonRelease(NavigationContext ctx)
         {
-            var holdContext = _heldInputs.Where(i => i.Context.IsSameAs(ctx));
-            YargLogger.AssertFormat(holdContext.Count() == 1, "Unexpected HoldContext.Count in MusicLibrary: {0}", holdContext.Count());
+            var holdContext = _heldInputs.FirstOrDefault(i => i.Context.IsSameAs(ctx));
 
-            if (ctx.Action == MenuAction.Orange && holdContext.FirstOrDefault()?.Timer > 0)
+            if (ctx.Action == MenuAction.Orange && holdContext?.Timer > 0)
                 _popupMenu.gameObject.SetActive(true);
 
             _heldInputs.RemoveAll(i => i.Context.IsSameAs(ctx));
+        }
+
+        private bool IsButtonHeldByPlayer(YargPlayer player, MenuAction button)
+        {
+            return _heldInputs.Any(i => i.Context.Player == player && i.Context.Action == button);
         }
 
         private void GoToNextSection()
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -16,7 +16,7 @@ using YARG.Player;
 using YARG.Playlists;
 using YARG.Settings;
 using YARG.Song;
-
+using static YARG.Menu.Navigation.Navigator;
 using Random = UnityEngine.Random;
 
 namespace YARG.Menu.MusicLibrary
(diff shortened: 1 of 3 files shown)


Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -317,9 +341,17 @@ private List<ViewType> CreatePlaylistViewList()
...
             }
 
+            CalculateCategoryHeaderIndices(list);
             return list;
         }
 
+        private void CalculateCategoryHeaderIndices(List<ViewType> list)
+        {
+            _sectionHeaderIndices = list.Select((v, i) => (v, i)).
+                Where(viewTypeEntry => viewTypeEntry.v is SortHeaderViewType || viewTypeEntry.v is CategoryViewType).
```
EliteAsian123, line 351: Can be simplified to `viewTypeEntry.v is SortHeaderViewType or CategoryViewType`

Assets/Script/Menu/Navigation/NavigationScheme.cs
```diff
@@ -48,6 +51,11 @@ public void Invoke(NavigationContext context)
             {
                 _handler?.Invoke(context);
             }
+
+            public void InvokeHolddOffHandler(NavigationContext context)
```
EliteAsian123, line 55: Typo

Assets/Script/Menu/Navigation/NavigationScheme.cs
```diff
@@ -27,19 +27,22 @@ public class NavigationScheme
...
+            public Entry(MenuAction action, string displayName, Action handler, Action onHoldOffHandler = null)
             {
                 Action = action;
                 DisplayName = displayName;
                 _handler = _ => handler?.Invoke();
+                _onHoldOffHandler = _ => onHoldOffHandler?.Invoke();
             }
 
-            public Entry(MenuAction action, string displayName, Action<NavigationContext> handler)
+            public Entry(MenuAction action, string displayName, Action<NavigationContext> handler, Action<NavigationContext> onHoldOffHandler = null)
```
EliteAsian123, line 40: Thoughts on, instead of doing this, add an extra field to `NavigationContext` that contains the hold state? This way, a new action is not needed in each entry, and it can consolidate the action in the `MusicLibrary`.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -474,6 +506,43 @@ private void Back()
             MenuManager.Instance.PopMenu();
         }
 
+        private void OnOrangeFretHit(NavigationContext ctx)
+        {
+            if (ctx.IsRepeat)
```
EliteAsian123, line 511: Using the orange fret repeat is kind of hacky. It should be a separate timer with a separate threshold within this class instead, as now *all* orange frets on all menus repeat.

EliteAsian123 (changes requested):
Thanks for the PR! A few adjustments and this should be good to go. Sorry for the wait!

Assets/Script/Menu/Navigation/NavigationScheme.cs
poliwhirlge: The second Action callback is more to get access to an "on button up" event, as the first Action is fired on "on button down". The way it's been working is as soon as you hit the orange fret, the handler Action fires and the more options menu immediately opens, and I wanted to move that to when the fret is released instead (otherwise the whole hold fret + strum up and down wouldn't be possible).

It's possible to add a field to NavigationContext that indicates whether it's a button on or button off event, but wouldn't that mean that we'd have to change all our other NavigationScheme.Entry objects to not do anything on the button off?

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
poliwhirlge: Yeah agree that it would be a good idea to just localize it to MusicLibrary, I just don't know if it's possible to know if the fret is being held down from MusicLibrary because I previously I believe the Navigator singleton only fires the event once unless we let the orange fret event repeat. I guess if we let the MusicLibrary listen directly to the InputManager we can have let it have its own timer within but that seems like breaking abstraction even more.

I'll keep thinking about ways to make this cleaner, thanks for the feedback.

Assets/Script/Menu/Navigation/NavigationScheme.cs
EliteAsian123: Fair

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
EliteAsian123: I'm pretty sure that won't be needed. If you track when it gets first held down, you can update the orange button state within the update method with a check. I'm on mobile so apologies if I'm missing something.

poliwhirlge: Yeah makes sense, that would work, I'm not sure how to check if the orange button is held down though since at that point we don't have the GameInput object. Would it be Unity's Input.GetKey? Not seeing that used anywhere else in the code base though.

EliteAsian123: You should be able to just do it off of the press and the new release actions you made.

In the press method, start a timer.

In the release method, if the timer has been active for X seconds, stop holding, otherwise open the more options menu.

poliwhirlge: Duh, yeah of course. Thanks!

poliwhirlge:
Made the changes (just re-used the HoldContext class), it also filters by player now so player A holding down the fret won't affect B's scrolling..

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -413,6 +442,9 @@ private void UpdateSearch(bool force, bool refresh)
 
         protected override void Update()
         {
+            foreach (var _heldInput in _heldInputs)
```
EliteAsian123, line 445: Only private fields should be prefixed with an underscore. Local variables should just be `heldInput`

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -474,6 +506,42 @@ private void Back()
...
         }
 
+        private void OnButtonHit(NavigationContext ctx)
+        {
+            _heldInputs.Add(new HoldContext(ctx));
+        }
+
+        private void OnButtonRelease(NavigationContext ctx)
+        {
+            var holdContext = _heldInputs.Where(i => i.Context.IsSameAs(ctx));
```
EliteAsian123, line 516: It may be worth doing this instead:
```cs
var holdContext = _heldInputs.FirstOrDefault(i => i.Context.IsSameAs(ctx));
```
This way you don't have to worry about the possibility of having multiple hold contexts, and you avoid potential multiple enumerations.

In the case where there are multiple hold contexts for whatever reason, they should be dealt with by the `RemoveAll`.

EliteAsian123 (changes requested):

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -88,23 +91,40 @@ protected override void Awake()
...
+        }
+
         private void OnEnable()
         {
             // Set navigation scheme
             Navigator.Instance.PushScheme(new NavigationScheme(new()
             {
                 new NavigationScheme.Entry(MenuAction.Up, "Up",
-                    () => SelectedIndex--),
+                    (NavigationContext ctx) => {
```
EliteAsian123, line 107: The `NavigationContext` type is not needed here. Just do `(ctx) => {` or `ctx => {`.
Also, braces should be on the next line.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -88,23 +91,40 @@ protected override void Awake()
             _sidebar.Initialize(this, _searchField);
         }
 
+        private List<HoldContext> _heldInputs = new();
+
+        private bool IsButtonHeldByPlayer(YargPlayer player, MenuAction button)
+        {
+            return _heldInputs.Any(i => i.Context.Player == player && i.Context.Action == button);
+        }
```
EliteAsian123, line 99: This method should be placed near `OnButtonHit`, instead of at the top.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -88,23 +91,40 @@ protected override void Awake()
             _sidebar.Initialize(this, _searchField);
         }
 
+        private List<HoldContext> _heldInputs = new();
```
EliteAsian123, line 94: This field should be placed with the other fields.

EliteAsian123 (changes requested):

poliwhirlge:
Cool, think I got everything, thanks for the reviews!

EliteAsian123:
Thanks!

PR #757, opened 2024-05-02
update 4L Cymbal binding strings to account for Lefty Flip

W-Livi:
Renumbered four-lane cymbals to match their lanes, expanded Green Cymbal string to encompass its role as Red Cymbal

### DIFF
--- a/Assets/Script/Input/Bindings/ControlBinding.cs
+++ b/Assets/Script/Input/Bindings/ControlBinding.cs
@@ -75,11 +75,11 @@ namespace YARG.Input
             Action = action;
         }
 		
-        public ControlBinding(string name, string namelefty, int action)
+        public ControlBinding(string name, string nameLefty, int action)
         {
             Key = name;
             Name = new("Bindings", name);
-            NameLefty = new("Bindings", namelefty);
+            NameLefty = new("Bindings", nameLefty);
             Action = action;
         }
 
@@ -224,7 +224,7 @@ namespace YARG.Input
         {
         }
 
-        public ControlBinding(string name, string namelefty, int action) : base(name, namelefty, action)
+        public ControlBinding(string name, string nameLefty, int action) : base(name, nameLefty, action)
         {
         }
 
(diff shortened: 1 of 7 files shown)


EliteAsian123:
image
I think having this displayed is a little bit clunky, so perhaps it should swap the displayed text when left flip is turned on

W-Livi:
Thank you for the constructive criticism; I agree that just mashing the two together was kinda messy. This new commit adopts your suggested approach.

The new result is still a _little_ awkward, in that a lefty player sees
- Yellow Cymbal (Second)
- Blue Cymbal (Third)
- Red Cymbal (First)

Perhaps I should look into some means of specifying an alternate sequence for lefty binds, to go alongside the alternate labels.

Assets/Script/Input/Bindings/ControlBinding.cs
```diff
@@ -66,6 +71,15 @@ public ControlBinding(string name, int action)
         {
             Key = name;
             Name = new("Bindings", name);
+            NameLefty = Name;
+            Action = action;
+        }
+		
+        public ControlBinding(string name, string namelefty, int action)
```
EliteAsian123, line 78: The argument should probably be named `nameLefty` to adhere to naming conventions.

Assets/Script/Menu/ProfileInfo/EditBinds/BindGroups/BindHeader.cs
```diff
@@ -38,7 +38,11 @@ public void Init(EditBindsTab editBindsTab, YargPlayer player, ControlBinding bi
             _player = player;
             _binding = binding;
 
-            _bindingNameText.StringReference = binding.Name;
+            if (player.Profile.LeftyFlip) {
```
EliteAsian123, line 41: Make sure to follow the proper brace styling (new line before brace).

EliteAsian123 (changes requested):

Assets/Script/Menu/ProfileInfo/EditBinds/BindGroups/BindHeader.cs
W-Livi: Thank you for pointing this out; I'm not sure how I missed the style y'all use, especially since it was already on clear display - `else` included - in this same method.

PR #759, opened 2024-05-03
sACN rewrite and expansion

TheFatBastid:
As I was asking the sACN Hardware class to do more and more, it was clear it needed its own interpreter. This makes the hardware class more pure as it no longer manipulates the DMX packet, only sends it. While the interpreter has grown to both mimic the Stage Kit and has a new 'pass through' mode that sends much more data in the DMX data packet so that external programs such as Lightjams can be used to create truly custom light shows.

This meant the master controller and master gameplay monitor changed in what info they send out and how they monitor things, as well.

Added a Universe setting.

Also, fixed a typo.


Assets/Script/Integration/MasterLightingController.cs
```diff
@@ -136,23 +203,33 @@ public static bool LargeVenue
...
         public static event Action<bool> OnLargeVenue;
         public static event Action<FogState> OnFogState;
-        public static event Action<DrumNote> OnDrumEvent;
+        public static event Action<int> OnDrumEvent;
+        public static event Action<int> OnGuitarEvent;
+        public static event Action<int> OnBassEvent;
         public static event Action<VocalNote> OnVocalsEvent;
         public static event Action<Beatline> OnBeatLineEvent;
         public static event Action<LightingEvent> OnLightingEvent;
         public static event Action<StageKitStrobeSpeed> OnStrobeEvent;
```
EliteAsian123, line 212: Is there a way to consolidate all or most of these events? Having an enum indicating the type attached with the action, along with a time or something may suffice.

EliteAsian123 (changes requested):

Assets/Script/Integration/MasterLightingGameplayMonitor.cs
```diff
@@ -10,16 +10,56 @@ namespace YARG.Integration
...
+            RedDrum = 1,
+            YellowDrum = 2,
+            BlueDrum = 4,
+            GreenDrum = 8,
+            Kick = 16,
+            YellowCymbal = 32,
+            BlueCymbal = 64,
+            GreenCymbal = 128,
+        }
+
```
EliteAsian123, line 37: Flag enums should always have the `[Flag]` attribute and should use bitshifts to denote the values for clarity and maintainability. For example:
```cs
[Flags]
public enum BackgroundType
{
    Yarground = 1 << 0,
    Video = 1 << 1,
    Image = 1 << 2,
}
```

**However, read the suggestions below of why these enums aren't needed.**

Assets/Script/Integration/MasterLightingGameplayMonitor.cs
```diff
@@ -45,12 +94,332 @@ private void Update()
...
+                        case (int) FourLaneDrumPad.GreenCymbal:
+                            padsHit += (int) DrumEnum.GreenCymbal;
+                            break;
+                    }
+
+                    MasterLightingController.CurrentDrumNotes = padsHit;
+                    _drumEndCheckIndex = _drumIndex;
+                    _drumIndex++;
+                }
+            }
```
EliteAsian123, line 413: The vast majority of this is extremely repetitive. 

Note that all usages of `FiveFretsEnum` and `DrumEnum` can be simplified to an int, since they are all casted to an int. Because of this, the majority of the switch statements can just be further simplified to something like:
```cs
fretsPressed += 1 << c.Fret;
```
The problem with this is that then, the flags don't match the values they were defined as above. However, from a quick look at the code (I might be wrong), this doesn't actually matter as there are checks on hardcoded values:
```cs
public override void HandleDrumEvent(int eventName)
{
    if (!_enabled)
    {
        return;
    }

    if ((_listenType & ListenTypes.RedFretDrums) == 0 || (eventName & 1) == 0)
    {
        return;
    }

    ProcessEvent();
}
```
The `1` here can just be replaced with `FourLaneDrumPad.RedDrum` which is both more clear and maintainable.

Assets/Script/Integration/Sacn/SacnHardware.cs
```diff
@@ -1,48 +1,15 @@
...
-            Sweep = 180,
-            BlackoutFast = 190,
-            BlackoutSlow = 200,
-            BlackoutSpotlight = 210,
-            FlareSlow = 220,
-            FlareFast = 230,
-            BigRockEnding = 240,
-            BonusEffect = 250,
-        }
+        Queue<byte>[] _commandQueue = new Queue<byte>[512];
```
EliteAsian123, line 12: This is very much not good memory-wise. I haven't looked deeply into how this is used, however, doing something like `Queue<(byte DataPacketIndex, byte Value)>` should be much better.

EliteAsian123 (changes requested):
Hi, sorry for the wait!

I did a quick review of some of the biggest problems I've noticed that should be fixed. After this, I'll do a more in-depth review.

Thanks!

Assets/Script/Integration/Sacn/SacnHardware.cs
EliteAsian123: Is there a reason why an array of queues, or a queue in general is needed? If two things are added to within one `TIME_BETWEEN_CALLS`, pointing at the same data packet index, is it supposed to wait for the next `Sender` invoke or override the value?

Assets/Script/Integration/MasterLightingGameplayMonitor.cs
```diff
@@ -45,12 +71,120 @@ private void Update()
...
+                while (_keysIndex < _keys.Notes.Count && _keys.Notes[_keysIndex].Time <= GameManager.SongTime)
+                {
+                    int fretsPressed = 0;
+
+                    foreach (var c in _keys.Notes[_keysIndex].ChildNotes)
+                    {
+                        fretsPressed += 1 << c.Fret;
+                    }
+
+                    fretsPressed += 1 << _keys.Notes[_keysIndex].Fret;
```
EliteAsian123, line 86: `.ChordEnumerator()` can be used instead of `.ChildNotes` to prevent having to duplicate logic below for the parent note.

EliteAsian123 (changes requested):
One small change and this should be good.

Overall, I still believe there is a lot of code duplication for each instrument that can be generalized into a class or something, however, this PR has been open for quite a while now so it's good enough *for now*.

Thanks!

Issue #760, opened 2024-05-03
Font issues with lyric text (missing glyphs)


rk-exxec:
Hey, me again with another font issue on the dev branch.

I found that the font used in the lyric bar has the extended ascii glyphs missing (such as öäü).
I also noticed, that its font atlas is set to static. (Red Hat Display Bold)
While other RHD weights have a dynamic atlas enabled.

What would be the best course of action here? 
Add the missing glyphs to the font atlas or set it to dynamic?

Also, the Barlow font has a mix of static and dynamic as well.
NotoSans is purely Dynamic.

Should these be unified or is there a performance / memory concern here?

EliteAsian123:
Dynamic fonts have the problem of increasing the build size by a very significant amount (and due to GitHub's size limits, we gotta keep the sizes below 2GB). 

Which reminds me, here are the build sizes before the addition of NotoSans:
![image](https://github.com/YARC-Official/YARG/assets/29520859/c9e17366-1242-49fc-ad08-54f3e09fb384)
And here is after:
![image](https://github.com/YARC-Official/YARG/assets/29520859/ea67100c-271d-4795-b091-beee8b2374a8)

So yeah, quite a significant increase after those additions.

Because of this, I am going to say the best course of action is to:
* Set all main fonts to use static atlases. We should, however, use a slightly wider range of characters to allow them to display. We obviously can't have *all* characters, as then that would defeat the purpose, but as said a slightly wider range should be enough.
* NotoSans should be kept dynamic. Backup fonts will be slower on the system, but since they are backup fonts they are less likely to be seen. Because of this, however, we should also reduce the amount of variations of NotoSans we are using. Using Regular along with Regular Italic should be enough (or a thinner variant, whatever is more suitable for the general use case). This should reduce the build size while allowing more characters to be available.

PR #762, opened 2024-05-04
Update fonts to static atlas and remove all but 2 fallback NotoSans fonts with dynamic atlas

rk-exxec:
This is to reduce the build size while keeping reasonable compatibility with other languages. (#760)

Static fonts now use this unicode char range : 
20-7E,A0-FF,2010,2013-2014,2018-201A,201C-201E,2020-2022,2026,2030,2032-2033,2039-203A,2044,20AC,2122

NotoSansJP Bold as fallback for all bold-ish font weights including italic, since NSJP has no italic variant.
NotoSansJP Regular for all others, including italic.

A note: NotoSans JP has a variable font weight file, which would reduce the memory footprint even more, but I found that Unity doesn't support that. Has anyone heard of a way to use that?


EliteAsian123:
Hi, thanks very much for this PR!

Something I noticed is that the fonts on the main menu look slightly more bubbly? 
Before:
image

After:
image

I'm not sure exactly what is causing this but there's definitely something a little different looking. Thanks!

rk-exxec:
Thats odd, Ill have a look

rk-exxec:
This font has been dynamic before and I forgot to increase the atlas size to 1024x1024 when converting to static. also re-checked all the others.
I also enabled the option to clear the dynamic font cache on build for the noto fonts.

PR #766, opened 2024-05-07
Fix More Options menu not opening on click

poliwhirlge:
After #756 since the menu opening was moved to on button release it doesn't trigger the opening the menu, adding in the onPointerUp on the HelperButton in the menu footer to trigger it.

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -514,7 +514,7 @@ namespace YARG.Menu.MusicLibrary
         {
             var holdContext = _heldInputs.FirstOrDefault(i => i.Context.IsSameAs(ctx));
 
-            if (ctx.Action == MenuAction.Orange && holdContext?.Timer > 0)
+            if (ctx.Action == MenuAction.Orange && (holdContext?.Timer > 0 || ctx.Player is null))
                 _popupMenu.gameObject.SetActive(true);
 
             _heldInputs.RemoveAll(i => i.Context.IsSameAs(ctx));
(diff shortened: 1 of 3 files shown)


EliteAsian123:
I'm not really too sure how good this would be UX-wise, but it may be better to make it so that holding the helpbar button has the same effect as holding the binding button (effectively just removing the `|| ctx.Player is null`).

In this particular case, it doesn't matter that much, however, in the future if we add more holding binds, it may make sense to keep the behaviour consistent throughout the help bar and the binds.

poliwhirlge:
Yeah UX-wise I don't see people clicking down the orange button then doing up+down on the keyboard (although maybe some will?) but I agree with making it consistent for the other buttons.

Right now though the help bar buttons fire onClick, which is basically a pointerDown + pointerUp event, whereas the binds (except orange) fire the moment we have a buttonDown, so I think we need to make this consistent and move the Invoke on the help bar to OnPointerDown instead. Once we do this I think if we add holds to the other binds the same way we did orange they should just work with the mouse.

Another thing I was wondering is the mouse doesn't have a profile assigned to it where as the keyboard does, so to get the mouse clicks to work I needed to identify the mouse and let that event go through (thus the `ctx.Player is null` check). If you'd like me to handle this another way I'm all ears.

If this sounds ok I'll go ahead and make the changes. Thanks!

EliteAsian123:
> Right now though the help bar buttons fire onClick, which is basically a pointerDown + pointerUp event, whereas the binds (except orange) fire the moment we have a buttonDown, so I think we need to make this consistent and move the Invoke on the help bar to OnPointerDown instead. Once we do this I think if we add holds to the other binds the same way we did orange they should just work with the mouse.

Yup, this is perfect.

> Another thing I was wondering is the mouse doesn't have a profile assigned to it where as the keyboard does, so to get the mouse clicks to work I needed to identify the mouse and let that event go through (thus the ctx.Player is null check). If you'd like me to handle this another way I'm all ears.

This is something I did not consider at all, but yeah it's not ideal to have to deal with this. Since the orange button is per player, it doesn't really make sense to have the help bar clickable in this case. My bad.

Making the change above though I think is still a good idea incase we do add more hold binds in the future.

Thanks!

poliwhirlge:
Alright, change made, thanks again!

PR #771, opened 2024-05-14
Wait countdown

Purplo-cf:
Adds support for new WaitCountdown prefab in 5-fret, drums, and vocals engines during long stretches with no notes


Assets/WaitCountdown.cs
```diff
@@ -0,0 +1,78 @@
...
+                    if (!gameObject.activeSelf)
+                    {
+                        gameObject.SetActive(true);
+                    }
+                }
+                else
+                {
+                    //hide countdown when there is one measure before the next set of notes
+                    gameObject.SetActive(false);
+                }
```
sonicfind, line 68: Early return here as well (we'll figure out how to make it fade out in the future)

Assets/WaitCountdown.cs
```diff
@@ -0,0 +1,78 @@
...
+                        _countdownText.text = measuresLeft.ToString();
+
+                        childToHide = _getReady.gameObject;
+                        childToShow = _backgroundCircle.gameObject;
+                    }
+                    else if (measuresLeft == 1)
+                    {
+                        childToHide = _backgroundCircle.gameObject;
+                        childToShow = _getReady.gameObject;
+                    }
```
sonicfind, line 45: Firstly, much rather have the active state directly set appropriately within these scopes over holding the objects in temporary variables.

Secondly, the `measuresLeft == 1` conditional is redundant because we already checked for `measuresLeft > 0` before this point.

Assets/WaitCountdown.cs
```diff
@@ -0,0 +1,78 @@
...
+                    if (childToHide?.activeSelf == true)
+                    {
+                        childToHide.SetActive(false);
+                    }
+
+                    //update visibility of parent object if hidden
+                    if (!gameObject.activeSelf)
+                    {
+                        gameObject.SetActive(true);
+                    }
```
sonicfind, line 62: I believe these checks of the `activeSelf` property are redundant. Unity should already run said test before making changes to the gameObject's state. Just running `SetActive` on it's own should suffice.

sonicfind (changes requested):
Just some minor syntactical changes. The functionality looks to be there (though I haven't actually tested it myself).

Assets/WaitCountdown.cs
EliteAsian123: Just to add on, using `?` on Unity objects is generally not advised since they treat null differently. It's best to check if it's null explicitly using `==` (which is slow by the way, because Unity).

Purplo-cf: Ready for another look

Assets/WaitCountdown.cs
Purplo-cf: Ready for another look

Assets/WaitCountdown.cs
Purplo-cf: Nevermind, it was simple enough to copy it over. I kept the added fade animation in its own commit so it could be reverted if you'd rather not have it

Assets/WaitCountdown.cs
```diff
@@ -0,0 +1,78 @@
...
+        private Image _getReady;
+
+        private uint _measuresLeft;
+
+        public void UpdateCountdown(uint measuresLeft)
+        {
+            if (measuresLeft != _measuresLeft)
+            {
+                _measuresLeft = measuresLeft;
+
```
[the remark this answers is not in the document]
Purplo-cf: Ready for another look

sonicfind:
image
I would make this smaller. Having it span the entire width of the highway is a bit much.
The height also doesn't match the height of the circle, so there's a bit of a disconnect.

sonicfind:
Also, I'll mention this here too: Don't *merge*, **rebase**. Because of all you're merging, your PR is mostly now just a bunch of merge commits. It makes it nearly impossible now to see what changes to your own code were made in the sea of the merged commits.

Purplo-cf:
I have adjusted the size of the "Get Ready!" box and added a hook for it into the localization table. Pinging @kaduwaengertner since he mentioned that he already created assets for this. Are those ready to implement?

Also, I just realized that Clone Hero only shows the the countdown and doesn't do the "Get Ready!" that RB4 does. Do you have a preference between the two styles?

If it helps, this was the original RB4 reference that I used for for sizing and positioning
RB4_Countdown_ref

kaduwaengertner:
Just added a reminder for myself to send you the assets tomorrow when Im at my pc.

TheNathannator:
> Also, I just realized that Clone Hero only shows the the countdown and doesn't do the "Get Ready!" that RB4 does. Do you have a preference between the two styles?

I personally prefer what Clone Hero does, as it feels more polished and cohesive *because* it doesn't use the second textbox style. In YARG, the "Get Ready!" textbox would definitely feel a little out of place, as there's nothing else that uses that style of box except for the solo box.

CH also has the progress ring around the countdown circle, which is a nice little indicator and compliments the circle design.

Issue #442, opened 2023-06-08
Logitech USB Microphone not working in 0.10.5 and 0.10.6 on Steam Deck (Linux)


thumtac:
I'm playing on steam deck and my logitech usb microphone is not working. The device is detected (I think) but when I go to play a song singing into the mic does not register.

The mic works fine on 0.10.4 and earlier. Patch notes say there is a new mic backend in 0.10.5 so it could be related?

TheNathannator:
Yeah there are issues on Linux with mics currently, we're looking into it! So far it seems to be something about not using the right recording format for the mic, bit of a strange issue.

EliteAsian123:
This *should* be fixed now. Please reopen if not!

thumtac:
I'm still getting this issue on 0.10.7. No microphones show up in the device list. No problems observed on 0.10.4 so I think my setup is still working

thumtac:
@EliteAsian123 I don't have privileges to re-open this issue though so would appreciate if you rep-opened. Thanks!!

coverslide:
I just tried this too on 0.11 using Ubuntu, I can confirm the USB microphone receives input, but the option does not show up when I add a new player.

Brahelli-Fails-At-Development:
I have the same issue here, microphone is picked up by my system, but the game does not pick it up as a possible player

RileyTheFox:
@coverslide @Brahelli-Fails-At-Development @thumtac 

Been a while since this issue was looked at, can you try out the nightly builds and confirm whether or not the problem has fixed since? Lots of changes since then and now and we made some changes to microphone detection.

Brahelli-Fails-At-Development:
The issue was fixed before the new engine moved the nightly, sorry for not updating. The issue now is on my end, haha. 
Thanks for checking up

thumtac:
@RileyTheFox microphone not working on the 11/26 nightly.

When adding a new profile, usb microphone appears as a device. When I added it to the connected profile it didn't register at all during song play. Other weird thing was that you couldn't remove the device,  the USB microphone didn't appear in the devices to remove list. I even tried adding a controller as an additional device on top of the microphone but that didn't fix the issue.

Crono141:
[...]

PR #776, opened 2024-05-18
Linux Microphone Fixes

RileyTheFox:
Related to https://github.com/YARC-Official/YARG/issues/442

Implements a number of fixes for Linux by slightly refactoring how microphone recording is handled:
- Only 1 mic recording is now used (Simultaneous recordings are sometimes not available on Linux)
  - The recording period of this stream's callback is the monitoring period (5ms) as the pitch processing is now batched.
- The channel count for the streams are now hard-coded to 1, as the vast majority of microphones are mono input. This gets around the fact that on Linux it is not possible to retrieve a recording device's available formats.
- A second push stream is used to buffer up microphone input ready for pitch processing. This data is buffered until the pitch processing period (40ms) has passed and is then processed. This is the alternative to using a 2nd recording stream.


Assets/Script/Audio/Bass/BassMicDevice.cs
```diff
@@ -220,33 +216,58 @@ public sealed class BassMicDevice : MicDevice
...
+            // And you can't just request as much data as possible into a NULL buffer because that only works for recording streams.
+            // So we have to allocate a buffer and keep requesting data until there's none left.
+            unsafe
             {
-                return (int) Bass.LastError;
+                const int bufferLength = 1024;
+
+                byte* buffer = stackalloc byte[bufferLength];
+                int bytesRead;
+                do
```
RileyTheFox, line 257: I'm not sure if this is the best way to go about clearing the pitch processing buffer. Due to it being a decoding channel, you cannot request the available amount of data in the channel as it has no playback buffer. I've read through the BASS documentation and can't find anything else that lets you clear a channel's data. `BASS_ChannelPlay(restart = true)` would clear it, but you cannot play decoding channels either

Assets/Script/Audio/Bass/BassMicDevice.cs
```diff
@@ -285,40 +307,65 @@ public override SerializedMic Serialize()
...
             if (!IsRecordingOutput)
             {
                 return true;
             }
 
-            CalculatePitchAndAmplitude(buffer, length);
+            // Copy the data to the batch handle to apply FX
+            Bass.StreamPutData(_recordHandle.ProcessedHandle, buffer, length);
+
+            _timeAccumulated += 10;//RecordingHandle.CLEAN_RECORD_PERIOD_MS;
```
RileyTheFox, line 334: I noticed when testing these changes that the callback period is not 5ms like specified, but was actually measured at around ~10ms. With the new batching and accumulating time until it reaches 40ms, this caused the pitch processing rate to be half.

I hard coded the value to 10 to fix this but there should probably be a better way of knowing how frequent the callback is ran because BASS seems to be overriding the 5ms that was passed

Assets/Script/Audio/Bass/BassMicDevice.cs
```diff
@@ -285,40 +307,65 @@ public override SerializedMic Serialize()
...
+            _processedBufferLength += length;
+
+            // Enough time has passed for pitch detection
+            if(_timeAccumulated >= RECORD_PERIOD_MS)
+            {
+                unsafe
+                {
+                    byte* procBuff = stackalloc byte[_processedBufferLength];
+
+                    var readOnlySpan = new ReadOnlySpan<byte>(procBuff, _processedBufferLength);
```
sonicfind, line 345: Move this line below the `GetData` call and instead create the Span of shorts directly from here with a manual cast of the pointer. There's no point in passing a span of bytes if casting is all that happens to it.

Assets/Script/Audio/Bass/BassMicDevice.cs
sonicfind: It's most likely bound by device period instead of the update period.

Assets/Script/Audio/Bass/BassMicDevice.cs
sonicfind: Beats me honestly. Maybe SetPosition to 0? idk how the recording works. I just copy and pasted that from the old stuff. lol

sonicfind (commented):
I didn't really mess with this much, so you can do what you need to.

PR #777, opened 2024-05-20
Implement Instrument Sorting

sonicfind:
+ Add options to sort by the instruments present to play in the user's library.
+ Add Playable sort option that takes the instrument modes of the currently active player profiles to only display songs that can be played with *everybody* included.
+ Change resulting behavior of clicking an instrument icon form the sidebar. Will now show *only* songs that contain the instrument at that intensity, but without forced sorting.


Assets/Art/Fonts/Barlow/Barlow-Italic.asset
EliteAsian123, line 1: It appears some random font values have been reset

Assets/Art/Fonts/NotoSans/NotoSansJP-Regular SDF.asset
EliteAsian123, line 1: Same here

EliteAsian123 (changes requested):

PR #779, opened 2024-05-21
Option to prioritize displaying higher % scores in MusicLibrary

poliwhirlge:
Discussion threads:
https://discord.com/channels/1086048856678084609/1211121737358512238
https://discord.com/channels/1086048856678084609/1086048857714069519/1236879158663970889

Tries to display the best FC/percentage result, ties are broken by difficulty.

If the PlayerScores table is from the older version without the Percentage column / is not populated with data, it will try to fill it with the calculated value.

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/ViewTypes/SongViewType.cs
+++ b/Assets/Script/Menu/MusicLibrary/ViewTypes/SongViewType.cs
@@ -19,6 +19,12 @@ namespace YARG.Menu.MusicLibrary
         Off
     }
 
+    public enum HighScorePriorityMode
+    {
+        Percentage,
+        Score
+    }
+
     public class SongViewType : ViewType
     {
         public override BackgroundType Background => BackgroundType.Normal;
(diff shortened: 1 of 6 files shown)


EliteAsian123 (changes requested):
Hi, thanks for the PR!

A few things:
* Can these changes be moved to the `engine-refactor` branch? This does mean that this will not be addressed ASAP, however, `engine-refactor` has a change regarding how percents are stored within the scores (which is what the `ScoresTableHasPercentColumn` is about I assume).
* Regarding *how* this should be done, I think the consensus and the overall precedent is that it should show the highest score *and* the highest percent *at the same time*. The problem with this now is that the difficulty and instrument don't make sense if we are querying for the highest percent and the highest score. To fix this, I think the best road would be to:
  1. Search the database for the highest score and use that to display the stars/score, instrument, and difficulty.
  2. Find the highest percent that was achieved with the same instrument and difficulty, and display that.
 
Now this is just my opinion, so I've messaged this on the "% bug" thread just to get the opinion of other folks, however, I'm pretty sure this is the best way of doing it.

Thanks!

poliwhirlge:
Re-submitted as #785

PR #785, opened 2024-05-25
Show best % score for instrument and difficulty of the highest score

poliwhirlge:
Re-submission of pull #779
Now finds and displays the highest % score achieved for the same instrument+difficulty and displays that instead of the % score achieved on the highest scoring run.

Assumes the `Percent` column exists in the `PlayerScores` table exists (which it does on this `engine-refactor` branch).

The additional query should be performant enough, runs in 14ms on my db of 1021 rows whereas the original query runs in 4ms.

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/ViewTypes/SongViewType.cs
+++ b/Assets/Script/Menu/MusicLibrary/ViewTypes/SongViewType.cs
@@ -52,6 +52,7 @@ namespace YARG.Menu.MusicLibrary
         public override string GetSideText(bool selected)
         {
             var score = ScoreContainer.GetHighScore(SongEntry.Hash);
+            var bestPctScore = ScoreContainer.GetBestPercentageScore(SongEntry.Hash);
 
             // Never played!
             if (score is null)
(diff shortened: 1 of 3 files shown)


EliteAsian123 (changes requested):
Hi, thanks so much for the PR!

A problem that I see with this is that the query is absolutely massive, which makes it quite hard to maintain in the event of changes.

One possible fix is to use the already-fetched instrument and difficulty from the high scores and use those to find the highest percent (much like in the `GetHighScoreByInstrument` method). This is much simpler, however, it may have some performance implications so this would have to be tested. I'm not sure which is faster as I don't have too much experience with SQL.

Another option is to return only the highest percent in the query, as that is all that really matters in this scenario (since all of the other information is attached to the normal high score).

Thanks again!

poliwhirlge:
Alright, now on start up if the Percent field isn't populated it'll fill it in so we won't have to worry about nulls, and past the first run it should take no time.

Query should be easier to parse through now, and I've added comments describing what it's doing.

EliteAsian123 (approved):
Hi, thanks for all of the changes!

~~This is pretty much good, except the `GetPercent` method in `PlayerScoreRecord` shouldn't be needed now since the database is updated with the nulls being removed.~~

Edit: nvm, I see now that the percent is stored as a nullable so it's still needed to an extent

poliwhirlge:
Thanks!

It's probably fine to remove the nullable tbh. The only time the Percent column will be NULL now is in the broken case of division by 0, and in that case it'll fill in the Percent field with 0 anyway, which seems acceptable to me. Up to you.

PR #794, opened 2024-06-07
Drum velocity

Purplo-cf:
Unity prefab setup for ghost/accent notes for both pads and cymbals in the Rectangular theme - Four lane and Five lane compatible. These are purely placeholders and need a LOT of design love. I got the models from Kadu but I'm pretty sure I assigned the materials to the wrong meshes.

Changed all drum IndividualButtonBindings to a new velocity-sensitive DrumPadButtonBinding subclass


Assets/Script/Input/Bindings/ControlBinding.cs
Purplo-cf, line 1: This was the easiest way that I could find to extend FireInputEvent on DrumPadButtonInput to accept a float value. I made this a virtual method in ControlInput and I'm overriding it in DrumPadButtonInput. I saw Nathan mention something recently about virtual methods aren't great for performance, so let me know if there's a better way to handle this.

Assets/Script/Input/Bindings/DrumPadButtonBinding.cs
Purplo-cf, line 1: IndividualButtonBinding sends out IsPressed as a bool value, so I'm trying to capture that FireInputEvent call here and send the highest axis value of all mapped SingleBindings. This should adjust automatically if a player adjusts their press threshold in their profile right?

Assets/Art/Fonts/Barlow/Barlow-Italic.asset
EliteAsian123, line 1: Was this file meant to be changed? If not, it should be reverted. Sometimes Unity randomly modifies files.

Assets/Art/Meshes/Gameplay/Notes/Rectangular/Cymbal_Accent.fbx
EliteAsian123, line 1: The model file names should match those of the already existing ones. For example, `AccentCymbalNote` or `CymbalAccentNote` may be a better name for this. Same with the other models added!

Assets/Scenes/PersistentScene.unity
EliteAsian123, line 1: I'm also assuming this was an unintentional change

Assets/Script/Gameplay/Visuals/TrackElements/Drums/FiveLaneDrumsNoteElement.cs
```diff
@@ -24,12 +24,14 @@ protected override void InitializeElement()
...
                 {
-                    NoteGroup = (FiveLaneDrumPad) NoteRef.Pad switch
+                    bool isCymbal = (FiveLaneDrumPad) NoteRef.Pad switch
                     {
-                        FiveLaneDrumPad.Yellow => noteGroups[(int) NoteType.Cymbal],
-                        FiveLaneDrumPad.Orange => noteGroups[(int) NoteType.Cymbal],
-                        _                      => noteGroups[(int) NoteType.Normal]
+                        FiveLaneDrumPad.Yellow => true,
+                        FiveLaneDrumPad.Orange => true,
+                        _                      => false
```
EliteAsian123, line 31: This can just be converted into a simpler expression of
```
bool isCymbal = (FiveLaneDrumPad) NoteRef.Pad is 
    FiveLaneDrumPad.Yellow or FiveLandDrumPad.Orange;
```

Assets/Script/Input/Bindings/BindingCollection.Templates.cs
```diff
@@ -59,28 +59,28 @@ public static BindingCollection CreateSixFretGuitarBindings() => new(GameMode.Si
         public static BindingCollection CreateFourLaneDrumsBindings() => new(GameMode.FourLaneDrums)
         {
             // TODO: Velocity support
```
EliteAsian123, line 61: This TODO can be removed

EliteAsian123 (changes requested):
Hi, thanks for the PR!

It appears the YARG.Core submodule pointer is broken, so I haven't tested this in-game, however, I've still pointed out some things I noticed below.

I'll also request a review from Nathan for the input stuff as he knows more than I do when it comes to that.

Thanks again!

Purplo-cf:
> It appears the YARG.Core submodule pointer is broken, so I haven't tested this in-game, however, I've still pointed out some things I noticed below.

Sorry about that! I'm not totally sure what happened the first time, but git let me update the pointer for today's changes so I'm hoping that fixed it 🤞

Thanks for the quick replies on all of this!

TheNathannator (approved):
Seems good from an initial look-over! I'll have to do some playing around with things when I get the chance, but this should work.

Purplo-cf:
Rad - Thank you 🤘!

I think I updated the pointer correctly - let me know if I need to make any changes!

PR #796, opened 2024-06-10
Generate drum od phrases

Purplo-cf:
Implementation of new YARG.Core method once an InstrumentDifficulty<DrumNote> chart is loaded.


Purplo-cf:
The most recent changes are now all contained within YARG.Core. Closing this request

PR #802, opened 2024-06-15
Drum Hit SFX

raphaelgoulart:
Should be good to be reviewed, but it's not final because I'm waiting on real drum samples by Jarod Fedele.

TODO:
- ~~Add localization string to the DrumSfxVolume setting because I literally forgot;~~
- ~~Replace placeholder samples with real ones;~~

How sampling works:

There are 3 velocity levels (weak/regular/strong), for 8 pads (kick, snare, 3 toms, 3 cymbals), with 3 samples each. When the player plays the same pad it rotates between those 3 samples, so even during a fast roll it'll sound natural.


Assets/Art/Fonts/Barlow/Barlow-Italic.asset
EliteAsian123, line 1: This probably was an unintentional change

EliteAsian123 (changes requested):
Overall pretty good; the sample stuff will likely need improvement but that's not really my expertise so I won't comment on that.

Also, when you get the real samples, if you could provide them as `.ogg` (since they're smaller, probably) that'd probably slightly speed the loading up along with decrease the game's file size.

Assets/Art/Fonts/Barlow/Barlow-Italic.asset
raphaelgoulart: Whoops, didn't notice as it was between other actual changes (Unity L) - will revert now

raphaelgoulart:
Also yeah I can provide final samples as `.ogg`, but does the game support `.opus` for SFX? They're slightly more efficient

If not, 112kbps oggs are probably more than good enough!

EliteAsian123:
Not sure honestly, but if they do that'd probably be even better. Thanks!

raphaelgoulart:
Can confirm .opus works, sample sizes went from 13mb to about 500kb! Yeah I'll def do that once the real samples arrive

Assets/Script/Audio/Bass/BassDrumSampleChannel.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            if (!Bass.ChannelPlay(_channel, true))
+            {
+                YargLogger.LogFormatError("Failed to play {Sample} channel: {0}!", Bass.LastError);
+            }
+        }
+
+        protected override void SetVolume_Internal(double volume)
+        {
+            volume *= GlobalAudioHandler.GetVolumeSetting(SongStem.DrumSfx);
```
sonicfind, line 53: To go with the Core PR review, the volume will already be pre-multiplied, so you won't need to do that here.

Assets/Script/Audio/Bass/BassAudioManager.cs
```diff
@@ -291,6 +292,35 @@ private void LoadSfx()
             YargLogger.LogInfo("Finished loading SFX");
         }
 
+        public void LoadDrumSfx()
```
sonicfind, line 295: `private`

sonicfind (changes requested):
On the audio side: pretty simple yea, just a few minor adjustments,

Assets/Script/Audio/Bass/BassAudioManager.cs
```diff
@@ -288,6 +289,35 @@ private void LoadSfx()
...
+                {
+                    string sfxPath = sfxBase + format;
+                    if (File.Exists(sfxPath))
+                    {
+                        var sfxSample = AudioHelpers.GetDrumSfxFromName(sfxFile);
+                        var sfx = BassDrumSampleChannel.Create(sfxSample, sfxPath, 8);
+                        if (sfx != null)
+                        {
+                            DrumSfxSamples[(int) sfxSample] = sfx;
+                            YargLogger.LogFormatInfo("Loaded {0}", sfxFile);
```
RileyTheFox, line 311: Given how many samples there are to load, I think it may be better to only log if a sample failed to load rather than every time one is loaded. Just to prevent useless log spam

raphaelgoulart: Removed log, since an error is printed if it fails to load in the bass sample part of the code

Assets/Script/Audio/Bass/BassDrumSampleChannel.cs
```diff
@@ -0,0 +1,64 @@
...
+    public sealed class BassDrumSampleChannel : DrumSampleChannel
+    {
+#nullable enable
+        public static BassDrumSampleChannel? Create(DrumSfxSample sample, string path, int playbackCount)
+#nullable disable
+        {
+            int handle = Bass.SampleLoad(path, 0, 0, playbackCount, BassFlags.Decode);
+            if (handle == 0)
+            {
+                YargLogger.LogFormatError("Failed to load {sample} {path}: {0}!", Bass.LastError);
```
EliteAsian123, line 18: Wait does `{sample} {path}` actually work? If so I did not know that lol

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -35,6 +35,9 @@ public class DrumsPlayer : TrackPlayer<DrumsEngine, DrumNote>
 
         public override int[] StarScoreThresholds { get; protected set; }
 
+        private int[] DrumSoundEffectRoundRobin = new int[8];
+        private float DrumSoundEffectAccentThreshold;
```
EliteAsian123, line 39: Private variables should be named `_likeThis` (camel case with underscore)

EliteAsian123 (changes requested):
Overall awesome, just a few style changes.

Are these the final samples now? I've seen the new ones in the Discord but the PR still says you're waiting for them. Just to make sure, we do have full legal permission to use the new samples, correct?

raphaelgoulart:
> Overall awesome, just a few style changes.
> 
> Are these the final samples now? I've seen the new ones in the Discord but the PR still says you're waiting for them. Just to make sure, we do have full legal permission to use the new samples, correct?

Yes, I just forgot to update the text
And also yes, Jarod Fedele himself recorded those specifically **for** the game. Contact him if you need any confirmation

Assets/Script/Audio/Bass/BassDrumSampleChannel.cs
raphaelgoulart: I haven't had any situations where samples failed to load, but this is borrowed from BassSampleChannel.cs (presumably sonicfind's code) so it should work

Assets/Script/Gameplay/Player/DrumsPlayer.cs
raphaelgoulart: Fixed

EliteAsian123 (approved):

Assets/Script/Audio/Bass/BassDrumSampleChannel.cs
sonicfind: pfft it doesn't there either lol, it just *never happened* so it never got noticed

raphaelgoulart: fixed then lol

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -265,5 +287,46 @@ protected override bool InterceptInput(ref GameInput input)
         {
             return false;
         }
+
+        private void PlayDrumSoundEffect(DrumsAction action, float velocity)
+        {   
+            if (IsDrumFreestyle())
```
sonicfind, line 293: This check isn't needed as that's already accounted for before calling the function on line 132

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -265,5 +287,46 @@ protected override bool InterceptInput(ref GameInput input)
...
+                int actionIndex = (int) action;
+                double sampleVolume = velocity;
+
+                // Define sample
+                int sampleIndex = (int) DrumSfxSample.Vel0Pad0Smp0;
+                if (velocity > _drumSoundEffectAccentThreshold)
+                {
+                    sampleIndex = (int) DrumSfxSample.Vel2Pad0Smp0;
+                }
+                else if (velocity > EngineParams.VelocityThreshold)
```
sonicfind, line 304: Probably wanna add a comment here stating `VelocityThreshold` maps to the "ghosting" line.

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -265,5 +287,46 @@ protected override bool InterceptInput(ref GameInput input)
...
+                {
+                    sampleIndex = (int) DrumSfxSample.Vel1Pad0Smp0;
+                    sampleVolume = velocity / _drumSoundEffectAccentThreshold;
```
sonicfind, line 307: Plus one here (and in the next branch) stating the division is normalizing the volume using `_drumSoundEffectAccentThreshold` as a pseudo `"1"`

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -265,5 +287,46 @@ protected override bool InterceptInput(ref GameInput input)
...
+                if (_drumSoundEffectRoundRobin[actionIndex] == DrumSampleChannel.ROUND_ROBIN_MAX_INDEX)
+                {
+                    _drumSoundEffectRoundRobin[actionIndex] = 0;
+                }
+            }
+        }
+
+        private bool IsDrumFreestyle()
+        {
+            return Engine.State.NoteIndex == 0 || Engine.State.NoteIndex >= Notes.Count; // TODO: add drum fill / BRE conditions
```
sonicfind, line 329: Assuming `wasNoteHit` being `true` in the `engine.OnPadHit` lambda also means that it already incremented `Engine.State.NoteIndex`, this may cause a drum sfx to play every time the last note of a chart is played.

You can only truly solve this by making drum sfx activate/deactivate based purely on the presence of notes in a certain timeframe instead of by note index. Something along the lines of: "If the distance between these notes is greater than [value] & current time is outside of either note's hit window, allow sfx" + plus the BRE fill and start/end cases.

sonicfind (changes requested):

Assets/Script/Gameplay/Player/DrumsPlayer.cs
raphaelgoulart: This isn't needed because `if (wasNoteHit) return;` (line 100) ensures the remaning of the `engine.OnPadHit` code won't be ran (and as such, `PlayDrumSoundEffect(action, velocity);` won't be triggered)

And in my tests this is exactly what happens - not once I had the drum sfx being triggered on the last note of a song, only afterwards

sonicfind (approved):

PR #808, opened 2024-06-22
Wrap-around navigation for menus

Purplo-cf:
Replaced clamp functions on nav selection index. More lines, but arguably better for UX. Addresses issue [YARG-178](https://yarg.youtrack.cloud/issue/YARG-178/Allow-wrap-around-when-navigating-lists-with-a-controller)

### DIFF
--- a/Assets/Script/Menu/Common/ListMenu/ListMenu.cs
+++ b/Assets/Script/Menu/Common/ListMenu/ListMenu.cs
@@ -40,9 +40,17 @@ namespace YARG.Menu.ListMenu
                 {
                     _selectedIndex = 0;
                 }
+                else if (value > _viewList.Count - 1)
+                {
+                    _selectedIndex = 0;
+                }
+                else if (value < 0)
+                {
+                    _selectedIndex = _viewList.Count - 1;
+                }
                 else
                 {
-                    _selectedIndex = Mathf.Clamp(value, 0, _viewList.Count - 1);
+                    _selectedIndex = value;
                 }
 
                 OnSelectedIndexChanged();
--- a/Assets/Script/Menu/Common/ListMenu/ListMenu.cs
+++ b/Assets/Script/Menu/Common/ListMenu/ListMenu.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.InputSystem;
 using UnityEngine.UI;
+using YARG.Settings;
 
 namespace YARG.Menu.ListMenu
 {
@@ -40,17 +41,26 @@ namespace YARG.Menu.ListMenu
                 {
                     _selectedIndex = 0;
                 }
-                else if (value > _viewList.Count - 1)
+                else if (SettingsManager.Settings.WrapAroundNavigation.Value)
                 {
-                    _selectedIndex = 0;
-                }
-                else if (value < 0)
-                {
-                    _selectedIndex = _viewList.Count - 1;
+                    // Wrap to bottom/top of list when moving past the start/end range
+                    if (value > _viewList.Count - 1)
+                    {
+                        _selectedIndex = 0;
+                    }
+                    else if (value < 0)
+                    {
+                        _selectedIndex = _viewList.Count - 1;
+                    }
+                    else
+                    {
+                        _selectedIndex = value;
+                    }
                 }
                 else
                 {
-                    _selectedIndex = value;
+                    // Do not allow selection to move past the start or end range
+                    _selectedIndex = Mathf.Clamp(value, 0, _viewList.Count - 1);
                 }
 
                 OnSelectedIndexChanged();
--- a/Assets/Script/Menu/Common/ListMenu/ListMenu.cs
+++ b/Assets/Script/Menu/Common/ListMenu/ListMenu.cs
@@ -136,7 +136,7 @@ namespace YARG.Menu.ListMenu
             SelectedIndex = Mathf.FloorToInt(_scrollbar.value * (_viewList.Count - 1));
         }
 
-        public void SetWrapAroundState (bool newState)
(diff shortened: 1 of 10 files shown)


EliteAsian123:
This was actually intentionally removed as we've received feedback from multiple people who have poor eyesight that it makes the UX a *lot* more confusing and hard to follow. If we were to add this back, it should be added as a setting that is disabled by default.

Purplo-cf:
Oh gotcha, thanks for the context. I turned this into a setting that is off by default.

EliteAsian123:
Looks fine to me now, however, the YARG.Core pointer gets completely messed up in this PR. Once that is reverted, this can be merged. Thanks!

RileyTheFox:
> This was actually intentionally removed as we've received feedback from multiple people who have poor eyesight that it makes the UX a _lot_ more confusing and hard to follow. If we were to add this back, it should be added as a setting that is disabled by default.

I really disagree, so many games have wrap around in their UIs and especially in a game like ours where menus can get *really* long (music library, practice section) it is super useful to have UI wrap around. If people struggle to notice it, that's a major drawback on how the UI is displayed, not on a wrap around feature.

EliteAsian123:
> I really disagree, so many games have wrap around in their UIs and especially in a game like ours where menus can get *really* long (music library, practice section) it is super useful to have UI wrap around. If people struggle to notice it, that's a major drawback on how the UI is displayed, not on a wrap around feature.

Obviously, I cannot speak on behalf of visually impaired individuals as I am not visually impaired. However, the people I have talked to about this told me that the problem is major jumps/changes within scrolling lists and highlights, which causes confusion (i.e., "getting lost"). This is generally unavoidable regardless of UI design.

One of the people also mentioned to me that a lot of games do *do* this, which is the problem. The were very appreciative when I actually went about the changes.

RileyTheFox:
What about a meet in the middle approach? When you scroll to the end, it stops. To wrap back around, you have to re-press the input. I'm pretty sure this is how most games implement it (rather than it just immediately wraps as soon as the end is reached). This is for auto scrolling like holding down the strumbar or down arrow for exampke (doesn't really work with scroll wheels) 

It would tell the user that the end of the menu has been reached, and that if they scroll down again it will go back to the top.

I think this sort of feature benefits more people than it would negatively impact, which is why I believe it should be on by default. Think of it this way: If somebody has problems with menu wrap-around, they are the type of person that is more likely to search for something to disable it (because it has an impact on them)

I think the game's default settings should cater to the majority. Before this problem was even brought up I never even knew that wrap-around could be a problem for some people and I'd wager many others were unaware of it also.

Purplo-cf:
I think both points are valid. In my experience, whenever I've started a game with text-to-speech support for the visually impaired, it's on by default with a first-time startup option to turn it off. Colorblind accessibility options on the other hand are usually off by default, and the user would have to know to check the game settings before turning it on.

I can't think of any precedent for a game to have something like this as an accessibility option, so even if a visually impaired person could benefit from turning off wrap-around, they might not think to check for it. Meanwhile, a lot of YARG's player base is already digging through settings for track speed, track length, etc.

If we're talking about somebody who already has a gigantic song library, the odds are pretty good that they would stumble on this setting

Purplo-cf:
> When you scroll to the end, it stops. To wrap back around, you have to re-press the input. I'm pretty sure this is how most games implement it (rather than it just immediately wraps as soon as the end is reached). This is for auto scrolling like holding down the strumbar or down arrow for exampke (doesn't really work with scroll wheels)

Oh I like that better. I'll start working on that today

EliteAsian123:
> I think the game's default settings should cater to the majority. Before this problem was even brought up I never even knew that wrap-around could be a problem for some people and I'd wager many others were unaware of it also.

For some reason the edit on your original message didn't show up for me, but yeah I agree. As long as the option is there it's fine.

EliteAsian123:
> When you scroll to the end, it stops. To wrap back around, you have to re-press the input. 

I'm having intense déjà vu right now because now that I think about it I feel like we already did this.

Purplo-cf:
Wrap-around is now being prevented during a hold while the setting is turned on. It is still set to Off by default

Assets/Script/Menu/Common/ListMenu/ListMenu.cs
```diff
@@ -116,6 +136,18 @@ public void OnScrollBarChange()
             SelectedIndex = Mathf.FloorToInt(_scrollbar.value * (_viewList.Count - 1));
         }
 
+        public void SetWrapAroundState (bool newState)
```
EliteAsian123, line 139: Remove space

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -100,6 +100,7 @@ private void OnEnable()
                         if (IsButtonHeldByPlayer(ctx.Player, MenuAction.Orange))
                             GoToPreviousSection();
                         else
+                            SetWrapAroundState(!ctx.IsRepeat);
```
EliteAsian123, line 103: Since there are no braces on these ifs, only the first statement is technically in the if (and not the `SelectIndex--;` below). Please add braces!

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -108,6 +109,7 @@ private void OnEnable()
                         if (IsButtonHeldByPlayer(ctx.Player, MenuAction.Orange))
                             GoToNextSection();
                         else
+                            SetWrapAroundState(!ctx.IsRepeat);
```
EliteAsian123, line 112: Same here

EliteAsian123 (changes requested):

Issue #810, opened 2024-06-29
Arrow Keys for Track Navigation


jbienz:
Sorry if I'm totally missing something, but I'd like to be able to navigate up and down the track list with the arrow keys on the keyboard (regular keyboard, not MIDI). I know the current way of navigating up and down the track list is with the scroll wheel, but would it be possible to add the arrow keys too?

Thanks!

TheNathannator:
You can add your keyboard to your profile as a second device and that'll allow you to navigate with it. The game currently doesn't do keyboard navigation without it being assigned to a profile, but we do plan on implementing that at some point.

jbienz:
Thanks @TheNathannator. While I was doing all my initial hardware setup, I just had some bots playing. And it doesn't appear you can add secondary devices to the bots. But I'm glad to see you're planning it, and thank you for this amazing software!

Dansla116:
Seems like this was fixed by #816 and just never closed.  If not, it's fixed by something else now, in Nightly.

PR #816, opened 2024-07-03
Global Keyboard Navigation

jbienz:
I've implemented basic global keyboard navigation for the following keys:

```
Key.UpArrow = MenuAction.Up
Key.DownArrow = MenuAction.Down
Key.LeftArrow = MenuAction.Left
Key.RightArrow = MenuAction.Right
Key.Enter = MenuAction.Green
Key.Escape = MenuAction.Red
```

When pressed, these keys simulate the specified input action for the first player. It works on all screens and does not require binding.

I'm happy with this and would like to see it merged, but I'd also like to start a conversation about more enhanced input.

For example, I'd love to see Page Down go to Next Section and Page Up go to Previous Section in the Music list. Unfortunately, there is no MenuAction for Next and Previous Section. This is currently implemented as a combo of holding Orange plus Up or Down. I was able to simulate holding and releasing Orange, but because this happens so quickly the popup menu appears. This is due to using a timer on Orange.

Ideally, I would prefer to see MenuActions added for Next Section and Previous Section. This could work on many menus (including settings) but it would require an update to Yarg.Core. I'm seeking input on this change. I could include it as part of this PR or as a separate one.

### DIFF
--- a/Assets/Script/Menu/Navigation/Navigator.cs
+++ b/Assets/Script/Menu/Navigation/Navigator.cs
@@ -1,8 +1,10 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using Cysharp.Threading.Tasks;
 using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.Controls;
 using YARG.Core.Input;
 using YARG.Input;
 using YARG.Menu.Persistent;
@@ -59,6 +61,18 @@ namespace YARG.Menu.Navigation
             MenuAction.Right,
         };
 
+        private static Keyboard menuKeyboard;
+
+        private static Dictionary<Key, MenuAction> KeyboardMenuActions = new()
+        {
+            { Key.UpArrow, MenuAction.Up },
+            { Key.DownArrow, MenuAction.Down },
+            { Key.LeftArrow, MenuAction.Left },
+            { Key.RightArrow, MenuAction.Right },
+            { Key.Enter, MenuAction.Green },
+            { Key.Escape, MenuAction.Red },
+        };
+
         public class HoldContext
         {
             public readonly NavigationContext Context;
@@ -99,8 +113,34 @@ namespace YARG.Menu.Navigation
                 }
             }
 
-            // TODO: Keyboard inputs for menus
-            // UpdateKeyboardInput();
+            // Process keyboard inputs for menus
+            UpdateKeyboardInput();
+        }
+
+        private void UpdateKeyboardInput()
+        {
+            // Get current keyboard
+            Keyboard keyboard = Keyboard.current;
+
+            // If there is no keyboard, bail
+            if (keyboard == null) { return; }
+
+            // Loop through mapped keys
+            foreach (var map in KeyboardMenuActions)
+            {
+                // Get the key control
+                KeyControl key = keyboard[map.Key];
+
+                // Just pressed or released?
+                if (key.wasPressedThisFrame)
+                {
+                    ProcessKeyboardAction(map.Value, true);
+                }
+                else if (key.wasReleasedThisFrame)
+                {
+                    ProcessKeyboardAction(map.Value, false);
+                }
+            }
         }
 
         private void ProcessInput(YargPlayer player, ref GameInput input)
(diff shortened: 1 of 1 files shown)


jbienz:
This PR addresses #810.

jbienz:
I'm thinking that I should probably add bindings for the colors too. Like B = Blue and O = Orange.

I could also add a setting that would disable these built-in keyboard navigation "bindings" in case the user wants to customize bindings manually by adding the keyboard device themselves. Though I still think these default binds should be enabled by default.

TheNathannator:
My thoughts on the comments here so far:

> For example, I'd love to see Page Down go to Next Section and Page Up go to Previous Section in the Music list.

This would certainly be a tricky thing to implement support for currently, we need to rethink how section seeking is done I think. In Rock Band, holding Orange will open a navigable menu of sections, and just pressing Orange will seek to the next section. We already have a menu for this IIRC, but it only exists in the More Options menu currently.

We can also take a page out of the Rock Band 3 Deluxe mod's book, and bind the left/right menu actions to seek up/down a section (or make it jump, say, 5 songs at a time).

> Ideally, I would prefer to see MenuActions added for Next Section and Previous Section.

This I'm a little hesitant on, as none of the instrument peripherals would be able to use them without sacrificing other menu actions. They would, however, be nice QoL actions to have for devices where it makes sense to implement (e.g. this could also be bound to the left/right triggers on gamepads, taking a page from the Xbox 360/One menus). Perhaps they can just be named `PageUp`/`PageDown` so we can be a little more flexible with how they function if needed?

> I'm thinking that I should probably add bindings for the colors too. Like B = Blue and O = Orange.

For color bindings, I'd suggest using the 1-5 number row keys, I imagine that's the most intuitive bindings for those. These can exist in addition to the current bindings on Enter/Escape.

Assets/Script/Menu/Navigation/Navigator.cs
```diff
@@ -99,8 +111,34 @@ private void Update()
...
-            // TODO: Keyboard inputs for menus
-            // UpdateKeyboardInput();
+            // Process keyboard inputs for menus
+            UpdateKeyboardInput();
+        }
+
+        private void UpdateKeyboardInput()
+        {
+            // Get current keyboard
+            Keyboard keyboard = Keyboard.current;
```
TheNathannator, line 121: I might suggest using the [`InputState.AddChangeMonitor`](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.5/api/UnityEngine.InputSystem.LowLevel.InputState.html#UnityEngine_InputSystem_LowLevel_InputState_AddChangeMonitor_UnityEngine_InputSystem_InputControl_System_Action_UnityEngine_InputSystem_InputControl_System_Double_UnityEngine_InputSystem_LowLevel_InputEventPtr_System_Int64__System_Int32_System_Action_UnityEngine_InputSystem_InputControl_System_Double_System_Int64_System_Int32__) API instead of doing manual frame-based polling, as these will allow multiple input events to be registered within a single frame (for cases like stuttering or lag spikes, our music library menu is rather inefficient currently lol).

I'm not adamant about this change though, as keyboard inputs don't seem to be threaded currently in the native input backend on Unity's side.

Assets/Script/Menu/Navigation/Navigator.cs
```diff
@@ -99,8 +111,34 @@ private void Update()
...
+
+            // If there is no keyboard, bail
+            if (keyboard == null) { return; }
```
TheNathannator, line 124: Code style: don't put single statements on the same line as `if`s (braces optional).
```suggestion
            if (keyboard == null)
            {
                return;
            }
```

Some of us used to do this, but we discourage it now since it mildly obscures the flow of code enough to make reading it more difficult (though I think the braces help with that a little); just a matter of keeping each line to only one action. (I think we still need to update the Rider config to change it to this.)

TheNathannator (approved):
And now the code side of things; looks good to me currently!

Assets/Script/Menu/Navigation/Navigator.cs
jbienz: Thank you @TheNathannator. I was unaware of InputState.AddChangeMonitor but that looks like a cool API. I'm happy to incorporate it.

I would like the defaults in this "global" bind to reflect the defaults when a keyboard is added to a profile. Other than just using the same keys, do you recommend a different path to making sure the two would remain in sync over time?

Finally, I think I should add a setting that would allow this "global binding" to be turned off. That would allow users to still add a keyboard to a profile and customize it like before without any conflicts. Do you agree?

PR #821, opened 2024-07-12
Gamepad Button Prompts

pikachufan06:
I made 3 sets of Gamepad Buttons for use with generic gamepads for the following three layouts:

**Xbox**
YARGGamepadConceptXbox

**Playstation**

YARGGamepadConceptPlaystation

**Nintendo**
YARGGamepadConceptNintendo


TheNathannator (changes requested):
Just adding the assets for these icons is not enough for a PR unfortunately. An implementation for displaying the appropriate icons for the right control would be needed alongside them.

I can look into doing that if you'd like, but there's a lot of other things we need to do lol, so no guarantees on when.

pikachufan06:
> Just adding the assets for these icons is not enough for a PR unfortunately. An implementation for displaying the appropriate icons for the right control would be needed alongside them.
> 
> I can look into doing that if you'd like, but there's a lot of other things we need to do lol, so no guarantees on when.
That's fine, I just wanted to submit them! I can submit it as an issue instead.

EliteAsian123:
I'm closing this for now because it's not a full PR. Feel free to open another one if the code for it is implemented!

PR #825, opened 2024-07-14
Countdown rework

Purplo-cf:
Top-level changes for countdown rework. Unity elements that are no longer necessary have been removed


Assets/Script/Gameplay/HUD/CountdownDisplay.cs
```diff
@@ -14,48 +14,27 @@ public class CountdownDisplay : MonoBehaviour
...
+        private Image _progressBar;
 
         [Space]
         [SerializeField]
         private CanvasGroup _canvasGroup;
 
         private Coroutine _currentCoroutine;
 
-        private int _measuresLeft;
+        private bool _displayState;
```
RileyTheFox, line 25: I think this boolean could use a slightly more clear name personally. I had to look at where the value is set to understand what it was. Something like `_displayActive` would be better I think

Assets/Script/Gameplay/HUD/CountdownDisplay.cs
```diff
@@ -69,11 +48,13 @@ public void ForceReset()
 
         private void ToggleDisplay(bool newState)
```
RileyTheFox, line 49: If going through with the state variable rename, this should also be `bool active` or something.

RileyTheFox (commented):
Besides the small comments, looks good. I think for some extra customisation we could have a setting to change the display into a seconds countdown instead of measures? I know some people prefer a countdown in seconds, and Clone Hero has this option too, so it would be nice to have here.

Display setting should default to measures and have the option to change it to seconds.

RileyTheFox:
@EliteAsian123 @kaduwaengertner what do you think regarding the suggestion above?

EliteAsian123:
I'm always for customization 👍

kaduwaengertner:
I like the idea

Purplo-cf:
The new setting and the requested fixes are in. This will need the latest `countdown-rework-core` changes to run

EliteAsian123:
Ah shoot wait, I'm dumb. I made some changes to the countdown prefab on `engine-refactor` because I didn't see that this PR changes those. 

Essentially I just made sure to change all `Transform`s to `RectTransform`s on the prefab and removed the scaling on the transforms as that's bad practice. I'll revert the changes on the main branch to prevent potentially broken prefabs from merging.

Purplo-cf:
No worries! I did a rebase this morning to double check that everything still looked okay. Okay had to make a few small adjustments

RileyTheFox (changes requested):
Some issues I've noticed:

Something has corrupted and the lyric bar is just a white box now. I got a warning in Unity when checking out the branch "Broken text PPtr in file". It was referencing fileID `6635242913724598267` which is found as the m_Father in some places in CountdownDisplay.prefab

image

Also in practice mode, if you select a section that occurs after a countdown earlier in the song, it will show the last part of the countdown. However this is a general engine issue with practice mode that has only come up now (queued updates being generated from time 0 up to the practice section time) so I will handle this at a later point.

RileyTheFox (approved):

PR #827, opened 2024-07-20
Disable SP in practice

Purplo-cf:
Adds a setting to automatically strip star power from notes while in Practice mode. On by default.

Requires YARG.Core branch `disable-sp-core` to run

### DIFF
--- a/Assets/StreamingAssets/lang/en-US.json
+++ b/Assets/StreamingAssets/lang/en-US.json
@@ -537,7 +537,7 @@
             },
             "EnablePracticeSP": {
                 "Name": "Star Power in Practice",
-                "Description": "Allow Star Power phrases and activation while in Practice Mode."
+                "Description": "Allow Star Power phrases while in Practice Mode."
             },
             "FpsCap": {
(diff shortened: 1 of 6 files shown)


Assets/StreamingAssets/lang/en-US.json
```diff
@@ -495,6 +495,10 @@
                 "Name": "Drums Volume",
                 "Description": "Adjusts the song's drum track volume. Only does something if the song is multi-track."
             },
+            "EnablePracticeSP": {
+                "Name": "Star Power in Practice",
+                "Description": "Allow Star Power phrases and activation while in Practice Mode."
```
RileyTheFox, line 500: Remove the text specifying it allows activations. These are not directly supported but are possible on drums only because of activation notes which is unintentional and a bug.

A practice mode that allows for activation and path practicing could be considered in the future but for now it is not supported and will be patched in the future.

RileyTheFox (changes requested):

RileyTheFox (approved):

PR #828, opened 2024-07-20
More missing localization

Purplo-cf:
Added missing localization keys for menu headers and actions in the Music Library popup menu and on the Difficulty Select screen.

### DIFF
--- a/Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
+++ b/Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
@@ -86,7 +86,8 @@ namespace YARG.Menu.DifficultySelect
 
         private void OnEnable()
         {
-            _subHeader.text = GlobalVariables.State.IsPractice ? "Practice" : "Quickplay";
+            string subHeaderKey = GlobalVariables.State.IsPractice ? "Practice" : "Quickplay";
+            _subHeader.text = Localize.Key("Main.Options", subHeaderKey);
 
             // Set navigation scheme
             Navigator.Instance.PushScheme(new NavigationScheme(new()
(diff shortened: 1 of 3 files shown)


Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
```diff
@@ -171,14 +172,16 @@ private void CreateMainMenu()
             // Only show all these options if there are instruments available
             if (_possibleInstruments.Count > 0)
             {
-                CreateItem("Instrument", player.Profile.CurrentInstrument.ToLocalizedName(),
+                CreateItem(Localize.Key("Menu.DifficultySelect.Header.Instrument"),
```
EliteAsian123, line 175: It might be worth making this into `CreateLocalizedItem` like in the music library popup

YARG.Core
EliteAsian123, line 1: These core changes aren't required for the PR!

EliteAsian123 (changes requested):

Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
Purplo-cf: I started going down that route here and the added complexity for the menu items compared to `PopupMenu` make that tough without making more drastic changes. We would need different overloads for every combination of header and body text needing or not needing localization, or header doesn't exist, etc.

I did add new `CreateLocalizedHeader` and `CreateLocalizedAction` methods so that the hard-coded addresses for `Menu.DifficultySelect.Header|Action` are only in here once. These are still getting passed to a `CreateItem` that doesn't handle localization.

Will that work for adding support to this menu?

PR #829, opened 2024-07-21
Remove the 70% scaling limiter on song stems

sonicfind:
Requires core PR

### DIFF
--- a/Assets/Script/Audio/Bass/BassAudioManager.cs
+++ b/Assets/Script/Audio/Bass/BassAudioManager.cs
@@ -41,14 +41,7 @@ namespace YARG.Audio.BASS
                 YargLogger.LogFormatError("Failed to create split stream: {0}!", Bass.LastError);
                 return null;
             }
-
-            var handle = new StreamHandle(BassFx.TempoCreate(streamSplit, tempoFlags));
-            handle.CompressorFX = BassHelpers.AddCompressorToChannel(handle.Stream);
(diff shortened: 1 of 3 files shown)


Assets/Script/Audio/Bass/BassAudioManager.cs
```diff
@@ -348,12 +341,20 @@ private static string GetBassDirectory()
...
+            // The float flag allows >0dB signals.
+            // Note that the compressor attempts to normalize signals >-2dB, but some mixes will pierce through.
+            mixerHandle = BassMix.CreateMixerStream(44100, 2, BassFlags.Float);
             if (mixerHandle == 0)
             {
                 YargLogger.LogFormatError("Failed to create mixer: {0}!", Bass.LastError);
                 return false;
             }
+
+            int CompressorFX = BassHelpers.AddCompressorToChannel(mixerHandle);
```
EliteAsian123, line 353: Is this supposed to be a local variable, or is it supposed to reference the field? If it's the former, make sure it use `camelCase`.

EliteAsian123 (changes requested):

Assets/Script/Audio/Bass/BassAudioManager.cs
sonicfind: ah yea. I copied and pasted

EliteAsian123 (approved):

PR #830, opened 2024-07-22
Music Library UX improvements

Purplo-cf:
- Go To Section popup will now combine entries for Artist/Album/Charter/Artist+Album sorts that begin with either digits or symbols. The logic for creating these groupings will only be run once per song cache build.
- LINQ queries for creating Go To Section shortcuts will only need to run once per Search/Filter/Sort change.
- Sort By and Go To Section popup actions will not be displayed if there are no `SortHeaderViewType` instances in the library (e.g., results from typing in search field)
- Changing Sort By type while highlighting a song in the Recommended Songs section will now jump to that song's position in the main library instead of staying in the recommended section.

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -71,6 +71,8 @@ namespace YARG.Menu.MusicLibrary
         protected override int ExtraListViewPadding => 15;
         protected override bool CanScroll => !_popupMenu.gameObject.activeSelf;
 
+        public bool HasSortHeaders { get; private set; }
+
         private SongCategory[] _sortedSongs;
         private IEnumerable<IGrouping<string, (ViewType, int)>> _shortcutQuery;
 
@@ -209,12 +211,11 @@ namespace YARG.Menu.MusicLibrary
             // Shortcuts will be re-queried every time the list is refreshed
             _shortcutQuery = null;
 
-            if (SelectedPlaylist is not null)
-            {
-                return CreatePlaylistViewList();
-            }
+            var viewList = (SelectedPlaylist is not null) ? CreatePlaylistViewList() : CreateNormalViewList();
 
-            return CreateNormalViewList();
+            HasSortHeaders = viewList.Any(x => x is SortHeaderViewType);
+
+            return viewList;
         }
 
         private List<ViewType> CreateNormalViewList()
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -72,6 +72,7 @@ namespace YARG.Menu.MusicLibrary
         protected override bool CanScroll => !_popupMenu.gameObject.activeSelf;
 
         private SongCategory[] _sortedSongs;
+        private IEnumerable<IGrouping<string, (ViewType, int)>> _shortcutQuery;
 
         private CancellationTokenSource _previewCanceller;
         private PreviewContext _previewContext;
@@ -205,6 +206,9 @@ namespace YARG.Menu.MusicLibrary
 
         protected override List<ViewType> CreateViewList()
         {
+            // Shortcuts will be re-queried every time the list is refreshed
+            _shortcutQuery = null;
+
             if (SelectedPlaylist is not null)
             {
                 return CreatePlaylistViewList();
@@ -251,7 +255,7 @@ namespace YARG.Menu.MusicLibrary
                         displayName = SongSources.Default.GetDisplayName();
                     }
                 }
-                list.Add(new SortHeaderViewType(displayName, section.Songs.Length));
+                list.Add(new SortHeaderViewType(displayName, section.Songs.Length, section.Shortcut));
 
                 // Add all of the songs
                 list.AddRange(section.Songs.Select(song => new SongViewType(this, song)));
@@ -593,9 +597,16 @@ namespace YARG.Menu.MusicLibrary
             UpdateSearch(true);
         }
 
-        public IEnumerable<(ViewType, int)> GetSections()
+        public IEnumerable<IGrouping<string, (ViewType, int)>> GetShortcuts()
         {
-            return ViewList.Select((v, i) => (v, i)).Where(i => i.v is SortHeaderViewType);
+            if (_shortcutQuery == null)
+            {
+                _shortcutQuery = ViewList.Select((v, i) => (v, i))
+                   .Where(i => i.v is SortHeaderViewType)
+                   .GroupBy(g => ((SortHeaderViewType) g.Item1).ShortcutName);
+            }
+
+            return _shortcutQuery;
         }
 
         public void SetSearchInput(SortAttribute songAttribute, string input)
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -84,6 +84,8 @@ namespace YARG.Menu.MusicLibrary
 
         private List<int> _sectionHeaderIndices = new();
         private List<HoldContext> _heldInputs = new();
+
+        private int _primaryHeaderIndex;
 
         protected override void Awake()
         {
@@ -262,23 +264,24 @@ namespace YARG.Menu.MusicLibrary
                 list.AddRange(section.Songs.Select(song => new SongViewType(this, song)));
             }
 
+            CategoryViewType primaryHeader;
+
             if (_searchField.IsSearching)
             {
                 // If the current search is NOT empty...
 
                 // Create the category
-                var categoryView = new CategoryViewType(
-                    Localize.Key("Menu.MusicLibrary.SearchResults"), count, _sortedSongs);
+                primaryHeader = new CategoryViewType(Localize.Key("Menu.MusicLibrary.SearchResults"), count, _sortedSongs);
 
                 if (_sortedSongs.Length == 1)
                 {
                     // If there is only one header, just replace it
-                    list[0] = categoryView;
+                    list[0] = primaryHeader;
                 }
                 else
                 {
                     // Otherwise add to the very top
-                    list.Insert(0, categoryView);
+                    list.Insert(0, primaryHeader);
                 }
             }
             else
@@ -286,13 +289,13 @@ namespace YARG.Menu.MusicLibrary
                 if (SettingsManager.Settings.LibrarySort < SortAttribute.Playable)
                 {
                     // Add "ALL SONGS" header right above the songs
-                    list.Insert(0, new CategoryViewType(Localize.Key("Menu.MusicLibrary.AllSongs"),
-                        SongContainer.Count, SongContainer.Songs));
+                    primaryHeader = new CategoryViewType(Localize.Key("Menu.MusicLibrary.AllSongs"), SongContainer.Count, SongContainer.Songs);
+                    
+                    list.Insert(0, primaryHeader);
 
                     if (_recommendedSongs != null)
                     {
                         // Add the recommended songs right above the "ALL SONGS" header
-
                         list.InsertRange(0, _recommendedSongs.Select(i => new SongViewType(this, i)));
                         list.Insert(0, new CategoryViewType(
                             Localize.Key("Menu.MusicLibrary.RecommendedSongs",
@@ -308,8 +311,8 @@ namespace YARG.Menu.MusicLibrary
                 }
                 else
                 {
-                    list.Insert(0, new CategoryViewType(
-                        Localize.Key("Menu.MusicLibrary.PlayableSongs"), count, _sortedSongs));
+                    primaryHeader = new CategoryViewType(Localize.Key("Menu.MusicLibrary.PlayableSongs"), count, _sortedSongs);
+                    list.Insert(0, primaryHeader);
                 }
 
                 // Add the buttons
@@ -325,6 +328,8 @@ namespace YARG.Menu.MusicLibrary
                     Refresh();
                 }, PLAYLIST_ID));
             }
+
+            _primaryHeaderIndex = list.IndexOf(primaryHeader);
 
             CalculateCategoryHeaderIndices(list);
             return list;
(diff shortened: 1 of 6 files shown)


Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -71,7 +71,10 @@ public static void SetReload(MusicLibraryReloadState state)
         protected override int ExtraListViewPadding => 15;
         protected override bool CanScroll => !_popupMenu.gameObject.activeSelf;
 
+        public bool HasSortHeaders { get; private set; }
+
         private SongCategory[] _sortedSongs;
+        private IEnumerable<IGrouping<string, (ViewType, int)>> _shortcutQuery;
```
Purplo-cf, line 77: Is there a way to create an alias for `IEnumerable<IGrouping<string, (ViewType, int)>>` that I can use as a type? It would be great if we could just call this `CategoryGroupQuery` instead.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -593,9 +609,16 @@ public void ChangeSort(SortAttribute sort)
...
 
-        public IEnumerable<(ViewType, int)> GetSections()
+        public IEnumerable<IGrouping<string, (ViewType, int)>> GetShortcuts()
         {
-            return ViewList.Select((v, i) => (v, i)).Where(i => i.v is SortHeaderViewType);
+            if (_shortcutQuery == null)
+            {
+                _shortcutQuery = ViewList.Select((v, i) => (v, i))
+                   .Where(i => i.v is SortHeaderViewType)
+                   .GroupBy(g => ((SortHeaderViewType) g.Item1).ShortcutName);
```
sonicfind, line 618: Been meaning to do this, but it'd be beneficial if we manually constructed the container. The current way is done relies on a functionality that isn't immediately obvious. Its not really communicated clearly that `Select` can return the index of the value alongside the value.

We could instead just hold a `List<(string, int)>`, dodging all the need for these interfaces and wack memory & performance degradations. Only needs a for loop. The condition for whether to add a new node would simply equate to `list.count == 0 || list[^1].Group != curr.Group`

Assets/Script/Song/SongContainer.cs
```diff
@@ -190,6 +200,38 @@ public static SongCategory[] GetSortedCategory(SortAttribute sort)
...
+                case "Artist":
+                case "Album":
+                case "Charter":
+                case "Artist_Album":
+                    return true;
+                
+                default:
+                    return false;
+            }
+        }
```
sonicfind, line 216: Don't know why you're using strings when we have the enums

Assets/Script/Song/SongContainer.cs
```diff
@@ -57,12 +57,22 @@ public enum SortAttribute
...
             Songs = songs;
+
+            if (createGroup)
+            {
+                CategoryGroup = SongContainer.CreateCategoryGroup(category);
+            }
+            else
+            {
+                CategoryGroup = category;
+            }
```
sonicfind, line 75: Again, don't do this in the constructor *at all.* The constructor should just take the string Group variable directly. That's it. Anything else complicates other parts of the system that may attempt to use the type.

Assets/Script/Song/SongSearching.cs
```diff
@@ -283,6 +284,7 @@ private static SongCategory[] SearchSongs(in FilterNode filter, SongCategory[] s
                 return SearchInstrument(filter, searchList);
             }
 
+            bool createCategoryGroups = SongContainer.AttributeShouldBeGrouped(SettingsManager.Settings.LibrarySort.ToString());
```
sonicfind, line 287: Case and point: don't do this *whatsoever*. All the work was done already by SongContainer. Doing it again for every category in the search is unnecessary/wasteful. Just pass the same category string to the new nodes.

Assets/Script/Song/SongContainer.cs
sonicfind: And there's a general reason why I specified to keep it scoped to `FillContainers()`. You mentioned that there *may* be a point where you'd use this outside of that.

But that's an assumption, and in doing so, you were tempted to still use this in an additional context where it was not needed. It's generally preferred to keep very specific non-general functionality scoped to where it is directly used if it's only used in one spot. Only once you come across a situation where having this accessible outside the class would be necessary should you publicize it.

Don't pollute a namespace with extra functionality based on if you *could* use it elsewhere but rather if you *need* to use it. And currently, we don't.

Assets/Script/Song/SongContainer.cs
```diff
@@ -383,6 +425,7 @@ static SongEntry[] SetAllSongs(Dictionary<HashWrapper, List<SongEntry>> entries)
             static SongCategory[] Convert(SortedDictionary<SortString, List<SongEntry>> list, SongAttribute attribute)
             {
                 var sections = new SongCategory[list.Count];
+                bool createCategoryGroups = AttributeShouldBeGrouped(attribute.ToString());
```
sonicfind, line 428: Your prior iteration with passing a bool to the Convert functions was the better option. Resorting to comparing by strings is usually not a good move.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
sonicfind: See my comment on `GetShortcuts()`

Assets/Script/Menu/MusicLibrary/ViewTypes/SortHeaderViewType.cs
```diff
@@ -11,12 +11,15 @@ public class SortHeaderViewType : ViewType
         public override bool UseWiderPrimaryText => true;
 
         public readonly string HeaderText;
+        public readonly string ShortcutName;
         private readonly int _songCount;
 
-        public SortHeaderViewType(string headerText, int songCount)
+        public SortHeaderViewType(string headerText, int songCount, string shortcutName = null)
```
sonicfind, line 17: This will never be null

sonicfind (changes requested):

Assets/Script/Song/SongContainer.cs
```diff
@@ -317,16 +321,47 @@ private static void UpdateSongUi(LoadingContext context)
...
+
+            _sortGenres    = Convert(_songCache.Genres, (genreName) => {
+                if (genreName.Length > 0 && char.IsLower(genreName[0]))
+                {
+                    genreName = char.ToUpperInvariant(genreName[0]).ToString();
+                    if (genreName.Length > 1)
+                        genreName += genreName[1..];
+                }
+
+                return genreName;
```
sonicfind, line 355: This being in the `Convert` function directly makes the interface for calling it much simpler. If you were to fix the `null` issue while still going with passing functions, you'd essentially need to pass a new instance of a function that would just return the string that was passed into it for the attribute types that don't use shortcuts. Additionally, it'd need to be the the non-formatless form (as they'd all be unique anyways).

Assets/Script/Song/SongContainer.cs
```diff
@@ -335,7 +370,8 @@ private static void FillContainers()
                 int index = 0;
                 foreach (var node in _songCache.DatesAdded)
                 {
-                    _sortDatesAdded[index++] = new(node.Key.ToLongDateString(), node.Value.ToArray());
+                    var dateTime = node.Key;
+                    _sortDatesAdded[index++] = new(dateTime.ToLongDateString(), node.Value.ToArray(), dateTime.ToString("MMMM yyyy"));
```
sonicfind, line 374: Use `"y"` as the ToString format. That generates the month and year.

Also, `node.Key` is already a temporary local due to how the `Dictionary` enumerator works with `KeyValuePair`, so making a separate one for DateTime is just extra work.

Assets/Script/Song/SongContainer.cs
```diff
@@ -380,31 +416,28 @@ static SongEntry[] SetAllSongs(Dictionary<HashWrapper, List<SongEntry>> entries)
...
                 {
-                    string key = node.Key;
-                    if (attribute == SongAttribute.Genre && key.Length > 0 && char.IsLower(key[0]))
-                    {
-                        key = char.ToUpperInvariant(key[0]).ToString();
-                        if (node.Key.Length > 1)
-                            key += node.Key.Str[1..];
-                    }
-                    sections[index++] = new SongCategory(key, node.Value.ToArray());
+                    string key = node.Key.Str;
```
sonicfind, line 426: SortString has a operator that auto grabs `Str`. Not necessarily telling you to swap back, but just saying if you weren't already aware.

Assets/Script/Song/SongContainer.cs
```diff
@@ -380,31 +416,28 @@ static SongEntry[] SetAllSongs(Dictionary<HashWrapper, List<SongEntry>> entries)
                 return songs;
             }
 
-            static SongCategory[] Convert(SortedDictionary<SortString, List<SongEntry>> list, SongAttribute attribute)
+            static SongCategory[] Convert(SortedDictionary<SortString, List<SongEntry>> list, Func<string, string> GenerateGroupName = null)
```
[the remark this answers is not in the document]
sonicfind: Now, there *is* a method to using a function instead that would be more fine in my eyes, but that requires some unsafe shenanigans here (and would start getting pretty ugly maintainability-wise). My main concern is trying to dodge needless GC heap allocations of function objects, which a boolean does not do.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -368,9 +380,33 @@ private void ExitPlaylistTab()
 
         private void CalculateCategoryHeaderIndices(List<ViewType> list)
         {
-            _sectionHeaderIndices = list.Select((v, i) => (v, i)).
-                Where(viewTypeEntry => viewTypeEntry.v is SortHeaderViewType or CategoryViewType).
-                Select(viewTypeEntry => viewTypeEntry.i).ToList();
+            _sectionHeaderIndices = new();
+            Shortcuts = new();
```
sonicfind, line 384: Don't reallocate new lists. Call `Clear()` instead.

Assets/Script/Song/SongContainer.cs
```diff
@@ -57,12 +58,15 @@ public enum SortAttribute
     public readonly struct SongCategory
     {
         public readonly string Category;
+        public readonly string CategoryGroup;
         public readonly SongEntry[] Songs;
 
-        public SongCategory(string category, SongEntry[] songs)
+        public SongCategory(string category, SongEntry[] songs, string categoryGroupName = null)
```
sonicfind, line 64: Essentially, there shouldn't be any situation where this is `null` *except* for "Search Results", since that doesn't even get used to begin with. It'd be best to take off the parameter default.

sonicfind (changes requested):
Sorry I keep dumping more on ya.

Purplo-cf:
> Sorry I keep dumping more on ya.

Keep it coming lol

I really appreciate you taking the time to analyze this, I'm learning a lot!

Assets/Script/Song/SongContainer.cs
```diff
@@ -380,34 +385,59 @@ static SongEntry[] SetAllSongs(Dictionary<HashWrapper, List<SongEntry>> entries)
                 return songs;
             }
 
-            static SongCategory[] Convert(SortedDictionary<SortString, List<SongEntry>> list, SongAttribute attribute)
+            static SongCategory[] Convert(SortedDictionary<SortString, List<SongEntry>> list, bool capitalizeFirstChar = false, bool createCategoryGroups = false)
```
sonicfind, line 388: Revert back to passing the SortAttribute instead of the `capitalizeFirstChar` boolean. After this PR, there is something else I need to add that relies on a specific sort.

Assets/Script/Song/SongContainer.cs
```diff
@@ -350,7 +355,7 @@ private static void FillContainers()
                     {
                         arr[index++] = new SongCategory(
                             $"{instrument.Key.ToSortAttribute().ToLocalizedName()} [{difficulty.Key}]",
-                            difficulty.Value.ToArray());
+                            difficulty.Value.ToArray(), difficulty.Key.ToString());
```
sonicfind, line 358: The category name and shortcut need to be the same for instruments.

Assets/Script/Song/SongContainer.cs
```diff
@@ -380,34 +385,59 @@ static SongEntry[] SetAllSongs(Dictionary<HashWrapper, List<SongEntry>> entries)
...
             {
                 var sections = new SongCategory[list.Count];
+                
                 int index = 0;
                 foreach (var node in list)
                 {
                     string key = node.Key;
-                    if (attribute == SongAttribute.Genre && key.Length > 0 && char.IsLower(key[0]))
+
+                    if (capitalizeFirstChar)
```
sonicfind, line 397: Also, `attribute == SongAttribute.Genre && key.Length > 0 && char.IsLower(key[0])` was very important. It'll error out in certain situations without it. One of the oldest song scanning bugs we had.

sonicfind (changes requested):

Assets/Script/Song/SongContainer.cs
```diff
@@ -380,34 +383,58 @@ static SongEntry[] SetAllSongs(Dictionary<HashWrapper, List<SongEntry>> entries)
...
+                    string categoryGroupName = createCategoryGroups ? GenerateGroupName(SortString.RemoveArticle(key)) : key;
+                    sections[index++] = new SongCategory(key, section.ToArray(), categoryGroupName);
                 }
                 return sections;
             }
+
+            static string GenerateGroupName(string singleCategoryName)
+            {
+                char first = singleCategoryName[0];
+
```
sonicfind, line 424: Need a special case to handle strings of zero length - preferably, just have it return `string.empty`. If you don't, `[0]` throws.

Assets/Script/Song/SongContainer.cs
```diff
@@ -380,34 +383,58 @@ static SongEntry[] SetAllSongs(Dictionary<HashWrapper, List<SongEntry>> entries)
...
+                {
+                    return first.ToAsciiUpper().ToString();
+                }
+                else if (char.IsDigit(first))
+                {
+                    return "0-9";
+                }
+                else
+                {
+                    return ".";
```
sonicfind, line 435: Return the character itself instead of `.`. The number of characters that'll appear outside the usual range isn't enough in the normal case to warrant this level of truncation (and also means non-Ascii characters can't be shown).

YARG.Core
sonicfind, line 1: During the rebase, try to make sure the YARG.Core update is taken out your PR. It'd be an accidental downgrade.

sonicfind (changes requested):
Some final touches for after you rebase the branch on top of some recent adjustments.

PR #833, opened 2024-07-26
Refactor debug text display into an IMGUI menu

TheNathannator:
Lots more information is now displayed in a more organized manner.

The display starts off compact and unobtrusive:
image

Clicking each of the options expands the window to reveal information about that option:
image image image

The window can be dragged around, and clicking the Close button will reset the window to its default state and position (toggling the window with Ctrl + Tab does *not* reset it currently.)


EliteAsian123 (approved):
I love it.

Although I approve of this PR in its current state, I will note that it may be better to move the "state/stats to string" parts into their respective classes (`GuitarEngineState`, etc.) instead so it can be reused in the replay analyzer for example. Something similar is already done for the engine parameters.

TheNathannator:
Maybe. Typically I'd re-implement that manually since not all places that need the printed state/stats will necessarily need the *same* fields to be printed, or in the same order/formatting. Too many variables to say for certain that it should be made into a common routine.

EliteAsian123:
Yeah, that's fair. The only reason why I mention this is a "problem" where, if a new state variable is added, it's hard to remember what things you gotta update as a result of that, especially from the perspective of a new contributor. Because of this I usually prefer to keep things near/in the class but... this is debug text who really cares lol.

EliteAsian123:
As I mentioned in the Discord you'll probably need to manually redo the gameplay scene changes, but after that I'll merge it (if it's done)

TheNathannator:
Scene changes were simple enough to recover by hand, but there's some errors I need to resolve which require YARG.Core changes (or removal of certain debug lines).

TheNathannator:
Exposed more engine state/stat variables and re-added the removed display lines. Also made it dynamically change the max width/height allowed for the debug window based on the window size, for a better experience in larger windows.

Everything should be good to go now!

PR #836, opened 2024-07-30
Golden note edges for SP phrases

Purplo-cf:
Change the color and material intensity of note metal for notes in Starpower phrases. Gold color by default to match the look of the RB series.


Purplo-cf:
After the discussion in the core PR, I have taken out all customization for the metal/gold color. Color is now being controlled by the static property `SP_GOLD_COLOR` in `NoteRef`. All changes are now contained in YARG without any core changes.

PR #839, opened 2024-08-01
Library popup hotfix

Purplo-cf:
Allow changing library sort when only one sort header is present, prevent *Go To* instead when there aren't enough headers

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -216,7 +216,9 @@ namespace YARG.Menu.MusicLibrary
 
             var viewList = (SelectedPlaylist is not null) ? CreatePlaylistViewList() : CreateNormalViewList();
 
-            HasSortHeaders = viewList.Any(x => x is SortHeaderViewType);
+            // Disable shortcuts if there are less than 2 sort headers in the viewlist
+            HasSortHeaders = viewList.Where(x => x is SortHeaderViewType)
+                             .ElementAtOrDefault(1) != null;
 
             return viewList;
         }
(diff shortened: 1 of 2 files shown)


Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -216,7 +216,9 @@ protected override List<ViewType> CreateViewList()
 
             var viewList = (SelectedPlaylist is not null) ? CreatePlaylistViewList() : CreateNormalViewList();
 
-            HasSortHeaders = viewList.Any(x => x is SortHeaderViewType);
+            // Disable shortcuts if there are less than 2 sort headers in the viewlist
+            HasSortHeaders = viewList.Where(x => x is SortHeaderViewType)
+                             .ElementAtOrDefault(1) != null;
```
sonicfind, line 221: You can just do `_sortedSongs.Count >= 2`

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -216,7 +216,8 @@ protected override List<ViewType> CreateViewList()
 
             var viewList = (SelectedPlaylist is not null) ? CreatePlaylistViewList() : CreateNormalViewList();
 
-            HasSortHeaders = viewList.Any(x => x is SortHeaderViewType);
+            // Disable shortcuts if there are less than 2 sort headers in the viewlist
+            HasSortHeaders = _sortedSongs is not null && _sortedSongs.Length > 1;
```
sonicfind, line 220: It's funny how this null check *shouldn't* be necessary, but our ListMenu implementation makes this function be called basically twice on initial creation of the menu - one of which being before any entries are added. It's great.

PR #841, opened 2024-08-05
Add Active Players and Bots displays to Status Bar

thomeval:
This PR adds two displays to the top Status Bar to display the currently active players, as well as the number of active bot profiles. These can be toggled in the Settings menu using the "Show Active Players" and "Show Active Bots" options.

image

Notes:

- Show Active Players is enabled by default. Show Active Bots is not.
- Only the number of active bots is shown, since their names are probably not useful.
- The active bots display is hidden automatically if no bot profiles are active (similar to the Battery status display).
- When 1 to 3 players are connected, their instrument icons and profile names are shown.
- When 4 or more players are connected, only their instrument icons are shown.
- However, when no players are connected (including bots), the active players list will instead show a "No profiles active" message.
- The current implementation is not very efficient, as it Destroys and Instantiates prefabs on every update (once a second). Advice on how to address this (or leave it as-is) is welcome.


.editorconfig
EliteAsian123, line 1: Were these changes intended to be a part of this PR?

Assets/ActivePlayerListItem.cs
```diff
@@ -0,0 +1,41 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+using YARG.Core.Game;
+
+namespace YARG
+{
+    public class ActivePlayerListItem : MonoBehaviour
+    {
+        [SerializeField]
+        private TextMeshProUGUI _PlayerNameText;
```
EliteAsian123, line 11: Should be `_playerNameText`

Assets/ActivePlayerList.cs
```diff
@@ -0,0 +1,39 @@
...
+using UnityEngine;
+using YARG.Player;
+
+namespace YARG
+{
+    public class ActivePlayerList : MonoBehaviour
+    {
+        public GameObject NoPlayersContainer;
+        public GameObject PlayerNamesContainer;
+        public GameObject PlayerNamesPrefab;
```
EliteAsian123, line 12: These can all be made into private `SerializeField`s

Assets/ActivePlayerList.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public int MaxShownPlayerNames = 3;
```
EliteAsian123, line 14: Same with this, or turn it into a constant

Assets/Script/Menu/Persistent/StatsManager.cs
```diff
@@ -79,6 +92,8 @@ public enum Stat
 
         private float _nextUpdateTime;
 
+        private int _activeBotCount => PlayerContainer.Players.Count(p => p.Profile.IsBot);
```
EliteAsian123, line 95: Properties should be name like this: `ActiveBotCount`

Assets/Script/Menu/Persistent/StatsManager.cs
```diff
@@ -88,11 +103,13 @@ private GameObject GetStat(Stat stat)
...
-                Stat.Memory  => _memoryStats,
-                Stat.Time    => _time,
+                Stat.FPS => _fpsCounter,
+                Stat.Memory => _memoryStats,
+                Stat.Time => _time,
                 Stat.Battery => _battery,
-                _            => throw new Exception("Unreachable.")
+                Stat.ActiveBots => _activeBots,
+                Stat.ActivePlayers => _activePlayerList.gameObject,
+                _ => throw new Exception("Unreachable.")
```
EliteAsian123, line 112: Switch case formatting was removed

EliteAsian123 (changes requested):
Thanks for the PR!

> The current implementation is not very efficient, as it Destroys and Instantiates prefabs on every update (once a second). Advice on how to address this (or leave it as-is) is welcome.

What you can do is, instead of checking every second, you can only when there are changes to active players. This is probably best done in the `PlayerContainer` class, and there is even [a method that gets called whenever there are player changes](https://github.com/YARC-Official/YARG/blob/8eb600f56c3a6c43eb764235b04e4c1af3586375/Assets/Script/Player/PlayerContainer.cs#L137). You can add on to this method and call something within `StatsManager` that then updates the active players only when there are changes. This would be WAY more efficient since creating and destroying objects in Unity is very expensive.

I've also left some comments on the code changes, however, I'll look into it within Unity once the fixes are made. Thanks!

.editorconfig
thomeval: Yes. The one change I needed to add was `csharp_prefer_braces = false:suggestion` , or VS would "fix" all the one-liner if statements. The rest were added automatically once I opened the .editorconfig file and I left them at defaults, but I think it might be good to have these all here for consistency, especially if other team members are using different IDEs.

Assets/ActivePlayerListItem.cs
thomeval: Oops. I'll add an .editorconfig rule for this.

Assets/Script/Menu/Persistent/StatsManager.cs
thomeval: I'll add an .editorconfig rule for this.

thomeval:
Thanks for the feedback. I'll make the changes discussed above and get back to you.

Assets/ActivePlayerList.cs
thomeval: Resolved

Assets/Script/Menu/Persistent/StatsManager.cs
thomeval: Resolved

Assets/Script/Menu/Persistent/StatsManager.cs
thomeval: Resolved

Assets/ActivePlayerList.cs
thomeval: Resolved

thomeval:
I've updated this branch with the requested changes, and did a quick test to see if everything is still working as it should. I've needed to update ProfileListMenu as well in order to fix a bug where player instrument icons weren't updating when players change their Game Mode.

.editorconfig
EliteAsian123: The problem with that is it looks like it's adding incorrect keys at the bottom (I may be wrong) for non-C# files, but the rules appear to be for C#.

EliteAsian123: Oh also, some of the rules are incorrect/changed such as this one:
image

Assets/ActivePlayerList.cs
```diff
@@ -0,0 +1,43 @@
...
+        private GameObject _noPlayersContainer;
+        [SerializeField]
+        private GameObject _playerNamesContainer;
+        [SerializeField]
+        private GameObject _playerNamesPrefab;
+
+        [SerializeField]
+        private int _maxShownPlayerNames = 3;
+
+        public void UpdatePlayerList(IReadOnlyCollection<YargPlayer> players)
```
EliteAsian123, line 20: This method only has one use, which provides the static field `PlayerContainer.Players`. Instead, this parameter should be removed and `PlayerContainer.Players` should be used directly instead (since I don't think there's a case where a different list is used).

Assets/Script/Player/PlayerContainer.cs
```diff
@@ -140,6 +139,8 @@ private static void ResetPlayableSongs()
             {
                 MusicLibraryMenu.SetReload(MusicLibraryReloadState.Full);
             }
+
+            StatsManager.Instance.UpdateActivePlayers();
```
EliteAsian123, line 143: This method should be renamed to something more generic such as `ActiveProfilesChanged` or something like that

Assets/Script/Menu/ProfileList/ProfileListMenu.cs
```diff
@@ -37,6 +38,8 @@ private void OnDisable()
         {
             PlayerContainer.SaveProfiles();
 
+            // Update player icons if a profile has changed its GameMode.
+            StatsManager.Instance.UpdateActivePlayers();
```
EliteAsian123, line 42: This is kinda hacky but there isn't really another way to watch for game mode changes, so it should be fine for now

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -67,6 +67,8 @@ public void OpenVenueFolder()
             public ToggleSetting ShowBattery { get; } = new(false);
             public ToggleSetting ShowTime { get; } = new(false, ShowTimeCallback);
             public ToggleSetting MemoryStats { get; } = new(false, MemoryStatsCallback);
+            public ToggleSetting ShowActivePlayers { get; } = new(true, ShowActivePlayersCallback);
```
EliteAsian123, line 70: This setting shouldn't be enabled by default in my opinion, since none of the other stats are by default.

.editorconfig
EliteAsian123, line 1: (see comments on the initial review of this file)

EliteAsian123 (changes requested):
Hello, thanks for the changes!

I did a little more of a thorough review this time so there is a lot. Sorry about that.

Please let me know if you need any assistance! Thanks.

.editorconfig
thomeval: That's odd. It works fine on my end:
image

I will look into this some more, but I suspect that the two rules in place for "private fields" and "constants" are clashing with each other, and different IDE setup have different ways of prioritizing which rule applies to private constants.

thomeval: Okay, I suggest we take the .editorconfig changes out of this PR. I'd rather not have this issue delay things, and I can always submit a separate PR for that later.

Assets/Script/Settings/SettingsManager.Settings.cs
thomeval: Fair point. I don't necessarily agree, but I'll defer to your judgement on this one.

Assets/ActivePlayerListItem.cs
```diff
@@ -0,0 +1,41 @@
...
+            }
+        }
+
+        public bool ShowName
+        {
+            get => _playerNameText.gameObject.activeSelf;
+            set => _playerNameText.gameObject.SetActive(value);
+        }
+        [SerializeField]
+        private Sprite[] _gameModeIcons = new Sprite[0];
```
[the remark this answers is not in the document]
thomeval: During testing I found this change caused a regression. I'll DM you to discuss this further.

PR #843, opened 2024-08-10
Add note speed reduction by difficulty

Hubbubble1:
This PR adds a note speed reduction for all players/bots playing charts at a lower difficulty. This can be toggled in the Settings menu using the "Reduce Note Speed by Difficulty" option.

[[video]](https://www.youtube.com/watch?v=ucWM8VI1ZhE)

Notes:
- Reduce Note Speed by Difficulty is enabled by default.
- The vocal track is not affected by this option: it remains the same speed on all difficulties.
- The specified note speed is applied on Expert and Expert+, and the note speeds are reduced by 25% for each difficulty down (X->H, H->M, M->E). These reductions in speed are similar to those in the Rock Band series.

### DIFF
--- a/Assets/Script/Gameplay/Player/BasePlayer.cs
+++ b/Assets/Script/Gameplay/Player/BasePlayer.cs
@@ -8,6 +8,7 @@ using YARG.Core.Engine;
 using YARG.Core.Input;
 using YARG.Core.Logging;
 using YARG.Gameplay.HUD;
+using YARG.Helpers.Extensions;
 using YARG.Input;
 using YARG.Player;
 using YARG.Settings;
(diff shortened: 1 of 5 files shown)


Assets/Script/Gameplay/Player/BasePlayer.cs
```diff
@@ -27,15 +28,15 @@ public float NoteSpeed
                 // which is saved in the engine parameter override.
                 if (GameManager.IsReplay)
                 {
-                    return Player.Profile.NoteSpeed / (float) Player.EngineParameterOverride.SongSpeed;
+                    return Player.Profile.NoteSpeed * (SettingsManager.Settings.ReduceNoteSpeedByDifficulty.Value ? Player.Profile.CurrentDifficulty.NoteSpeedScale() : 1.0f) / (float) Player.EngineParameterOverride.SongSpeed;
```
grishhung, line 31: NItpicking here, but these lines are really long; you could probably:

1. Consolidate these into a variable shared between all three conditional branches (you're going to have to make the calculation at least once anyway, right)
2. Split the ternary operator into three lines to make it more readable, like so:

```
float highwaySpeedMultiplier = SettingsManager.Settings.ReduceNoteSpeedByDifficulty.Value
    ? Player.Profile.CurrentDifficulty.NoteSpeedScale()
    : 1f
```

sonicfind: I would even take it a step further: put the ternary in the `NoteSpeedScale()` function, having it return `1` if the setting is `false`.

TheNathannator:
I'd suggest moving this to the `engine-refactor` branch instead of `dev`, large features or ones that touch on gameplay code should be targeted towards v0.13 so that we don't have to keep going through merge conflicts when bringing `engine-refactor` up-to-date lol

PR #845, opened 2024-08-10
Added note speed reduction by difficulty

Hubbubble1:
This PR adds a note speed reduction for all players/bots playing charts at a lower difficulty. This can be toggled in the Settings menu using the "Reduce Note Speed by Difficulty" option.

[[video]](https://www.youtube.com/watch?v=ucWM8VI1ZhE)

Notes:
- Reduce Note Speed by Difficulty is enabled by default.
- The vocal track is not affected by this option: it remains the same speed on all difficulties.
- The specified note speed is applied on Expert and Expert+, and the note speeds are reduced by 25% for each difficulty down (X->H, H->M, M->E). These reductions in speed are similar to those in the Rock Band series.

### DIFF
--- a/Assets/Script/Gameplay/Player/BasePlayer.cs
+++ b/Assets/Script/Gameplay/Player/BasePlayer.cs
@@ -8,6 +8,7 @@ using YARG.Core.Engine;
 using YARG.Core.Input;
 using YARG.Core.Logging;
 using YARG.Gameplay.HUD;
+using YARG.Helpers.Extensions;
 using YARG.Input;
 using YARG.Player;
 using YARG.Settings;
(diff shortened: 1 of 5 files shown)


Assets/Script/Gameplay/Player/BasePlayer.cs
```diff
@@ -29,15 +30,15 @@ public float NoteSpeed
                 // which is saved in the engine parameter override.
                 if (GameManager.IsReplay)
                 {
-                    return Player.Profile.NoteSpeed / (float) Player.EngineParameterOverride.SongSpeed;
+                    return Player.Profile.NoteSpeed * Player.Profile.CurrentDifficulty.NoteSpeedScale() / (float) Player.EngineParameterOverride.SongSpeed;
```
EliteAsian123, line 33: `Player.Profile.NoteSpeed * Player.Profile.CurrentDifficulty.NoteSpeedScale()` is duplicated everywhere so it's probably best that this is put into a local variable called `noteSpeed` or something like that.

EliteAsian123 (changes requested):
One quick change and this should be good to go!

Assets/Script/Gameplay/Player/BasePlayer.cs
Hubbubble1: Done; let me know if the change I made is in the right style

EliteAsian123 (approved):

PR #846, opened 2024-08-10
Whammy FX

jbienz:
This PR replaces #812 because changes were getting difficult to follow due to merge rather than rebase.

This PR depends on [YARG.Core PR 192](https://github.com/YARC-Official/YARG.Core/pull/192).

Whammy

I have done my best to incorporate the changes requested by @sonicfind. This includes making sure the effect is also applied to the reverb channel. 

Having said that, I honestly still don't fully follow the logic regarding user sharing of stems. I would like to request either:

1. We merge this feature as-is since its experimental anyway and we can add better handling for instrument sharing later.
2. @sonicfind, @TheNathannator and I have a voice chat on Discord to help me better understand the request and implementation.

### DIFF
--- a/Assets/Script/Audio/Bass/BassStemChannel.cs
+++ b/Assets/Script/Audio/Bass/BassStemChannel.cs
@@ -35,23 +35,45 @@ namespace YARG.Audio.BASS
 
         protected override void SetWhammyPitch_Internal(float percent)
         {
-            if (_streamHandles.PitchFX == 0 || _reverbHandles.PitchFX == 0)
-                return;
-
             percent = Mathf.Clamp(percent, 0f, 1f);
 
-            float shift = Mathf.Pow(2, -(GlobalAudioHandler.WhammyPitchShiftAmount * percent) / 12);
-            _pitchParams.fPitchShift = shift;
-
-            if (!BassHelpers.FXSetParameters(_streamHandles.PitchFX, _pitchParams))
+            if (percent > 0f)
             {
-                YargLogger.LogFormatError("Failed to set params (normal fx): {0}", Bass.LastError);
-            }
+                // If pitch effect hasn't been added yet, add it.
+                if (_streamHandles.PitchFX == 0)
+                {
+                    _streamHandles.PitchFX = BassHelpers.AddPitchShiftToChannel(_streamHandles.Stream, _pitchParams);
+                }
 
-            if (!BassHelpers.FXSetParameters(_reverbHandles.PitchFX, _pitchParams))
-            {
-                YargLogger.LogFormatError("Failed to set params (reverb fx): {0}", Bass.LastError);
+                // If we have pitch effect, pitch
+                if (_streamHandles.PitchFX != 0)
+                {
+                    float shift = Mathf.Pow(2, -(GlobalAudioHandler.WhammyPitchShiftAmount * percent) / 12);
+                    _pitchParams.fPitchShift = shift;
+
+                    if (!BassHelpers.FXSetParameters(_streamHandles.PitchFX, _pitchParams))
+                    {
+                        YargLogger.LogFormatError("Failed to set pitch params: {0}", Bass.LastError);
+                    }
+                }
             }
+            /*
+            else
+            {
+                // If pitch is effect running we could remove it.
+                // This would help with delay but there's a skip when adding or removing.
+                // Probably better to do this after at zero whammy rest for a period of time.
+
+                if (_streamHandles.PitchFX != 0)
+                {
+                    if (!Bass.ChannelRemoveFX(_streamHandles.Stream, _streamHandles.PitchFX))
+                    {
+                        YargLogger.LogFormatError("Failed to remove pitch effect: {0}!", Bass.LastError);
+                    }
+                    _streamHandles.PitchFX = 0;
+                }
+            }
+            */
         }
 
         protected override void SetPosition_Internal(double position)
--- a/Assets/Script/Audio/Bass/BassStemChannel.cs
+++ b/Assets/Script/Audio/Bass/BassStemChannel.cs
@@ -45,17 +45,20 @@ namespace YARG.Audio.BASS
                 _reverbHandles.PitchFX = BassHelpers.AddPitchShiftToChannel(_reverbHandles.Stream, _pitchParams);
             }
 
+            // Calculate shift
+            float shift = Mathf.Pow(2, -(GlobalAudioHandler.WhammyPitchShiftAmount * percent) / 12);
+            _pitchParams.fPitchShift = shift;
+
             // If we have pitch effect, pitch
             if (_streamHandles.PitchFX != 0)
             {
-                float shift = Mathf.Pow(2, -(GlobalAudioHandler.WhammyPitchShiftAmount * percent) / 12);
-                _pitchParams.fPitchShift = shift;
-
                 if (!BassHelpers.FXSetParameters(_streamHandles.PitchFX, _pitchParams))
                 {
                     YargLogger.LogFormatError("Failed to set pitch on stream: {0}", Bass.LastError);
                 }
-
+            }
+            if (_reverbHandles.PitchFX != 0)
+            {
                 if (!BassHelpers.FXSetParameters(_reverbHandles.PitchFX, _pitchParams))
                 {
                     YargLogger.LogFormatError("Failed to set pitch on reverb: {0}", Bass.LastError);
--- a/Assets/Script/Audio/Bass/BassStemChannel.cs
+++ b/Assets/Script/Audio/Bass/BassStemChannel.cs
@@ -35,26 +35,30 @@ namespace YARG.Audio.BASS
 
         protected override void SetWhammyPitch_Internal(float percent)
         {
-            percent = Mathf.Clamp(percent, 0f, 1f);
-
-            if (percent > 0f)
+            // If pitch effect hasn't been added yet, add it.
+            if (_streamHandles.PitchFX == 0)
             {
-                // If pitch effect hasn't been added yet, add it.
-                if (_streamHandles.PitchFX == 0)
+                _streamHandles.PitchFX = BassHelpers.AddPitchShiftToChannel(_streamHandles.Stream, _pitchParams);
+            }
+            if (_reverbHandles.PitchFX == 0)
+            {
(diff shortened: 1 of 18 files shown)


Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -160,9 +161,11 @@ public void OpenVenueFolder()
             public ToggleSetting ClapsInStarpower { get; } = new(true);
 
             public ToggleSetting OverstrumAndOverhitSoundEffects { get; } = new(true);
+            
+            public ToggleSetting UseWhammyFx { get; } = new(false, v => GlobalAudioHandler.UseWhammyFx = v);
+
+            public SliderSetting WhammyPitchShiftAmount { get; } = new(1, 1, 5, v => GlobalAudioHandler.WhammyPitchShiftAmount = v);
```
TheNathannator, line 168: For parity with the other settings, this should use an explicit function for the callback instead of a lambda. Also would prefer for there to be no spacing between these properties, as they are related settings.

Assets/Script/Audio/Bass/BassStemChannel.cs
```diff
@@ -35,23 +35,57 @@ internal BassStemChannel(AudioManager manager, SongStem stem, bool clampStemVolu
...
-            percent = Mathf.Clamp(percent, 0f, 1f);
+            // If pitch effect hasn't been added yet, add it.
+            if (_streamHandles.PitchFX == 0)
+            {
+                _streamHandles.PitchFX = BassHelpers.AddPitchShiftToChannel(_streamHandles.Stream, _pitchParams);
+            }
+            if (_reverbHandles.PitchFX == 0)
+            {
+                _reverbHandles.PitchFX = BassHelpers.AddPitchShiftToChannel(_reverbHandles.Stream, _pitchParams);
+            }
```
TheNathannator, line 46: Adding the pitch FX should probably be done at channel initialization instead of on-demand, though I do see the commented code below about removing the FX when the pitch shift stops.

TheNathannator (commented):
I'd say this should target `engine-refactor` instead if you're willing to rebase again, otherwise this will do for now. We want to avoid further merge conflict issues with features being added on `dev`.

Mostly just code style nitpicks in my review, we can revisit the finer details at another time. Will wait for sonicfind to look over this as well.

Assets/Script/Settings/SettingsManager.Settings.cs
sonicfind: Well that's awkward, because I said the opposite lmao

sonicfind: There's argument for either really, especially if we ever consider using function pointers instead of `Action`s for the settings.

TheNathannator: I just want to keep it consistent lol, everything else already uses an explicit function.

> especially if we ever consider using function pointers 

Why *would* we consider that here though? Very little benefit and requires modifying every single callback argument to be taken by address. We don't need to hyper-optimize every aspect of the game lol

Assets/StreamingAssets/lang/en-US.json
```diff
@@ -745,6 +749,10 @@
                 "Name": "Enable VSync",
                 "Description": "Enables VSync. VSync caps the framerate at your monitor's maximum framerate, and prevents tearing."
             },
+            "WhammyPitchShiftAmount": {
+                "Name": "Whammy Pitch Shift Amount",
+                "Description": "Number of semitones the pitch is changed when the whammy bar is fully depressed. Most guitars shift between 1 and 3 semitones."
```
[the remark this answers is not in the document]
jbienz: I respectfully disagree. This setting originally allowed a range of 1 - 12. I didn't even understand what that range meant, and I had to look up what a semitone was. It appears that most guitars range between 1 and 3 and only very specialized guitars go up to 5. Without a music background, and without looking this up, I would have never understood what I should set this value to. Can you help me better understand what the concern is about providing this kind of information to our users who are not familiar with these terms? Or perhaps help me better understand the benefit of omitting this information?

Assets/Script/Gameplay/GameManager.Audio.cs
```diff
@@ -48,6 +49,14 @@ public bool SetReverb(bool reverb)
                 return ReverbCount > 0;
             }
 
+            public float SetWhammyPitch(float percent)
+            {
+                // TODO: Would be nice to handle multiple inputs
+                // but for now last one wins
+                WhammyPitch = (percent < 0 || percent > 1 ? 0 : percent);
```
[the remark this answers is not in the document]
jbienz: Agreed. I will manually make this change.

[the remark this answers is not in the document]
jbienz: P.S. how should I choose which branch I should be rebasing from? The guidance I kept seeing was to rebase from Dev. Is there anything that would have indicated I should have rebased from engine-refactor instead?

Assets/Script/Settings/SettingsManager.Settings.cs
jbienz: I am personally a fan of using lambdas for simple one-line calls and using functions when an expansion is needed. I personally feel embedded multi-line lambdas start to look cluttered. The other place I believe functions make sense is when the function needs to be virtual so it can potentially be overridden in an inherited class. That is obviously not the case here though. I'm fine with either way, I just think we should be consistent.

I'm OK with grouping related settings together if that's a team preference. Generally I prefer to sort things alphabetically so it's easy to determine where to insert something new and where I to find something old. The main way to find a setting when things are grouped together is by knowing the name and searching for it (or using the member dropdown). But I do totally understand the desire for grouping related settings and I'm happy to follow that convention if that's the consensus.

Assets/Script/Audio/Bass/BassStemChannel.cs
jbienz: If we do it during channel initialization, what method would that be done in? I noticed that ReverbFX isn't initialized until it's needed, so I followed that same pattern.

Also, there was one time where I thought I heard audio desync on my machine and when I disabled the effect it seemed the track "caught back up". So I was thinking that after a period of time with Whammy being 0, we might want to disable the effect. That could have been a false perception though.

Assets/StreamingAssets/lang/en-US.json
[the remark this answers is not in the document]
TheNathannator: Ah, I didn't realize it was for reference. GH and RB only use one semitone for their whammy pitch bend iirc, YARG allowing you to go beyond that is just a "for kicks and giggles" thing.

My intent for leaving this out was for conciseness and because I found it confusing, it can be difficult to try and pack all prerequisite information into the description. I think I'd suggest something like this:
`Number of semitones the pitch is changed when the whammy bar is fully pressed. Larger amounts result in lower pitches; default is 1.`

Assets/Script/Settings/SettingsManager.Settings.cs
TheNathannator: I don't think we ever established a spoken convention for this in particular, I've just done my usual. In this case the lambda is small enough to not really bother me, it's just not consistent with the rest of the settings. I'm not hard-set on one or the other.

Assets/Script/Gameplay/GameManager.Audio.cs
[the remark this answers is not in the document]
TheNathannator: The indication was our announcements in the Discord, but the readme hasn't been updated to reflect that. I don't think we'll have to have to deal with it for too much longer, but it would probably be a good change to make in the interim.

Assets/Script/Audio/Bass/BassStemChannel.cs
TheNathannator: It's probably fine to leave it as on-demand initialization so long as the audio isn't negatively affected (i.e. it doesn't hitch for a moment). I was just noting this code in the removed lines and putting some significance on it:
```cs
if (_streamHandles.PitchFX == 0 || _reverbHandles.PitchFX == 0)
    return;
```

Assets/StreamingAssets/lang/en-US.json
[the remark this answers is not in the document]
jbienz: Okay. I'll change it to be what you said.

Assets/Script/Settings/SettingsManager.Settings.cs
jbienz: Sounds good. Thank you for the flexibility @TheNathannator.

Assets/Script/Audio/Bass/BassStemChannel.cs
jbienz: In my opinion, our biggest issue with the Whammy FX feature comes down to issues with BASS and pitch. In my testing, it appears that just having the pitch effect enabled (even if the pitch is the default pitch) can lead to audio desync. If that wasn't a factor, I'd recommend we create the effect at the start of a song and keep it on the whole time. But since it can cause desync, I wanted to delay creating it until it's used. I'm even debating removing it after it hasn't been used for a bit to try and help the stem "catch up".

It's been my experience that sometimes there is a tiny hitch when the pitch effect is first added. It's very brief. Perhaps only a few milliseconds. And it doesn't always happen. But this is why I chose not to disable the effect as soon as the whammy bar returns to zero. I think we could disable it after a second or two, but I don't want to enable and disable it quickly in succession as it increases the likelihood of a hitch.

This really comes down to the least amount of sacrifices we have to make with the current state of BASS. I don't like the tradeoffs either, but I think this is part of why we're calling it an "Experimental" feature. It doesn't work flawlessly, but it's the best we can do for now.

I would like to investigate lazy disabling it after a second or two of sitting at Whammy 0, but I don't know how I'd do that. This event only gets called when the bar changes. I'd need to do a callback. Maybe schedule a coroutine or something. I'll look into that, but lazy initialization may be the best we can do for now.

jbienz: After doing more research, I would like to leave this as-is for now. Since this code is in the DLL and not the Unity code, I can't use a coroutine. It also isn't async code so I can't easily do a delay. I would like to start with this and we can follow up with tweaks as users report their experience.

jbienz:
@TheNathannator I believe all of your requests have been addressed at this point. Please let me know if this can be merged, and thank you again for taking the time to review!

TheNathannator:
Merged via #855.

PR #849, opened 2024-08-13
Datastream implementation (for real this time)

TheFatBastid:
Created a UDP data stream that YARG will output on port 36107. This data stream contains information on virtually everything YARG is doing at the moment. This replaces all built-in IRL lighting protocols: Stage Kit, DMX, RB3E, HUE, etc. These are now all handled by the external program, YALCY. (all related settings removed as well)


Assets/Script/Integration/MasterLightingController.cs
```diff
@@ -1,283 +1,602 @@
...
+            _timer.Elapsed += (sender, e) => Sender();
+            _timer.Start();
+        }
+
+        private void Sender()
+        {
+            if (!SettingsManager.Settings.EnableYALCYDatastream.Value) return;
+            try
+            {
+                _sendClient.Send(_dataPacket, _dataPacket.Length, IPAddress.ToString(), PORT);
```
RileyTheFox, line 497: Cache this ToString result somewhere. The IP address is readonly so the hostname is not going to change so this should not need to be called every time.

I couldn't see if this was allocating anything but its best to be safe and just cache the result anyway.

RileyTheFox (changes requested):
I tested this out and it worked perfectly after I built YALCY (I don't have any lighting hardware so I was just going off of the visualiser in the app)

Profiled it as I am concerned about performance with a feature like this and did not notice any memory allocations during gameplay or high CPU usage which is good.

I am however very concerned about the way the packet construction is done, as well as the fact that *everything* is using an enum. The packet structure should not be determined by values in an enum. This is not a maintainable solution if it were to be added to in the future. It should be a *structure* (struct or class) with fields for the data that will be in the packet which is then serialized to binary.

Something along the lines of this:
```csharp
public struct LightingMessage
{
    public float BeatsPerMinute;
    public LightingEvent LightingCue;
    etc...
}
```

this Message type is then serialized to binary and sent off over the network.

I haven't put too many review comments because I'd just be repeating the same thing. My comments are highlighted usually at the first instance I came across something, but what its saying can be applied to many other places too.

I want this to be done right because if done correctly, this data stream setup can be reused for sending packets for lots of other game information beyond lighting. So if its done correctly now there's less work to do for those features down the line.

TheFatBastid:
Ready for round two review. Don't merge as I still need to redo the YALCY side to match this.

Assets/Script/Integration/MasterLightingGameplayMonitor.cs
```diff
@@ -67,196 +85,208 @@ protected override void OnChartLoaded(SongChart chart)
...
+            if (listIndex < list.Count && list[listIndex].StartTime <= GameManager.SongTime)
             {
-                //so instant notes should at least be on for 1 frame because of the else statement
-                if (_guitar.Notes[_guitarEndCheckIndex].TimeEnd <= GameManager.SongTime)
-                {
-                    MasterLightingController.CurrentGuitarNotes = 0;
-                    //guitarEndCheckIndex is set to -1 when the note ends
-                    _guitarEndCheckIndex = -1;
-                }
+                return (int) list[listIndex].Pitch;
```
RileyTheFox, line 176: Pitch is a `float`, is the cast to int necessary? Data is being lost with this cast

Assets/Script/Integration/MasterLightingGameplayMonitor.cs
```diff
@@ -67,196 +85,208 @@ protected override void OnChartLoaded(SongChart chart)
...
-            // End of vocal phrase. SilhouetteSpot is the only cue that uses vocals, listening to the end of the phrase.
-            while (_vocalsIndex < _vocals.Count &&
-                Math.Min(_vocals[_vocalsIndex].PhraseParentNote.ChildNotes[^1].TotalTimeEnd,
-                    _vocals[_vocalsIndex].TimeEnd) <= GameManager.SongTime)
+            var Harmony2Note = VocalEventChecker(_harmony2Notes, ref _harmony2Index);
+            if (Harmony2Note != -1)
             {
-                MasterLightingController.CurrentVocalNote = _vocals[_vocalsIndex].PhraseParentNote.ChildNotes[^1];
-                _vocalsIndex++;
+                MasterLightingController.MLCCurrentHarmony2Note = (byte)Harmony2Note;
```
RileyTheFox, line 222: The `Pitch` value from VocalEventChecker has now gone from a 32 bit float, to a 32 bit integer to an 8 bit integer. What is with casting it down so much?

Assets/Script/Integration/MasterLightingGameplayMonitor.cs
```diff
@@ -67,196 +85,208 @@ protected override void OnChartLoaded(SongChart chart)
...
-                {
-                    MasterLightingController.CurrentGuitarNotes = 0;
-                    //guitarEndCheckIndex is set to -1 when the note ends
-                    _guitarEndCheckIndex = -1;
-                }
+                return (int) list[listIndex].Pitch;
             }
 
-            //----
+            return -1;
```
RileyTheFox, line 179: Because this value gets casted to a byte in the end, this will end up as `255` because a byte is not signed. Is that intentional?

Assets/Script/Integration/MasterLightingController.cs
```diff
@@ -1,283 +1,305 @@
...
-        public static int CurrentDrumNotes
-        {
-            get => _currentDrumNote;
-            set
-            {
-                PreviousDrumNote = _currentDrumNote;
-                _currentDrumNote = value;
-                OnInstrumentEvent?.Invoke(InstrumentType.Drums, value);
+                //Keyframes are indicators and not really lighting cues themselves, also chorus and verse act more as modifiers and section labels and also not really lighting cues, they can be stacked under a lighting cue.
+                if (value.Type != LightingType.Keyframe_Next && value.Type != LightingType.Keyframe_Previous &&
```
RileyTheFox, line 122: Consider using pattern matching here to make it easier to read.

```csharp
if (value.Type is not (LightingType.One or LightingType.Next or etc etc))
```

Assets/Script/Integration/MasterLightingController.cs
```diff
@@ -1,283 +1,305 @@
...
 
-        public static int PreviousBassNote;
-
-        public static PerformerEvent CurrentPerformerEvent
+        private static byte SetPlatformByte()
         {
-            get => _currentPerformerEvent;
-            set
+            var platform = Application.platform;
+            switch (platform)
```
RileyTheFox, line 194: Running this each time is unnecessary as the platform is never going to change during runtime. Use the `#if #elif #else #endif` preprocessors and check for unity symbol definitions for each platform

[Preprocessor Directives](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives)

[Unity Platform Scripting Symbols](https://docs.unity3d.com/Manual/PlatformDependentCompilation.html)

You only need to use the editor/standalone for Windows, OSX and Linux.

Look at this [example](https://github.com/YARC-Official/YARG/blob/86bba9fee15c92d80b2afa96fd93091dc85c5825/Assets/Script/Audio/Bass/BassAudioManager.cs#L345) from YARG to see how they are used.

RileyTheFox (changes requested):
Big improvements over the first iteration but still some way to go. You've introduced allocation every time a packet is created and sent (which is 88 times a second!) which is bad for gameplay as it increases the likelihood the Garbage Collector stalls the game causing lag spikes.

You also used BinaryFormatter, which is a huge no no. See the comment on that for details.

Assets/Script/Integration/MasterLightingGameplayMonitor.cs
TheFatBastid: Old dmx code and thought process, as it was to fit it in into 1 dmx channel byte, and was 'close enough' but that isn't applicable anymore, will fix it.

TheFatBastid:
Ready for round three!

Assets/Script/Integration/MasterLightingController.cs
```diff
@@ -267,14 +238,30 @@ private void OnApplicationQuit()
...
 
-            var message = new LightingMessage
-            {
-                HeaderByte1 = 0x59, // Y
-                HeaderByte2 = 0x41, // A
-                HeaderByte3 = 0x52, // R
-                HeaderByte4 = 0x47, // G
-                //everything else is 0
-            };
+            message.Header = 0x59415247; // Y A R G
```
RileyTheFox, line 241: If you're sending a blank packet to turn everything off, just recreate the LightingMessage object. Saves you having to retype every variable within the message type just to set it to 0.

Assets/Script/Integration/MasterLightingController.cs
```diff
@@ -81,32 +71,37 @@ private enum SceneIndexByte
...
+        public static LightingType MLCKeyframe;
+        public static LightingType MLCCurrentLightingCue;
+        public static PostProcessingType MLCPostProcessing;
 
         public static ushort MLCudpPort;
         public static string MLCudpIP;
 
         private static LightingEvent _currentLightingCue;
 
+        public static LightingMessage message = new LightingMessage();
```
RileyTheFox, line 101: These should be private, and also do not need to be static from what I can tell. Rename appropriately, `_privateVar`

RileyTheFox (changes requested):

TheFatBastid:
Ready for round four!

PR #850, opened 2024-08-13
Voice Activated Star Power setting for Vocalists

thomeval:
By default, vocalists are able to activate Star Power by making noise when not singing. This PR adds adds a new setting, which allows them to disable this behaviour if desired. Regardless of this setting, they can still activate SP by using a keybinding.

image

Note to the reviewer: Internally, this feature is named "SingToActivateStarPower". Since vocalists don't technically need to sing to activate Star Power - any noise will do, including shouting, dropping the mic, or heckling the drummer - I've named this setting "Voice Activated Star Power (Vocals)" instead. Let me know if you'd like it to be named something else.

Requires https://github.com/YARC-Official/YARG.Core/pull/203 from YARG.Core.

### DIFF
--- a/Assets/Script/Settings/SettingsManager.Settings.cs
+++ b/Assets/Script/Settings/SettingsManager.Settings.cs
@@ -1,4 +1,4 @@
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Net;
 using UnityEngine;
 using UnityEngine.InputSystem;
@@ -73,6 +73,8 @@ namespace YARG.Settings
 
             public ToggleSetting UseCymbalModelsInFiveLane { get; } = new(true);
             public SliderSetting KickBounceMultiplier { get; } = new(1f, 0f, 2f);
+
+            public ToggleSetting VoiceActivatedVocalStarPower {get; } = new(true);
 
             public SliderSetting ShowCursorTimer { get; } = new(2f, 0f, 5f);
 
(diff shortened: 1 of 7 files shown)


Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -76,6 +76,8 @@ public void OpenVenueFolder()
             public ToggleSetting ReduceNoteSpeedByDifficulty { get; } = new(true);
             public SliderSetting KickBounceMultiplier { get; } = new(1f, 0f, 2f);
 
+            public ToggleSetting VoiceActivatedVocalStarPower {get; } = new(true);
```
EliteAsian123, line 79: Missing space before `get`

Assets/StreamingAssets/lang/en-US.json
```diff
@@ -767,6 +767,10 @@
                 "PauseName": "Vocals",
                 "Description": "Adjusts the song's vocal track volume. Only does something if the song is multi-track."
             },
+            "VoiceActivatedVocalStarPower" : {
+				"Name": "Voice Activated Star Power (Vocals)",
+				"Description": "If enabled, vocalists can activate their Star Power by making noise when not singing (or by singing badly). Keybindings can be used instead, regardless of this setting."
```
EliteAsian123, line 772: Should probably be:
```
If enabled, vocalists can activate their Star Power by making a noise. Keybinds can be used instead, regardless of this setting.
```

EliteAsian123 (changes requested):
A few simple changes, and we're good here too!

As a reminder, once the YARG.Core PR is merged you'll have to update the YARG.Core pointer on this PR as well.

Thanks!

Assets/StreamingAssets/lang/en-US.json
thomeval: Description has been changed.

PR #851, opened 2024-08-14
Fix vocal particle colours being blue for HARM2/HARM3 notes

thomeval:
This PR is to fix the vocal particle colours always being rendered blue, regardless of whether they are solo, HARM1 (blue), HARM2 (red), or HARM3 (yellow). 

vocalParticleColours

Note to the reviewer: This implementation matches the implementation for colouring the vocal notes themselves, which was clearly marked as a temporary solution (see VocalTrack.cs). As such, the hard-coded colour set should be replaced with colour profiles for vocals later on.

### DIFF
--- a/Assets/Script/Gameplay/Player/VocalsPlayer.cs
+++ b/Assets/Script/Gameplay/Player/VocalsPlayer.cs
@@ -18,7 +18,7 @@ namespace YARG.Gameplay.Player
     public class VocalsPlayer : BasePlayer
     {
         public VocalsEngineParameters EngineParams { get; private set; }
-        public VocalsEngine Engine { get; private set; }
+        public VocalsEngine           Engine       { get; private set; }
 
         public override BaseEngine BaseEngine => Engine;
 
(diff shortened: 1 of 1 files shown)


Assets/Script/Gameplay/Player/VocalsPlayer.cs
```diff
@@ -60,6 +60,13 @@ public class VocalsPlayer : BasePlayer
 
             base.Initialize(index, player, chart, lastHighScore);
 
+            hud.Initialize(player.EnginePreset);
+            _hud = hud;
```
EliteAsian123, line 64: Was this change intentional? It looks like you just duplicated the hud initialization code from below.

EliteAsian123 (changes requested):

Assets/Script/Gameplay/Player/VocalsPlayer.cs
thomeval: Sorry about that, looks like I messed up when merging from the target branch.

PR #853, opened 2024-08-17
Display unique needle graphics for each Vocals player (up to 7)

thomeval:
This PR adds unique vocal needle graphics for each vocals player, based on the order their profiles joined/connected. This uses the existing needle sprites, although in a different order to maximize visibility.

I've chosen the sprites shown below for the first three vocalists, as it provides at least some visibility to each player even in the worst case scenario where all three are on top of each other. 
YARG-3pVocals

And this is what it looks like with six vocalists, two per harmony part. If there are more than seven vocalists, each needle graphic will be reused, in order.

YARG-6pVocals


Assets/Script/Gameplay/Player/VocalsPlayer.cs
```diff
@@ -53,13 +59,21 @@ public class VocalsPlayer : BasePlayer
...
         {
             if (IsInitialized) return;
 
             base.Initialize(index, player, chart, lastHighScore);
+            VocalPlayerIndex = vocalIndex;
 
+            // Needle materials have names starting from 1.
+            var needleIndex = (VocalPlayerIndex % NEEDLES_COUNT) + 1;
+            var materialPath = $"Assets/Art/Materials/Gameplay/Visual/VocalElements/VocalNeedle{needleIndex}.mat";
+            _needleRenderer.material = Addressables.LoadAssetAsync<Material>(materialPath).WaitForCompletion();
```
EliteAsian123, line 75: The vocal  needle addressable group name can be simplified a bit (to something like `VocalNeedle/N`), just so the naming fits in with the other addressables

Assets/Script/Gameplay/Player/VocalsPlayer.cs
```diff
@@ -53,13 +59,21 @@ public class VocalsPlayer : BasePlayer
...
-        public void Initialize(int index, YargPlayer player, SongChart chart,
+        private const int NEEDLES_COUNT = 7;
+
+        public void Initialize(int index, int vocalIndex, YargPlayer player, SongChart chart,
             VocalsPlayerHUD hud, VocalPercussionTrack percussionTrack, int? lastHighScore)
         {
             if (IsInitialized) return;
 
             base.Initialize(index, player, chart, lastHighScore);
+            VocalPlayerIndex = vocalIndex;
```
EliteAsian123, line 70: Is this property needed? It looks like it's only used to determine the sprite at the moment.

Assets/Script/Gameplay/Player/VocalsPlayer.cs
```diff
@@ -53,13 +59,21 @@ public class VocalsPlayer : BasePlayer
...
 
+            // Needle materials have names starting from 1.
+            var needleIndex = (VocalPlayerIndex % NEEDLES_COUNT) + 1;
```
EliteAsian123, line 73: The parentheses are not needed here

EliteAsian123 (changes requested):
Hello! Thanks for the PR and sorry for the late review.

Just a few very minor things and this should be good to merge! Thank you!

Assets/Script/Gameplay/Player/VocalsPlayer.cs
thomeval: Would you mind if we kept these parentheses? We've all memorized the order of operations by this point, but I feel like this line is easier to read with them. I'd understand if we'd rather want to be consistent throughout though.

thomeval:
Okay, this is ready for another review.

PR #858, opened 2024-08-22
Sorted Profiles List

thomeval:
This PR adds sorting and groups to the Profiles list. All profiles are now grouped into three groups:
- Active profiles. These are profiles that are currently connected, both players and bots.
- Players. These are human profiles that aren't currently being used.
- Bots. These are bot profiles that aren't currently being used.

All entries in the above groups are sorted alphabetically as well.

image

Note to the reviewer: For the group headings, I borrowed the Prefab used for headers on the Settings Menu, though I needed to add a Component to them to be able to set their text at runtime.


Assets/Script/Menu/ProfileList/ProfileListMenu.cs
```diff
@@ -49,12 +53,26 @@ private void RefreshList()
...
+            AddListGroup("Active Profiles", activeProfiles);
+            AddListGroup("Players", otherProfiles.Where(e => !e.IsBot));
+            AddListGroup("Bots", otherProfiles.Where(e => e.IsBot));
+
+        }
+
+        private void AddListGroup(string header, IEnumerable<YargProfile> profiles)
+        {
+            var headerGo = Instantiate(_profileListHeaderPrefab, _profileList);
+            headerGo.GetComponent<ListHeader>().Text = header;
```
EliteAsian123, line 68: Since you're doing `GetComponent` here anyways, it may be easier to just do `GetComponentInChildren<TextMeshProUGUI>().text = ...` instead, and remove the new script.

Assets/Script/Menu/ProfileList/ProfileListMenu.cs
```diff
@@ -49,12 +53,26 @@ private void RefreshList()
             _profileList.transform.DestroyChildren();
             _navigationGroup.ClearNavigatables();
 
+            // Is there a better way of doing this?
+            var activeProfiles = PlayerContainer.Players.Select(e => e.Profile).OrderBy(e => e.Name).ToArray();
+            var otherProfiles = PlayerContainer.Profiles.Except(activeProfiles).OrderBy(e => e.Name).ToArray();
+            AddListGroup("Active Profiles", activeProfiles);
+            AddListGroup("Players", otherProfiles.Where(e => !e.IsBot));
+            AddListGroup("Bots", otherProfiles.Where(e => e.IsBot));
```
EliteAsian123, line 61: The more efficient way would be to loop through all of the profiles using just a normal foreach loop. Will that be more readable? Probably about the same. It's up to you if you wanna do that instead.

EliteAsian123 (changes requested):
For some reason I'm having trouble loading into the game with this PR. Did you accidently leave the profile list enabled in the menu scene? Because of this I can't really test it, but I'll will comment on what I noticed below.

Thanks for your patience!
image

Assets/Script/Menu/ProfileList/ProfileListMenu.cs
thomeval: I think that would be tricky, since there are two collections at play here (`PlayerContainer.Profiles` and `PlayerContainer.Players` ). You'd also have to implement the sorting manually if you really wanted to avoid using LINQ. It would be more efficient, but by how much? I haven't run it through a profiler, but I'm guessing we're probably talking less than a 1ms difference (assuming you have less than a thousand profiles), and be far less relevant that the `Instantiate()` calls that are used in `AddListGroup()`.

thomeval:
@EliteAsian123 Sorry about that. I've disabled the profile list, so its working again. I've also updated the branch to remove the ListHeader script as suggested.

Assets/Script/Menu/ProfileList/ProfileListMenu.cs
```diff
@@ -49,12 +54,24 @@ private void RefreshList()
             _profileList.transform.DestroyChildren();
             _navigationGroup.ClearNavigatables();
 
+            var activeProfiles = PlayerContainer.Players.Select(e => e.Profile).OrderBy(e => e.Name).ToArray();
+            var otherProfiles = PlayerContainer.Profiles.Except(activeProfiles).OrderBy(e => e.Name).ToArray();
+            AddListGroup("Active Profiles", activeProfiles);
+            AddListGroup("Players", otherProfiles.Where(e => !e.IsBot));
+            AddListGroup("Bots", otherProfiles.Where(e => e.IsBot));
```
EliteAsian123, line 61: The header text should be localized (with `Localize.Key`)

EliteAsian123 (commented):
Thanks for doing that!

One thing I think would be good is to disable the header when there are no items in, as I think having two headers in a row is kinda weird:
image

Other than that (and the one review comment below) this is pretty much good to go!

Sorry for the wait!

thomeval:
Okay, I've pushed changes to the branch that resolves those two issues.

EliteAsian123 (approved):

PR #861, opened 2024-08-30
Allow drum freestyle during countdown breaks

raphaelgoulart:
Overhitting during those breaks is already allowed in YARG.Core, so it's a matter of adding that check to `IsDrumFreestyle()` in the YARG side


EliteAsian123 (changes requested):
Sorry for the late review,

if you could just indent the newly added condition that would be perfect!

Thanks!

raphaelgoulart:
Done! I meant to do that but somehow forgot/didn't notice it wasn't indented correctly

PR #862, opened 2024-08-30
Datastream, 3rd time is the charm

TheFatBastid:
Contains all changes from previous PR


TheFatBastid:
Sounds good. I'll get on it soon.

PR #867, opened 2024-09-04
Add Active Players and Bots displays to Status Bar v2

thomeval:
This PR adds two displays to the top Status Bar to display the currently active players, as well as the number of active bot profiles. These can be toggled in the Settings menu using the "Show Active Players" and "Show Active Bots" options.

Notes:

- Replaces PR #841
- Requires YARG.Core PR [#209](https://github.com/YARC-Official/YARG.Core/pull/209)
-  Neither of the above options are enabled by default.
-  Only the number of active bots is shown, since their names are probably not useful.
-  The active bots display is hidden automatically if no bot profiles are active (similar to the Battery status display).
-  When 1 to 3 players are connected, their instrument icons and profile names are shown.
-  When 4 or more players are connected, only their instrument icons are shown.
-  However, when no players are connected (including bots), the active players list will instead show a "No profiles active" message.

### DIFF
--- a/Assets/Script/Menu/Persistent/ActivePlayerList.cs
+++ b/Assets/Script/Menu/Persistent/ActivePlayerList.cs
@@ -1,0 +1,42 @@
+﻿using System.Linq;
+using UnityEngine;
+using YARG.Player;
+
+namespace YARG.Menu.Persistent
+{
+    public class ActivePlayerList : MonoBehaviour
+    {
+        [SerializeField]
+        private GameObject _noPlayersContainer;
+        [SerializeField]
+        private GameObject _playerNamesContainer;
+        [SerializeField]
+        private GameObject _playerNamesPrefab;
+
+        [SerializeField]
+        private int _maxShownPlayerNames = 3;
+
+        public void UpdatePlayerList()
+        {
+            var players  = PlayerContainer.Players;
+            // Only show this message if there are no players, including bots.
+            _noPlayersContainer.SetActive(players.Count == 0);
+
+            players = players.Where(e => !e.Profile.IsBot).ToList();
+            var showPlayerNames = players.Count <= _maxShownPlayerNames;
+
+            foreach (Transform child in _playerNamesContainer.transform)
+            {
+                Destroy(child.gameObject);
+            }
+
+            foreach (var player in players)
+            {
+                var newObj = Instantiate(_playerNamesPrefab, _playerNamesContainer.transform);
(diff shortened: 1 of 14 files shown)


Assets/Script/Menu/Persistent/StatsManager.cs
```diff
@@ -214,5 +229,20 @@ private void UpdateBattery()
                 _ => _batterySpriteCritical
             };
         }
+
+        public void UpdateActivePlayers()
+        {
+            if (!(IsShowing(Stat.ActivePlayers) || IsShowing(Stat.ActiveBots)))
+                return;
```
EliteAsian123, line 236: Make sure to add braces to this

Assets/Script/Player/PlayerContainer.cs
```diff
@@ -249,5 +251,16 @@ public static void Destroy()
...
             }
         }
+
+        public static void EnsureValidInstruments()
+        {
+            foreach (var profile in _profiles)
+            {
+                if (!profile.HasValidInstrument)
+                {
+                    profile.CurrentInstrument =  profile.GameMode.PossibleInstruments()[0];
```
EliteAsian123, line 261: It would be my preference to move this statement to the YARG.Core side as a method because it could be used in other applications that make use of the `YargProfile`.

Assets/Script/Menu/Persistent/ActivePlayerList.cs
```diff
@@ -0,0 +1,42 @@
...
+            // Only show this message if there are no players, including bots.
+            _noPlayersContainer.SetActive(players.Count == 0);
+
+            players = players.Where(e => !e.Profile.IsBot).ToList();
+            var showPlayerNames = players.Count <= _maxShownPlayerNames;
+
+            foreach (Transform child in _playerNamesContainer.transform)
+            {
+                Destroy(child.gameObject);
+            }
```
EliteAsian123, line 31: I believe we have an extension method for doing this called `DestroyChildern()` that can be called on the transform.

EliteAsian123 (changes requested):
Thanks for the PR!

Just a few review points and this should be good to go!

Assets/Script/Menu/Persistent/StatsManager.cs
thomeval: Oh dear, what have I done?!

thomeval:
Okay, this is ready for another review.

PR #889, opened 2024-10-04
Always On Drum SFX

nameonascreen:
Added an "Always-On Drum SFX" toggle.

Not the best implementation but will do as a temporary method.. This is global and not per-profile

Feel free to deny, I'm terrible at coding stuff and already know this is bad lol.

### DIFF
--- a/Assets/Script/Gameplay/Player/DrumsPlayer.cs
+++ b/Assets/Script/Gameplay/Player/DrumsPlayer.cs
@@ -108,8 +108,8 @@ namespace YARG.Gameplay.Player
                     if (SettingsManager.Settings.AlwaysOnDrumSFX.Value)
                     {
                         PlayDrumSoundEffect(action, velocity);
-                    }
-                } return;
+                    } return;
+                }
 
                 // Choose the correct fret
                 int fret;
--- a/Assets/Script/Gameplay/Player/DrumsPlayer.cs
+++ b/Assets/Script/Gameplay/Player/DrumsPlayer.cs
@@ -13,6 +13,7 @@ using YARG.Gameplay.HUD;
 using YARG.Gameplay.Visuals;
 using YARG.Helpers.Extensions;
 using YARG.Player;
+using YARG.Settings;
 
 namespace YARG.Gameplay.Player
 {
@@ -101,7 +102,17 @@ namespace YARG.Gameplay.Player
             engine.OnPadHit += (action, wasNoteHit, velocity) =>
             {
                 // Skip if a note was hit, because we have different logic for that below
-                if (wasNoteHit) return;
+            if (SettingsManager.Settings.AlwaysOnDrumSFX.Value)
+            {
+                if (!wasNoteHit)
+                {
+                    PlayDrumSoundEffect(action, velocity);
+                }
+                else
+                {
+                    PlayDrumSoundEffect(action, velocity);
+                }
+            }
 
                 // Choose the correct fret
                 int fret;
--- a/Assets/Script/Gameplay/Player/DrumsPlayer.cs
+++ b/Assets/Script/Gameplay/Player/DrumsPlayer.cs
@@ -101,17 +101,15 @@ namespace YARG.Gameplay.Player
 
             engine.OnPadHit += (action, wasNoteHit, velocity) =>
             {
-
-            // Skip if a note was hit, because we have different logic for that below
-            if (wasNoteHit)
-            {
-                // If AODSFX is turned on and a note was hit, Play the drum sfx. Without this, drum sfx will only play on misses.
-                if (SettingsManager.Settings.AlwaysOnDrumSFX.Value)
+                // Skip if a note was hit, because we have different logic for that below
+                if (wasNoteHit)
                 {
-                    PlayDrumSoundEffect(action, velocity); return;
+                    // If AODSFX is turned on and a note was hit, Play the drum sfx. Without this, drum sfx will only play on misses.
+                    if (SettingsManager.Settings.AlwaysOnDrumSFX.Value)
+                    {
+                        PlayDrumSoundEffect(action, velocity);
+                    } return;
                 }
-                else return;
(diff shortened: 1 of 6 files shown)


Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -172,6 +172,8 @@ public void OpenVenueFolder()
 
             public ToggleSetting OverstrumAndOverhitSoundEffects { get; } = new(true);
 
+            public ToggleSetting AlwaysOnDrumSFX { get; } = new(true);
```
sonicfind, line 175: Set to `false` by default to match *expected* behavior of the feature.

ProjectSettings/ProjectSettings.asset
```diff
@@ -139,7 +139,7 @@ PlayerSettings:
     Others: 1
   bundleVersion: 0.1.0
   preloadedAssets:
-  - {fileID: 11400000, guid: 2ffed9dcd47320f4986a850f05ff2ad5, type: 2}
+  - {fileID: 0}
```
sonicfind, line 142: Make sure to remove changes like these before pushing. It's usually Unity being funky, but it can cause problems.

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -101,7 +102,17 @@ protected override DrumsEngine CreateEngine()
...
+            {
+                if (!wasNoteHit)
+                {
+                    PlayDrumSoundEffect(action, velocity);
+                }
+                else
+                {
+                    PlayDrumSoundEffect(action, velocity);
+                }
+            }
```
sonicfind, line 115: 1. This breaks hit logic because of the removal of the `return` on `wasNoteHit`. That's not optional; it *has* to happen.
2. Without checking the state of `SettingsManager.Settings.AlwaysOnDrumSFX.Value` on line 147, you're forcing it to make **two** calls to `PlayDrumSoundEffect` in freestyle sections when it is set to `true`
3. 
```
if (!wasNoteHit)
{
    PlayDrumSoundEffect(action, velocity);
}
else
{
    PlayDrumSoundEffect(action, velocity);
}
```
Both branches are doing the same exact thing, so the if-else serves no purpose.
4. Make sure that the formatting of whitespace before each line matches the rest of the code. You're one "tab" (four spaces) too few on each.

sonicfind (changes requested):

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -100,8 +101,17 @@ protected override DrumsEngine CreateEngine()
...
+            // Skip if a note was hit, because we have different logic for that below
+            if (wasNoteHit)
+            {
+                // If AODSFX is turned on and a note was hit, Play the drum sfx. Without this, drum sfx will only play on misses.
+                if (SettingsManager.Settings.AlwaysOnDrumSFX.Value)
+                {
+                    PlayDrumSoundEffect(action, velocity); return;
+                }
+                else return;
+            }
```
sonicfind, line 114: 1. No reason to call `return` in both the the `if` AND the `else`. Just call return *after( the if branch, removing the `else` entirely.
2. Again, make sure that the formatting of whitespace before each line matches the rest of the code. You're one "tab" (four spaces) too few on each.

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -133,7 +143,8 @@ protected override DrumsEngine CreateEngine()
 
                 bool isDrumFreestyle = IsDrumFreestyle();
 
-                if (isDrumFreestyle)
+                // Figure out wether its a drum freestyle or if AODSFX is enabled
+                if (isDrumFreestyle || SettingsManager.Settings.AlwaysOnDrumSFX.Value)
```
sonicfind, line 147: Swap the two. If `SettingsManager.Settings.AlwaysOnDrumSFX.Value` is active, then it'd be always active, making the freestyle unnecessary to test for in those situations.

sonicfind (changes requested):

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -101,7 +102,14 @@ protected override DrumsEngine CreateEngine()
...
                 // Skip if a note was hit, because we have different logic for that below
-                if (wasNoteHit) return;
+                if (wasNoteHit)
+                {
+                    // If AODSFX is turned on and a note was hit, Play the drum sfx. Without this, drum sfx will only play on misses.
+                    if (SettingsManager.Settings.AlwaysOnDrumSFX.Value)
+                    {
+                        PlayDrumSoundEffect(action, velocity);
+                    }
+                } return;
```
sonicfind, line 112: `return` needs to be in the notehit scope.

nameonascreen: Had it right the first time then, thanks.

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -101,7 +102,14 @@ protected override DrumsEngine CreateEngine()
...
             {
                 // Skip if a note was hit, because we have different logic for that below
-                if (wasNoteHit) return;
+                if (wasNoteHit)
+                {
+                    // If AODSFX is turned on and a note was hit, Play the drum sfx. Without this, drum sfx will only play on misses.
+                    if (SettingsManager.Settings.AlwaysOnDrumSFX.Value)
+                    {
+                        PlayDrumSoundEffect(action, velocity);
+                    } return;
```
sonicfind, line 111: `return` on the next line

EliteAsian123 (approved):
Looks good to me

sonicfind (approved):

PR #892, opened 2024-10-06
Countdown rework

Purplo-cf:
CountdownDisplay changes to support YARG.Core PR 218, as well as a fix for this issue:
https://discord.com/channels/1086048856678084609/1288965949411823616

### DIFF
--- a/Assets/Prefabs/Gameplay/HUD/CountdownDisplay.prefab
+++ b/Assets/Prefabs/Gameplay/HUD/CountdownDisplay.prefab
@@ -315,7 +315,7 @@ MonoBehaviour:
   m_fontWeight: 400
   m_enableAutoSizing: 1
   m_fontSizeMin: 18
-  m_fontSizeMax: 72
+  m_fontSizeMax: 64
   m_fontStyle: 0
   m_HorizontalAlignment: 2
   m_VerticalAlignment: 512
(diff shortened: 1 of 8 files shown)


polson:
YARG.Core PR: https://github.com/YARC-Official/YARG.Core/pull/218

polson:
Confirmed that the issue above is fixed in this PR

polson:
However it seems that the countdown may still have issues.  In this video, it ends after 2 instead of after 1

Also, the countdown is not synced to the start of the measures

https://github.com/user-attachments/assets/235cedad-920c-4014-b357-6893d67c6942

Purplo-cf:
> However it seems that the countdown may still have issues. In this video, it ends after 2 instead of after 1
Part of this update was to link countdown appearance times exclusively to the note start/end times instead of the measure lines. The old behavior was reported as an issue in this thread:
https://discord.com/channels/1086048856678084609/1291509593117167688

The countdown fade-out time is now being calculated as _next note time minus 1 second_. Depending on the current tempo and time signature, the measure count in the circle can vary at the one-second mark.

> Also, the countdown is not synced to the start of the measures

That was the other half of this fix - the old system calculated measures off of the BEATS track in the chart, which was independent of the actual time signature in cases where the original charter wanted the character and lighting animations to move at a different pace. I can't tell if that is what's happening in the video you posted without audio to compare.

This video shows a song in 6/8 that renders beat lines in 4/4:
https://discord.com/channels/1086048856678084609/1086052453801283624/1292279296190185594

polson:
Thanks for the explanation.  Sounds like this is working as intended then

Purplo-cf:
> Thanks for the explanation. Sounds like this is working as intended then

Maybe! I will take a look at the I Melt With You chart and see if anything weird is going on with the sync track that I'm not accounting for. Thank you for all of the testing on this branch!

EliteAsian123 (approved):
I still need to check the prefabs in-game, but I'm away right now so I'll do it when I'm back. Code looks good though!

PR #896, opened 2024-10-10
Fix crunchy audio when whammy is enabled

polson:
Fix is to use the `PitchShiftParameters` instead of `PitchShiftParametersStruct` when passing into `Bass.FXSetParameters` initially

After that we can use the struct and it will work fine.

If I try to use `PitchShiftParameters` instead of `PitchShiftParametersStruct` everywhere, pitch shift just stops working.  Probably something to do with the pointer reference, but I couldn't figure it out.  

Crunchy:
https://github.com/user-attachments/assets/47852b74-7c68-4e93-9ce9-bf02f6fdb91f

Fixed:
https://github.com/user-attachments/assets/39f8c618-6571-41d8-b77d-1c8649cad38d

Relates to:
https://discord.com/channels/1086048856678084609/1287813697577685084

### DIFF
--- a/Assets/Script/Audio/Bass/BassAudioManager.cs
+++ b/Assets/Script/Audio/Bass/BassAudioManager.cs
@@ -302,7 +302,7 @@ namespace YARG.Audio.BASS
                             DrumSfxSamples[(int) sfxSample] = sfx;
                         }
                         break;
-                    }  
+                    }
                 }
             }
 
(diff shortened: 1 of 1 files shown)


sonicfind:
LGTM, as I don't have the proper means to test this.

We'll see if this causes any issues.

RileyTheFox:
This change broke pitching completely for some songs, so I'm going to revert it until a better fix is discovered. 
image

PR #898, opened 2024-10-10
Prevent invalid high scores from being saved

thomeval:
This PR adds validation to Band and Solo high scores, meaning that they will be considered invalid and thus not saved when certain conditions are met. This also includes a "Band score not saved" message on the Result screen. Unfortunately, it is not possible to filter out already existing invalid high scores, so the validation rules will only apply to scores set after this PR is merged.

Currently, a band high score will be considered invalid if the Song Speed is less than 100% :
Screenshot 2024-10-08 142841

Or if any bots are active:

Screenshot 2024-10-10 141616

It will be fairly simple to add more rules later by extending the `ScoreContainer.IsSoloScoreValid()` and `ScoreContainer.IsBandScoreValid()` methods.

**Known Limitation:**  In some circumstances, it is possible for a Band Score to be invalid, but a player's Solo Score to still be valid (for example, when playing with bots). In such cases, we should save the player score, but not the band score. Assuming there are no blockers preventing it, this will be addressed in the Solo High Score overhaul PR.

### DIFF
--- a/Assets/Scenes/ScoreScene.unity
+++ b/Assets/Scenes/ScoreScene.unity
@@ -125,6 +125,142 @@ NavMeshSettings:
     debug:
       m_Flags: 0
   m_NavMeshData: {fileID: 0}
+--- !u!1 &126113574
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
+  - component: {fileID: 126113577}
+  - component: {fileID: 126113576}
+  - component: {fileID: 126113575}
+  m_Layer: 5
+  m_Name: Band Score Not Saved Message
+  m_TagString: Untagged
+  m_Icon: {fileID: 0}
+  m_NavMeshLayer: 0
+  m_StaticEditorFlags: 0
+  m_IsActive: 1
+--- !u!114 &126113575
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 126113574}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: f4688fdb7df04437aeb418b961361dc5, type: 3}
+  m_Name: 
+  m_EditorClassIdentifier: 
+  m_Material: {fileID: 0}
+  m_Color: {r: 1, g: 1, b: 1, a: 1}
+  m_RaycastTarget: 1
+  m_RaycastPadding: {x: 0, y: 0, z: 0, w: 0}
+  m_Maskable: 1
+  m_OnCullStateChanged:
+    m_PersistentCalls:
+      m_Calls: []
+  m_text: (Band Score not saved)
+  m_isRightToLeft: 0
+  m_fontAsset: {fileID: 11400000, guid: c45bdba5d22921a4a961427fb9eac360, type: 2}
+  m_sharedMaterial: {fileID: 241437652069387127, guid: c45bdba5d22921a4a961427fb9eac360,
+    type: 2}
+  m_fontSharedMaterials: []
+  m_fontMaterial: {fileID: 0}
+  m_fontMaterials: []
+  m_fontColor32:
+    serializedVersion: 2
+    rgba: 4286940159
+  m_fontColor: {r: 1, g: 0.514151, b: 0.5232028, a: 1}
+  m_enableVertexGradient: 0
+  m_colorMode: 3
+  m_fontColorGradient:
+    topLeft: {r: 1, g: 1, b: 1, a: 1}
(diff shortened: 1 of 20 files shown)


Assets/Script/Menu/ScoreScreen/ScoreScreenContainer.cs
```diff
@@ -19,6 +19,7 @@ public struct ScoreScreenStats
 
         public int BandStars;
         public int BandScore;
+        public float SongSpeed;
```
EliteAsian123, line 22: There already exists a `SongSpeed` property in `PersistentState`, so you can just use that.

Assets/Script/Gameplay/GameManager.cs
```diff
@@ -462,9 +477,9 @@ private bool EndSong()
                     Percent = player.BaseStats.Percent
                 });
             }
-
-            // Record the score into the database (if there's at least 1 non-bot player)
-            if (playerEntries.Count > 0)
+
```
EliteAsian123, line 480: Please ensure you strip whitespace!

Assets/Script/Gameplay/GameManager.cs
```diff
@@ -462,9 +477,9 @@ private bool EndSong()
                     Percent = player.BaseStats.Percent
                 });
             }
-
-            // Record the score into the database (if there's at least 1 non-bot player)
-            if (playerEntries.Count > 0)
+                    
+            // Record the score into the database (but only if there are no bots, and Song Speed is at least 100%)
+            if (ScoreContainer.IsBandScoreValid( SongSpeed))
```
EliteAsian123, line 482: This check should be at the start of the function since that means all of the calculations above aren't actually used. (i.e., `if (...) return;`)

Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
```diff
@@ -172,7 +178,12 @@ private bool AnalyzeReplay(SongEntry songEntry, ReplayInfo? replayEntry)
                 _analyzingReplay = false;
                 return true;
             }
-
+            if (GlobalVariables.State.ScoreScreenStats.Value.PlayerScores.Length == 0)
+            {
+                YargLogger.LogInfo("No human players in ReplayEntry.");
+                _analyzingReplay = false;
+                return true;
+            }
```
EliteAsian123, line 186: This check doesn't actually do anything because `PlayerScores` includes bots. So this is basically just checking if any profiles are in the song (which is always the case).

Assets/Script/Scores/ScoreContainer.cs
```diff
@@ -58,6 +56,32 @@ private static void InitDatabase()
...
+            if (!PlayerContainer.Players.Any())
+            {
+                return false;
+            }
+
+            // If any player is disqualified from a valid Solo Score, this should disqualify the Band Score as well.
+            if (PlayerContainer.Players.Any(e => !IsSoloScoreValid(songSpeed, e)))
+            {
+                return false;
+            }
```
EliteAsian123, line 70: This check fails if bots are sitting out, even though players that sit out don't effect the score

Assets/Script/Player/PlayerContainer.cs
```diff
@@ -47,6 +48,11 @@ public static class PlayerContainer
         /// </summary>
         public static List<YargPlayer>.Enumerator PlayerEnumerator => _players.GetEnumerator();
 
+        /// <summary>
+        /// Returns whether or not there are any bots in the game.
+        /// </summary>
+        public static bool HasBots => _players.Any(i => i.Profile.IsBot);
```
EliteAsian123, line 54: This property is unused

Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
```diff
@@ -85,6 +89,8 @@ private void OnEnable()
             // Set text
             _songTitle.text = song.Name;
             _artistName.text = song.Artist;
+            _bandScoreNotSavedMessage.gameObject.SetActive(!ScoreContainer.IsBandScoreValid(scoreScreenStats.SongSpeed));
+            _bandScoreNotSavedMessage.text = Localize.Key("Menu.ScoreScreen.BandScoreNotSaved");
```
EliteAsian123, line 93: Use the "Localize Text" component instead, and just reference the game object in the serialized field.

Assets/Scenes/ScoreScene.unity
EliteAsian123, line 1: image
The text should be a child of the header and it should be anchored to the right-side for it to work on other resolutions

EliteAsian123 (changes requested):
Hello! Sorry for the late review again.

After some changes below, this should be good to go!

EliteAsian123:
> In some circumstances, it is possible for a Band Score to be invalid, but a player's Solo Score to still be valid (for example, when playing with bots). 

For future proofing, it's probably better for this not to be a thing whenever band multipliers and stuff come into play.

Assets/Script/Scores/ScoreContainer.cs
thomeval: Nice catch! I completely missed that.

EliteAsian123 (approved):
The code is all good! I'm away right now so once I get back I'll test this in-game and this should be good to merge (if one of the other maintainers could do that, that could work too)

PR #902, opened 2024-10-13
Add an option to disable the search button

nameonascreen:
Adds an option to disable the search button. Disabled by default, but useful for users who accidentally press search and then need to use their kbm to back out of it.


EliteAsian123:
Hello, I've thought about this PR for a long while, however, I've decided to close it for now.

The reason is that the underlying problem is that the navigation system is kind of not great at the moment, so adding an option to turn off this button seems more like a work around. We're planning to redo all of the menus in v0.14 or v0.15.

Thanks anyways!

PR #905, opened 2024-10-18
Customizable Highway Presets

thomeval:
This PR introduces a new type of preset to YARG: Highway Presets. These are used to customize the colours of the note highway. This includes the "default" colours, as well as the glow from Groove (when at max multiplier) and when Star Power is active.

This change also includes three default presets out of the box:
YARG-HighwayPresets
*From left to right: Default, Basic, No Groove, and a custom preset.*

Custom presets can be edited from the Settings menu:
image

Requires YARG.Core PR https://github.com/YARC-Official/YARG.Core/pull/220


Assets/Script/Gameplay/GameManager.cs
```diff
@@ -524,6 +524,7 @@ public void SetEditHUD(bool on)
             var replayStats = new List<ReplayStats>(_players.Count);
             var colorProfiles = new Dictionary<Guid, ColorProfile>();
             var cameraPresets = new Dictionary<Guid, CameraPreset>();
+            var highwayPresets = new Dictionary<Guid, HighwayPreset>();
```
EliteAsian123, line 527: This variable doesn't appear to be used anymore (contents are updated, but not utilized)

Assets/Script/Gameplay/Visuals/TrackMaterial.cs
```diff
@@ -32,10 +35,38 @@ public struct Layer
             public Layer Layer2;
             public Layer Layer3;
             public Layer Layer4;
+
+            public static Preset FromColors(System.Drawing.Color[] colors)
```
EliteAsian123, line 39: Instead of doing something like `FromColors`, it might be better to do something like `FromHighwayPreset` where you can directly provide the highway preset as an argument and set up the struct's values from there, instead of allocating an addition array. The `BackgroundBaseColors` and the `BackgroundGrooveBaseColors` can then also be removed.

Assets/Script/Settings/Preview/FakeTrackPlayer.cs
```diff
@@ -264,9 +269,11 @@ private void OnSettingChanged()
             var cameraPreset = PresetsTab.GetLastSelectedPreset(CustomContentManager.CameraSettings);
             var colorProfile = PresetsTab.GetLastSelectedPreset(CustomContentManager.ColorProfiles);
             var enginePreset = PresetsTab.GetLastSelectedPreset(CustomContentManager.EnginePresets);
+            var highwayPreset = PresetsTab.GetLastSelectedPreset(CustomContentManager.HighwayPresets);
 
+            var currentColorProfile = CurrentGameModeInfo.FretColorProvider(colorProfile);
```
EliteAsian123, line 274: Is this addition intentional? I don't see the variable being used.

EliteAsian123 (changes requested):
Hello! Thanks for the PR, this is awesome! 

I've just requested some simple changes below. Once the YARG.Core PR is merged, you can update the YARG.Core pointer in this PR and then this can be merged as well!

Assets/Script/Gameplay/Visuals/TrackMaterial.cs
EliteAsian123: For now, you can even just completely get rid of the `Layer` struct and replace those fields with colors. Once we add texture support, it's likely some other things in this class will have to be redone anyways, so removing it for now is fine!

thomeval:
Okay, PR is updated, ready for review.

PR #907, opened 2024-10-25
Slightly improve high score loading performance

thomeval:
This PR slightly improves performance when loading high scores on the Music Library screen, which triggers when entering the Music Library screen, as well as whenever the sorting is changed. Note that the performance improvement is modest at best under normal circumstances - we're looking at a 0.2ms improvement with a music library of about 500 songs. Note that the first load is always much slower as the scores have to be fetched from the database. Subsequent calls are cached, and this is where the majority of the benefit will be seen.

Screenshot 2024-10-25 134408
Before change

Screenshot 2024-10-25 134433
After change


EliteAsian123:
I'll merge this since it's an improvement, however, it's still somewhat slow so more should be made in the future. 

Thanks!

PR #875, opened 2024-09-21
Trill and Tremolo lanes

Purplo-cf:
Support for 5-fret, drums, and pro keys. Requires YARG.Core PR 212

lanes_example_guitar
lanes_example_drums
lanes_example_keys

PR #909, opened 2024-10-27
Trill and Tremolo Lanes

Purplo-cf:
Requires YARG.Core PR 222.

Assets and rendering behavior for trill and tremolo lanes.


Purplo-cf:
Refer to the previous PR for reasoning behind the Unity prefab and material changes:
https://github.com/YARC-Official/YARG/pull/875

wyrdough:
Not going to do a review or anything, but just to flag something found in play testing so it doesn't get lost, the lane colors are not flipped for five fret instruments when lefty flip is active.

PR #920, opened 2024-11-19
Display player names at start of song

thomeval:
This PR adds a display for the player's name and current instrument that is shown for 3 seconds at the beginning of the song (similar to RB behaviour).

YARG-PlayerNames

All instruments are supported. For harmony vocalists, each player's harmony part and their needle is displayed as well, making it much easier to see who's singing what.

YARG-VocalPlayerNames

Solves these feature requests: https://discord.com/channels/1086048856678084609/1301954616065265805
https://discord.com/channels/1086048856678084609/1296184560270446602

Known issue: If a non-vocals part starts empty, the countdown display will obscure the player name display.


EliteAsian123 (changes requested):
Hello, thanks for the PR and again, sorry for the long wait.

I haven't looked at the code yet (I'll be doing that within the next few days), but I thought I'd point out some things just from testing it out in-game.

First off, everything looks great, however there are two things that are a bit off. Firstly, I'm not a big fan of the animation used for showing/hiding the player name display. I think just a simple fade-out would look better in this case. Secondly, the font for the vocal player name being different from the track player name is weird to me. I know the performance text uses a different font, but making them match would be the best I believe. I talked to Kadu about this and he agreed and said he's gonna do some reorganizing of the fonts we use overall.

Along with this, the colors for the harmony icons shouldn't be baked into the texture as eventually those will be customizable. The color should just be applied to the image component, and if you make the icon white, this should work just fine.

Overall, great work! I'll do some code review later, as I've mentioned. Have a great night!

Assets/Script/Gameplay/HUD/PlayerNameDisplay.cs
```diff
@@ -0,0 +1,99 @@
...
+        public void ShowPlayer(YargPlayer player, int needleId)
+        {
+            var textureNeedle = $"VocalNeedleTexture/{needleId}";
+            _needleIcon.texture = Addressables.LoadAssetAsync<Texture2D>(textureNeedle).WaitForCompletion();
+            _instrumentIcon.color = GetHarmonyColor(player);
+            ShowPlayer(player);
+        }
+
+        // TODO: Temporary until color profiles for vocals. Duplicated from VocalTrack class.
+        public readonly Color[] HarmonyColors =
```
EliteAsian123, line 67: All fields should be at the top of the file (if possible, the VocalTrack class should just expose this as a static variable)

EliteAsian123 (changes requested):
Hello, I'll look at the prefabs first thing tomorrow, however there's just a simple adjustment in the code that I've pointed out here. Thanks!

Assets/Script/Gameplay/HUD/PlayerNameDisplay.cs
thomeval: Good idea, I've updated the PR to do just that.

EliteAsian123 (approved):
Right so at this point the PR is good to go but I did notice one thing that I will bring up in the Discord to others, and that is that the player icons are still shown in the first second of the chart, which likely won't be a problem, but it's definitely possible that there could be notes there. Now, we could either (quite easily) extend the delay before the chart starts, or just say that any song that has notes at the very beginning of the chart is bad practice and ignore the problem, but I'm not sure what the best course of action is. I'd prefer the former but I noticed that RB basically does the latter so, there's that. Thanks!

PR #921, opened 2024-11-20
Added linux file opener support

8D1:
Linux typically uses XDG-open to handle file paths and opening things. 

Dealing with .NET could be considered torture


8D1:
I closed it accidentally thinking i opened it in the MASTER branch, thats on me. 
However this worked on gentoo linux w/ kde plasma & dolphin

PR #923, opened 2024-11-21
Change default tilt activation threshold for Riffmaster

wyrdough:
The default tilt sensor activation threshold being 1.0 makes it impossible to activate SP with tilt by default on the Riffmaster. The maximum possible value appears to be somewhere between 0.9 and 0.95.

This PR changes the default activation threshold to 0.7, which seems to still provide plenty of headroom for avoiding unintentional activation.

### DIFF
--- a/Assets/Script/Input/Bindings/Defaults/BindingCollection.FiveFretGuitar.cs
+++ b/Assets/Script/Input/Bindings/Defaults/BindingCollection.FiveFretGuitar.cs
@@ -7,6 +7,7 @@ namespace YARG.Input
     public partial class BindingCollection
     {
         private static ActuationSettings _tiltSettings = new() { ButtonPressThreshold = 1.0f };
+        private static readonly ActuationSettings _riffmasterTiltSettings = new() { ButtonPressThreshold = 0.7f };
 
         private bool SetDefaultGameplayBindings(FiveFretGuitar guitar)
         {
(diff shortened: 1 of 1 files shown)


TheNathannator (changes requested):
This will mess greatly with Les Paul guitars which have an over-sensitive tilt axis, they max out at around a 45 degree angle. I'd prefer if this change was applied only for Riffmaster guitars specifically, which you can do by testing for the `RiffmasterGuitar` type in the function directly below this.

wyrdough:
> This will mess greatly with Les Paul guitars which have an over-sensitive tilt axis, they max out at around a 45 degree angle. I'd prefer if this change was applied only for Riffmaster guitars specifically, which you can do by testing for the `RiffmasterGuitar` type in the function directly below this.

Done. 

It compiles and works for me, but I don't know C# so let me know if I inadvertently did a stupid.

Assets/Script/Input/Bindings/Defaults/BindingCollection.FiveFretGuitar.cs
```diff
@@ -38,8 +35,18 @@ private bool SetDefaultGameplayBindings(FiveFretGuitar guitar)
                 AddBinding(GuitarAction.YellowFret, rb.soloYellow);
                 AddBinding(GuitarAction.BlueFret, rb.soloBlue);
                 AddBinding(GuitarAction.OrangeFret, rb.soloOrange);
+
+                if (guitar is RiffmasterGuitar riff)
+                {
+                    BindingCollection._tiltSettings = new() { ButtonPressThreshold = 0.7f };
```
TheNathannator, line 41: `_tiltSettings` is static (global) and so modifying it here will change the default for *all* guitars which get connected from this point onward in the current session. Instead of overwriting it, make a new static variable and use the type test as a condition for which one to use:

```diff
    private static ActuationSettings _tiltSettings = new() { ButtonPressThreshold = 1.0f };
+   private static readonly ActuationSettings _riffmasterTiltSettings = new() { ButtonPressThreshold = 0.7f };

...

-   AddBinding(GuitarAction.StarPower, guitar.tilt, _tiltSettings);
+   var tiltSettings = guitar is RiffmasterGuitar ? _riffmasterTiltSettings : _tiltSettings;
+   AddBinding(GuitarAction.StarPower, guitar.tilt, tiltSettings);
```

(As a side-note, there's also no need to do the full `BindingCollection._tiltSettings` access specifically since this is already inside that class, just `_tiltSettings` will do.)

wyrdough: Thanks for the feedback. I believe the latest commit addresses the issue.

TheNathannator: Yep, looks good now!

TheNathannator (approved):

TheNathannator:
Thanks for the adjustment!

Probably not a bad idea to more carefully tune the default threshold for other guitars as well, but for the time being there isn't a way to determine what platform/model a guitar is outside of what's provided as a base class. Something to look into in the future lol

PR #925, opened 2024-11-27
Add bindings for Rock Band and Riffmaster guitar solo buttons

wyrdough:
This PR adds bindings for solo buttons as found on Rock Band guitars and the Riffmaster and changes the default bindings for those devices to bind solo buttons as solo buttons instead of regular buttons.

See [YARG.Core PR #224](https://github.com/YARC-Official/YARG.Core/pull/224) for engine support.

Related graphical effects are included in #936

### DIFF
--- a/Assets/Art/Menu/Common/ModifierIcons.png
+++ b/Assets/Art/Menu/Common/ModifierIcons.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b901f0b9b0a49354b441d7b9027d44591f8fd493d1c81a2133643f0f30d10410
-size 1874372
+oid sha256:aa97cf3e3c5ee473c3a91bdbfb8a0ecc2b09b90483ea0a76f03075574ff61e59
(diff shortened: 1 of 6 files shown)


wyrdough:
Feedback on the shader effect stuff in the last couple of commits would be greatly appreciated.

wyrdough:
This has been updated to use an object for solo effects. All shader stuff relating to solos has also been removed. The engine modifier icon is not yet final and I'm sure changes to the texture on the solo object will be required, but I believe the code is in a good enough state to merit review.

wyrdough:
(I know that did not go well, giving it another shot in a minute)

wyrdough:
I don't quite understand why GitHub is saying there is still a conflict with ModifierIcons since those were added after merging engine-fixes into fivefret-solo. It is updated to have the latest set of icons plus the solo taps icon, though, so you should be able to use the versions from fivefret-solo without screwing anything up.

wyrdough:
Fixed now...

PR #931, opened 2024-12-24
Add option to sort by play count

wyrdough:
This PR adds a new sorting option to the song library to sort by play count.

When sorting by play count, two categories are displayed, played and unplayed. Played songs are sorted by the number of times played by the first connected profile, in descending order. Unplayed songs are sorted by whatever sort order was in use immediately prior to switching to play count order.

### DIFF
--- a/Assets/Script/Scores/ScoreContainer.cs
+++ b/Assets/Script/Scores/ScoreContainer.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using SQLite;
 using YARG.Core;
+using YARG.Core.Game;
 using YARG.Core.Logging;
 using YARG.Core.Song;
 using YARG.Helpers;
@@ -405,6 +406,34 @@ namespace YARG.Scores
             }
 
             return new List<SongEntry>();
+        }
+
+        // order true sorts descending, false ascending
+        // this is the same as GetMostPlayedSongs, but is limited to the one profile and returns the entire list
+        public static List<SongEntry> GetPlayedSongsForUserByPlaycount(YargProfile profile, bool order)
+        {
+            // This may actually need to return a list of structs like (SongEntry, playCount)
+            // For now we can try returning a list of songs in order of play count
+            var sortOrder = order ? "DESC" : "ASC";
+            var profileId = profile.Id;
+            var songList = new List<SongEntry>();
+            var query = "SELECT COUNT(GameRecords.Id), GameRecords.SongChecksum from GameRecords, PlayerScores " +
+                "WHERE PlayerScores.GameRecordId = GameRecords.Id " +
+                $"AND PlayerScores.PlayerId = '{profileId}' " +
+                $"GROUP BY GameRecords.SongChecksum ORDER BY COUNT(GameRecords.Id) {sortOrder}";
+
+            YargLogger.LogDebug($"SQL Query: {query}");
+
+            var playCounts = _db.Query<PlayCountRecord>(query);
+            foreach (var record in playCounts)
+            {
+                var hash = HashWrapper.Create(record.SongChecksum);
+                if (SongContainer.SongsByHash.TryGetValue(hash, out var list))
+                {
+                    songList.Add(list.Pick());
+                }
+            }
+            return songList;
         }
 
         public static void Destroy()
--- a/Assets/Script/Scores/ScoreContainer.cs
+++ b/Assets/Script/Scores/ScoreContainer.cs
@@ -412,8 +412,6 @@ namespace YARG.Scores
         // this is the same as GetMostPlayedSongs, but is limited to the one profile and returns the entire list
         public static List<SongEntry> GetPlayedSongsForUserByPlaycount(YargProfile profile, bool order)
         {
-            // This may actually need to return a list of structs like (SongEntry, playCount)
-            // For now we can try returning a list of songs in order of play count
             var sortOrder = order ? "DESC" : "ASC";
             var profileId = profile.Id;
             var songList = new List<SongEntry>();
--- a/Assets/Script/Scores/ScoreContainer.cs
+++ b/Assets/Script/Scores/ScoreContainer.cs
@@ -414,13 +414,19 @@ namespace YARG.Scores
         {
             var sortOrder = order ? "DESC" : "ASC";
             var profileId = profile.Id;
+            var profileInstrument = (int) profile.CurrentInstrument;
             var songList = new List<SongEntry>();
             var query = "SELECT COUNT(GameRecords.Id), GameRecords.SongChecksum from GameRecords, PlayerScores " +
                 "WHERE PlayerScores.GameRecordId = GameRecords.Id " +
-                $"AND PlayerScores.PlayerId = '{profileId}' " +
-                $"GROUP BY GameRecords.SongChecksum ORDER BY COUNT(GameRecords.Id) {sortOrder}";
+                $"AND PlayerScores.PlayerId = '{profileId}' ";
 
-            YargLogger.LogDebug($"SQL Query: {query}");
+            // If the profile instrument is bad, we can still return all scores for the profile
+            if (profile.HasValidInstrument)
+            {
+                query += $"AND PlayerScores.Instrument = {profileInstrument} ";
+            }
+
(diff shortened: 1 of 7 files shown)


Assets/Script/Song/SongContainer.cs
```diff
@@ -280,6 +286,32 @@ public static SongEntry GetRandomSong()
             return _songs.Pick();
         }
 
+        private static SongCategory[] GetPlaycounts()
+        {
+            var player = PlayerContainer.Players.First();
```
sonicfind, line 291: Could be a bot. Need a more full-proof method to determine this which would involve reworking `PlayerContainer` fundamentally.

Assets/Script/Song/SongContainer.cs
```diff
@@ -2,6 +2,7 @@
 using YARG.Core.Song.Cache;
 using YARG.Core.Song;
 using System;
+using System.Linq;
```
sonicfind, line 5: Please refrain from LINQ

Assets/Script/Song/SongContainer.cs
```diff
@@ -280,6 +286,32 @@ public static SongEntry GetRandomSong()
...
+            {
+                // I don't think this should ever happen, but I'm not certain,
+                // so belt and suspenders wins.
+                previousSort = SortAttribute.Name;
+            }
+
+            foreach (var category in GetSortedCategory(previousSort))
+            {
+                var songs = category.Songs.Except(counts);
+                zeroPlaySongs.AddRange(songs);
```
sonicfind, line 307: Please refrain from LINQ. We should always aim to dodge hidden allocations of enumerable containers.
```suggestion
                foreach (var song in category.Songs)
                {
                    if (!counts.Contains(song))
                    {
                        zeroPlaySongs.Add(song);
                    }
                }
```

Assets/Script/Song/SongContainer.cs
```diff
@@ -94,6 +97,8 @@ public static class SongContainer
         private static SongCategory[] _sortDatesAdded = Array.Empty<SongCategory>();
         private static Dictionary<Instrument, SongCategory[]> _sortInstruments = new();
 
+        private static SongCategory[] _sortPlaycounts = Array.Empty<SongCategory>();
+
```
sonicfind, line 101: You should follow the example of the `Playable` sort here. The point of `_playable` starting as null is to communicate to `GetPlayableSongs()` that it needs to regenerate the collection after a re-scan. See `FillContainers()`.

Assets/Script/Song/SongContainer.cs
sonicfind: Additionally, without checking for the prior state of the player used to determine the order, you'll end up needlessly re-generating the same list even when the player is the same. You need to keep track of the playerId last used to know whether the whole process actually needs to occur.

Assets/Script/Song/SongContainer.cs
```diff
@@ -280,6 +286,32 @@ public static SongEntry GetRandomSong()
...
+            foreach (var category in GetSortedCategory(previousSort))
+            {
+                var songs = category.Songs.Except(counts);
+                zeroPlaySongs.AddRange(songs);
+            }
+            var countCategories = new SongCategory[2];
+            countCategories[0] = new SongCategory("PLAYED SONGS", counts.ToArray(), "Played Songs");
+            countCategories[1] = new SongCategory("UNPLAYED SONGS", zeroPlaySongs.ToArray(), "Unplayed Songs");
+            return countCategories;
```
sonicfind, line 312: You're not even using the `_sortPlayCounts` variable here to cache the result.

Assets/Script/Song/SongContainer.cs
sonicfind: OH, and the last pre-sort used too. There's a lot you need to track (and is generally why I opted to only ever use the name sort as the sub-sort).

That only gets *more* problematic when taking into account how playable sort works.

sonicfind (changes requested):

Assets/Script/Song/SongContainer.cs
wyrdough: I get the concern in the game loop, but I do wonder why the use of LINQ is a problem in menus etc?

wyrdough:
Updated to remove _sortPlayCounts variable, deal properly with case where no profiles are connected or only bots are connected, and remove the use of Except to filter played songs from unplayed list.

I'm still using a LINQ expression to evaluate the connected profiles, as that is how it is done every other place in YARG that checks whether a non-bot profile is connected.

Assets/Script/Song/SongContainer.cs
sonicfind: You already changed it but:

Whenever you perform some LINQ operation that creates a collection, especially within a loop, you add extra pressure on GC due to the collection that's generated behind the scenes. It may not seem like it's too important due to it being a menu, but GC may hold onto that collection for a while. That creates a higher likelihood that gameplay will halt from GC collecting that memory back.

sonicfind:
> I'm still using a LINQ expression to evaluate the connected profiles, as that is how it is done every other place in YARG that checks whether a non-bot profile is connected.

I'm cool with that. The only allocations there would be *maybe* the lambda (if that isn't inlined) and a stack allocated integer.

Assets/Script/Song/SongContainer.cs
```diff
@@ -191,7 +195,14 @@ public static SongCategory[] GetSortedCategory(SortAttribute sort)
...
                 SortAttribute.Band           => _sortInstruments[Instrument.Band],
-                _ => throw new Exception("stoopid"),
+                // Make life better when people go back a version and we
+                // encounter sorts we don't understand by providing a
+                // default rather than a blank song library
+                _ => new Func<SongCategory[]>( () =>
+                {
+                    YargLogger.LogInfo("Invalid Sort Attribute. Defaulting to Name sort.");
+                    return _sortTitles;
+                })(),
```
EliteAsian123, line 205: This syntax here is kinda hacky. It's probably just better to first, put this switch expression result into a variable, and set the default case to return `null`. Then, check to see if the result variable is `null`, and then put this code there. Return the result variable.

Assets/Script/Song/SongContainer.cs
```diff
@@ -280,6 +291,47 @@ public static SongEntry GetRandomSong()
...
         }
 
+        // Play count sorting is intentionally not cached, as it must be regenerated after
+        // every play, when profiles change, and probably a bunch of other stuff
+        private static SongCategory[] GetPlaycounts()
+        {
+            // This should never happen since play count shouldn't be selectable without
+            // a non-bot profile and MusicLibraryMenu already checks for this, but let's double check
+            if (PlayerContainer.Players.Count(e => !e.Profile.IsBot) == 0)
+            {
```
EliteAsian123, line 301: The `PlayerContainer.Players.Count(e => !e.Profile.IsBot) == 0` is reused a couple of times, so it should probably become a method within `PlayerContainer`.

Assets/Script/Scores/ScoreContainer.cs
```diff
@@ -407,6 +408,39 @@ public static List<SongEntry> GetMostPlayedSongs(int maxCount)
...
+            query += $"GROUP BY GameRecords.SongChecksum ORDER BY COUNT(GameRecords.Id) {sortOrder}";
+
+            var playCounts = _db.Query<PlayCountRecord>(query);
+            foreach (var record in playCounts)
+            {
+                var hash = HashWrapper.Create(record.SongChecksum);
+                if (SongContainer.SongsByHash.TryGetValue(hash, out var list))
+                {
+                    // If there are songs with a duplicate hash, one will be chosen randomly
+                    songList.Add(list.Pick());
```
EliteAsian123, line 438: I'm pretty sure in this circumstance, all of the songs in the `out var list` should be added to the `songList`. This is because there won't be more than one element in the `out var list` when the "Allow Duplicate Songs" option is disabled. I may be wrong, but to me, that makes more sense here.

EliteAsian123 (changes requested):
Hello!

In general, the queries to the databases are very slow, so it could hurt performance when sorting by play count. However, there's nothing you can really do about that at the moment, so it's fine for now.

Basically now just waiting for @sonicfind's approval, and we're good!

PR #940, opened 2025-01-15
Inhibit drum SP activation gems during active SP

wyrdough:
If the player has more than half a bar of SP during a drum fill phrase, the activation gem will be generated even if SP is active. 

This is because FourLaneDrumsNoteElement checks Player.Engine.CanStarPowerActivate to determine whether notes with the IsStarPowerActivator flag set should have their gem colored with the activation color or the normal color. CanStarPowerActivate does not consider SP activation status, so the gem ends up being purple when it shouldn't be.

This PR conditions the use of the activation color on SP not currently being active when the activator gem is spawned.

### DIFF
--- a/Assets/Script/Gameplay/Visuals/TrackElements/Drums/FiveLaneDrumsNoteElement.cs
+++ b/Assets/Script/Gameplay/Visuals/TrackElements/Drums/FiveLaneDrumsNoteElement.cs
@@ -71,7 +71,7 @@ namespace YARG.Gameplay.Visuals
             // Get colors
             var colorNoStarPower = colors.GetNoteColor(pad);
             var color = colorNoStarPower;
-            if (NoteRef.IsStarPowerActivator && Player.Engine.CanStarPowerActivate)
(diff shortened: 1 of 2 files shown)


EliteAsian123 (changes requested):
Hello, thanks for the PR, and sorry for the late review as this slipped under my radar for some reason!

My question is whether its intention that this is only changed visually and not engine-wise. The code that determines the activator note here, has the exact same condition:
image
And here as well:
image

I'm not actually sure which is correct, so I'm open to feedback.

Thanks!

wyrdough:
My thinking was that since SP is already active it doesn't really matter whether the engine thinks there are activation notes or not since the worst case scenario (in my mind) was that the engine attempts to activate SP again if the player plays the now indistinguishable note(s).

EliteAsian123:
The thing is, it's impossible to miss activator notes. As you can see from the first picture there if the activation note exists in the backend of the hit window, the note automatically gets hit, so there is a technical difference.

PR #941, opened 2025-01-15
Don't attempt to pulse the star meter once gold has been achieved.

wyrdough:
After earning 5 gold stars, an error is repeatedly emitted to the debug log: "game object with animator is inactive". This happens because the gold star animator's Play method continues being called every second beatline even after the normal star object has been disabled.

This PR conditions the call to PulseGoldMeter on 5GS not having been reached, resolving the log spam.

### DIFF
--- a/Assets/Script/Gameplay/HUD/ScoreBox/StarScoreDisplay.cs
+++ b/Assets/Script/Gameplay/HUD/ScoreBox/StarScoreDisplay.cs
@@ -37,7 +37,7 @@ namespace YARG.Gameplay.HUD
                 return;
 
             _goldMeterBeatCount++;
-            if (_goldMeterBeatCount % 2 == 0 && _currentStar == 5)
+            if (_goldMeterBeatCount % 2 == 0 && _currentStar == 5 && !_isGoldAchieved)
             {
                 foreach (var star in _starObjects)
                 {


Assets/Script/Gameplay/HUD/ScoreBox/StarScoreDisplay.cs
```diff
@@ -37,7 +37,7 @@ private void PulseGoldMeter(Beatline beat)
                 return;
 
             _goldMeterBeatCount++;
-            if (_goldMeterBeatCount % 2 == 0 && _currentStar == 5)
+            if (_goldMeterBeatCount % 2 == 0 && _currentStar == 5 && _isGoldAchieved == false)
```
EliteAsian123, line 40: For booleans, it is more concise to use `!_isGoldAchieved` instead of `_isGoldAchieved == false`.

EliteAsian123 (changes requested):
Thanks for the PR! Please see the requested changes below. After they are fixed, this can be merged!

Assets/Script/Gameplay/HUD/ScoreBox/StarScoreDisplay.cs
wyrdough: Fixed! Thanks for taking the time to review.

PR #944, opened 2025-01-17
Fix venue custom shaders on linux and OSX

theli-ua:
Its unclear why that shader code was added but it actually breaks all custom shaders used in venues, removing that linux/osx specific code makes custom shaders in venues work on Linux


theli-ua:
I have also now tested this on Intel OSX, this also fixes custom shaders for it as well. That code that this is removing does not seem to be required

PR #949, opened 2025-01-25
Generalize and localize `prc`-based practice sections

Frickitickitavi:
**Requires YARG.Core PR 230**

RB uses [a static list of practice sections](http://docs.c3universe.com/rbndocs/index.php?title=All_Practice_Sections). YARG currently parses these events by transforming the contents of the events themselves, but this doesn't always result in the correct name. For example, `[prc_a1]` is meant to render as "A Section 1", but appears in YARG as "A1". Furthermore, using preset section names means that they can be localized, and the charting team has been sticking to these events as much as possible for that reason.

One drawback of RB's system is that it has arbitrary limitations on how many times a section can be repeated (e.g. verses go up to 9, but pre-verses only go up to 5) and how many subsections a section can have (e.g. Guitar Solo 1 goes up to 1S, but Guitar Solo 9 only goes up to 9N).

This PR integrates the standard practice sections into the localization system, and generalizes them to remove the arbitrary limitations. Section numbers can be any integer with any number of digits - you can have Guitar Solo 345678 - while subsection letters have to be single-letter and are capped at Z. Special consideration is given to the "letter-based" sections (A Section, B Section, etc), which are also generalized all the way up to Z Section.

In all cases, the section number and subsection letter are preserved regardless of language setting, and it is only the proper _name_ of the section that gets localized, like Guitar Solo, Breakdown, or Post-Chorus.

If a `prc`-based practice section fails to match any standardized tag, then it is parsed in the same way that the game currently parses all practice sections. Behavior for `section`-based events is unchanged.

Because the localization must happen in YARG to take advantage of localization, YARG.Core now performs none of the transformation and only checks whether an event begins with `prc` or `section`, then passes it to YARG unedited if so.

The internal list of recognized tags includes some weird edge cases - for example, the RBN docs contain a typo for the "Scratch break" section tag (`prc_sctrach_break`), so the list recognizes both the typo'd name and the intended one. Similarly, "Keyboard enters" is strangely encoded as `[prc_keyb_enters]` despite all other keyboard-related sections spelling the word in full, so I accepted both.


Frickitickitavi:
Accidentally targeted `master` instead of `dev`; reopening with the correct target

PR #950, opened 2025-01-25
Generalize and localize standard practice section events

Frickitickitavi:
RB uses [a static list of practice sections](http://docs.c3universe.com/rbndocs/index.php?title=All_Practice_Sections). YARG currently parses these events by transforming the contents of the events themselves, but this doesn't always result in the correct name. For example, `[prc_a1]` is meant to render as "A Section 1", but appears in YARG as "A1". Furthermore, using preset section names means that they can be localized, and the charting team has been sticking to these events as much as possible for that reason.

One drawback of RB's system is that it has arbitrary limitations on how many times a section can be repeated (e.g. verses go up to 9, but pre-verses only go up to 5) and how many subsections a section can have (e.g. Guitar Solo 1 goes up to 1S, but Guitar Solo 9 only goes up to 9N).

This PR integrates the standard practice sections into the localization system, and generalizes them to remove the arbitrary limitations. Section numbers can be any integer with any number of digits - you can have Guitar Solo 345678 - while subsection letters have to be single-letter and are capped at Z. Special consideration is given to the "letter-based" sections (A Section, B Section, etc), which are also generalized all the way up to Z Section.

In all cases, the section number and subsection letter are preserved regardless of language setting, and it is only the proper _name_ of the section that gets localized, like Guitar Solo, Breakdown, or Post-Chorus.

If a practice section fails to match any standardized tag, then it is parsed in the same way that the game currently parses all practice sections.


Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
```diff
@@ -101,5 +105,223 @@ public void UpdateView()
...
+                    return Localize.Key("Gameplay.PracticeSections." + prcNameToLocalizationKey[key]);
+                }
+
+                if (prcSectionName.StartsWith(key))
+                {
+                    var remainder = prcSectionName[(key.Length + "_".Length)..];
+
+                    if (isSectionNumber(remainder))
+                    {
+                        return Localize.Key("Gameplay.PracticeSections." + prcNameToLocalizationKey[key]) + " " + remainder;
```
TheNathannator, line 150: Again here, the localization string should be a format string instead of manually constructed.

```suggestion
                        return Localize.KeyFormat(("Gameplay.PracticeSections", prcNameToLocalizationKey[key], "WithNumber"), remainder);
```

This would somewhat complicate the localization strings for each section, as both no-number and with-number variants would need to be provided for each.

Assets/StreamingAssets/lang/en-US.json
```diff
@@ -336,7 +336,138 @@
...
+				"Chant": "Chant",
+				"SpokenWord": "Spoken word",
+				"Outro": "Outro",
+				"OutroSolo": "Outro solo",
+				"OutroChorus": "Outro chorus",
+				"Ending": "Ending",
+				"BigRockEnding": "Big Rock Ending!",
+				"FadeOut": "Fade out",
+				"Section": "section"
+		}
```
TheNathannator, line 470: Don't add untranslated entries for new keys into the files for other languages, please. The game can handle them being empty just fine, it'll display the localization key as-is to make it obvious something isn't translated yet.

TheNathannator (changes requested):
I like the concept, but I'm not so sure about the current implementation of things.

There are also plenty of section names supported by Rock Band which aren't present within the RBN section name list, which we could consider adding down the line, but for now I think the current list will do.

Assets/StreamingAssets/lang/en-US.json
TheNathannator: i put this comment on the wrong file lol, but you get my point

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
```diff
@@ -101,5 +105,223 @@ public void UpdateView()
...
+            {"prayer", "Prayer"},
+            {"chant", "Chant"},
+            {"spoken_word", "SpokenWord"},
+            {"outro", "Outro"},
+            {"outro_solo", "OutroSolo"},
+            {"outro_chorus", "OutroChorus"},
+            {"ending", "Ending"},
+            {"bre", "BigRockEnding"},
+            {"fade_out", "FadeOut"}
+        };
```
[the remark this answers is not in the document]
Frickitickitavi: I'm perfectly happy to have them as snake case in the localization files too, and that was actually how I started doing it at first before I got cold feet about going against convention. I'll do that

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
```diff
@@ -101,5 +105,223 @@ public void UpdateView()
...
+
+        private string ParseSectionName(string sectionName)
+        {
+            const string SECTION_PREFIX = "section ";
+            const string PRC_PREFIX = "prc_";
+
+            if (sectionName.StartsWith(SECTION_PREFIX))
+            {
+                return sectionName[SECTION_PREFIX.Length..].TrimStart('_').Trim();
+            }
```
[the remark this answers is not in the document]
Frickitickitavi: Clarification - do you want `section` events to be localized when they correspond to a `prc` event with spaces instead of underscores (e.g. `[section gtr solo 2a]`), or when they correspond to the English rendering of a `prc` event (e.g. `[section Gtr Solo 2A]`)?

Assets/StreamingAssets/lang/en-US.json
Frickitickitavi: I noticed that the non-English localization files are full of English localizations for (almost?) everything. Is there a reason why those are filled in with English, but these should remain empty?

RileyTheFox: > I noticed that the non-English localization files are full of English localizations for (almost?) everything. Is there a reason why those are filled in with English, but these should remain empty?

I think this is something to do with crowdin filling it in as the non-English files are based off of the English file.

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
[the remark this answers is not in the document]
TheNathannator: No, as the section names you want to localize will always contain underscores. Treat `[section ` and `[prc_` as substitutable prefixes which have no bearing on the semantics of the section name. As I mentioned in the Core PR just now, both `[section gtr_solo]` and `[prc_gtr_solo]` should be localized identically.

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
```diff
@@ -101,5 +105,223 @@ public void UpdateView()
...
+                }
+            }
+
+            // Not a standard section name, so return unlocalized
+            return prcSectionName.Replace('_', ' ').Trim();
+        }
+
+        // Whether the string follows the convention for letter-based section names
+        // e.g., "a" for "A section", "b3" for "B section 3", or "z92b" for "Z section 92B"
+        private bool isLetterBasedSectionName(string text)
```
TheNathannator, line 161:
```suggestion
        private bool IsLetterBasedSectionName(string text)
```

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
```diff
@@ -101,5 +105,223 @@ public void UpdateView()
...
+            if (text.Length == 1)
+            {
+                return text[0].IsAsciiLetterLower();
+            }
+
+            return text[0].IsAsciiLetterLower() && text[1].IsAsciiDigit() && isSectionNumber(text[1..]);
+        }
+
+        // Whether the string follows the convention for section numbers (a, 1, 1a)
+        private bool isSectionNumber(string text)
```
TheNathannator, line 177:
```suggestion
        private bool IsSectionNumber(string text)
```

TheNathannator (commented):
Forgot about a couple more nitpicks I had lol, name casing of functions should generally be PascalCase regardless of visibility.

Assets/StreamingAssets/lang/en-US.json
TheNathannator: I don't think it's something deliberately done by us, yeah. Didn't realize that was already done for these files lol

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
Frickitickitavi: I suppose we'd want to move `LetterSection` and `WithNumber` out of the same grouping as the actual section flags, to avoid false positives if someone actually uses something like `[section WithNumber]` for some reason. `Gameplay.Practice.Sections` for the literal flags and `Gameplay.Practice.SectionFormats` for those two?

Assets/StreamingAssets/lang/en-US.json
Frickitickitavi: In light of that, does your comment still apply? If so, should I replace each value with `""` or `undefined`/`null`?

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
```diff
@@ -101,5 +105,223 @@ public void UpdateView()
...
+        }
+
+        // Receives a prc-based section name with "prc_" removed
+        private string ParsePrcSectionName(string prcSectionName)
+        {
+            // Handle letter-based sections like "A section" (prc_a) and "B section 3" (prc_b3)
+            if (isLetterBasedSectionName(prcSectionName))
+            {
+                var name = char.ToUpper(prcSectionName[0]) + " " + Localize.Key("Gameplay.PracticeSections.Section");
+                return prcSectionName.Length == 1 ? name : (name + " " + prcSectionName[1..]);
```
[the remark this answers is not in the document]
Frickitickitavi: Cosmetic change - going with `LetterSection` instead of `LetterSubsection` for the name, because (in my mind, at least) "subsection" refers to the letter that comes after a section number, like the "B" in "Guitar Solo 2B".

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
Frickitickitavi: Pushed a commit with this approach implemented - let me know what you think 🙂

Assets/StreamingAssets/lang/en-US.json
Frickitickitavi: Latest push still includes the populated values for now, pending this conversation

EliteAsian123: Crowdin deals with (or will deal with as we've disabled it for now) the non-English translation files, so only the `en-US.json` file should be modified in this case.

Frickitickitavi: Got it, will revert the other four files

Frickitickitavi:
Screwed up pulling latest changes like a big dummy, and I have some significant changes in mind anyway, so going to close and reopen fresh

PR #952, opened 2025-01-25
Add active graphics api to watermark text

theli-ua:
Makes it a little easier to debug user graphics issues when we need to see which graphics api is currently active


EliteAsian123 (changes requested):
Hi, thanks for the PR!

This should definitely be a nightly/test build only thing, and it shouldn't be displayed in stable builds.

Thanks!

theli-ua:
>and it shouldn't be displayed in stable builds.

I think there should be a way to see it, but I guess it is in the logs. I've changed it to non-stable

PR #953, opened 2025-01-27
Fix some issues with score database querying/caching, and allow selecting whether high-score display searches all difficulties or only the highest difficulty

TheNathannator:
- Highest score and percentage are now always queried separately from each other for the high score display.
- The high score display now has the option to choose between whether all difficulties are searched for the highest score/percentage, or whether only the highest difficulty is used.
  - This defaults to the highest difficulty only option, since, from a naive/ideal perspective, it provides a more natural way of showcasing progression up the difficulty ladder and accounts for most general cases.


EliteAsian123 (changes requested):
The only problem is see is if "High Score Display" in the settings is set to "Off". In nightly it looks like this:

image

However, in this PR it looks like this with "Highest across all difficulties" (which I'm assuming isn't intended).

image

Amazing work nonetheless!

TheNathannator:
The one high score mode I didn't check lol, thanks for catching that!

PR #957, opened 2025-02-02
Fix venues shaders on OSX (again)

theli-ua:
Previously we were exporting shaders for OSX separately. Unfortunately that does not guarantee that all shader variants were exported. To mitigate that we now export materials instead. However to avoid doubling of resulting yarground size we replace all textures with dummy on export.
On load we fixup all the shaders to pick up from osx-specific subbundle

### DIFF
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -6,8 +6,8 @@ using Cysharp.Threading.Tasks;
 using UnityEngine;
 using UnityEngine.UI;
 using UnityEngine.Video;
-using YARG.Core.Extensions;
 using YARG.Core.IO;
+using YARG.Core.Logging;
 using YARG.Core.Venue;
 using YARG.Helpers.Extensions;
 using YARG.Settings;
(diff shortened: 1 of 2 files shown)


EliteAsian123 (changes requested):
Hello, thanks for the PR!

Unfortunately, this is beginning to look like more of a hack than an actual fix to the cross-platform venue problem, which we would like to avoid. You mention that not all shader variants were exported. Do you think there's a more elegant way to get the shader variants without looking through all of the materials, removing the textures, and then exporting those?

Thanks!

Assets/Script/Gameplay/BackgroundManager.cs
```diff
@@ -91,7 +91,7 @@ private async UniTaskVoid Start()
                     // Yarground comes with shaders for dx11/dx12/glcore/vulkan
                     // Metal shaders used on OSX come in this separate bundle
                     // Update our renderers to use them
-                    var renderers = bg.GetComponentsInChildren<Renderer>();
```
theli-ua, line 94: This is a fix that I found while testing. I've seen venues that have an object inactive in the beginning but then show it later. Passing true here will ensure we fixup all objects regardless if they are active in hierarchy or not

theli-ua:
It looks like I may have broken something in latest update here and it still wouldn't work with all the shaders on the mac. I'll switch to draft for now and reopen when confirming

theli-ua:
Fixed and retested, exported bundles work with the current nightly (as they mostly should)

PR #958, opened 2025-02-03
Allow vocal note lines' draw calls to be batched

theli-ua:
Current code code directly sets `.color` on the material which creates a separate instance of material for every mesh renderer and prevents srp batcher from combining them.

Tested in a frame debugger. Before the change every line is drawn as a separate "draw" event, after (with single vocal player) in  a single "Dynamic Draw" one.


theli-ua:
This doesn't change much when there are multiple vocal players as the way rendering order is done in unity it won't have a lot of chances combining them. Could be improved later

Assets/Script/Gameplay/Visuals/VocalElements/VocalNoteElement.cs
```diff
@@ -25,11 +25,13 @@ public class VocalNoteElement : VocalElement
         protected override void InitializeElement()
         {
             var color = Player.VocalTrack.Colors[NoteRef.HarmonyPart];
+            var matPropBlock = new MaterialPropertyBlock();
```
TheNathannator, line 28: This should be cached into a field to avoid repeated allocation, MaterialPropertyBlock allocates a native object and doesn't free it until it gets finalized by the garbage collector.

TheNathannator (changes requested):

Assets/Script/Gameplay/Visuals/VocalElements/VocalNoteElement.cs
theli-ua: :+1:

TheNathannator (approved):
Looks good code-wise now. I could nitpick how the property block initialization is done (IIRC it can just be `_materialProperties = new()`), but that's just me being a little OCD lol, I'll leave that decision up to you.

PR #959, opened 2025-02-03
Add support for shift indicators for five fret instruments

wyrdough:
This PR adds support for shift indicators on five fret instruments, using range shift text events found in some of the YARG setlist songs.

During initialization the list of text events attached to the note track is scanned for events with the form `[ld_range_shift d r s]` where d is the difficulty, r is the 1 indexed range, and s is the number of frets in the range. Any frets not in the initial range are grayed out.

inactive fret

When a range shift event occurs, range shift markers are shown at the four preceding beat lines on the track and the currently inactive frets coming into use flash at each range shift marker.

Shift marker, currently identical to the pro keys range shift marker:
shift marker

Fading out from the beat line flash:
mid flash

Once the event has passed, the now active frets are colored and the now inactive frets fade to gray.

inactive flipped


wyrdough:
Closing because this has been superseded by #1007.

PR #962, opened 2025-02-04
Show warning when player has no input device or microphone

thomeval:
This PR adds a warning message to the Difficulty Select Screen when a non-bot player has no input devices assigned:
image

 For vocalists, having no input device is considered valid. Instead, a different warning is displayed when a vocalist has no microphone connected:

image

Players without input devices are also shown differently on the Active Players list in the top status bar:

image

<s>Note that icons used for this are placeholders and are identical to the regular instrument icons, except for the Vocals one.</s> This PR now includes Kadu's new instrument icons.


kaduwaengertner:
We recently recreated all the instrument icons so they all match in style, so if you can update your PR to utilize the new ones that would be awesome, also sending the regular icons to be updated.

NoInstrumentIcons
InstrumentIcons

thomeval:
@kaduwaengertner Great, I've included those into this PR.

EliteAsian123 (changes requested):
Hello again!

The code itself is fine, however, the anchoring positions of the UI elements are incorrect on certain resolutions:

image

Please see the "Loading Phrase" element in the same prefab to see how it can be fixed in this scenario. Thanks!

thomeval:
Hi, I've pushed an update to the branch that fixes the positioning of the warning message. I tested it on 4:3, 16:9, 16:10 and 21:9 and all of them work without issue now.

PR #963, opened 2025-02-04
Video calibration

dornED:
This is the first half/part of https://yarg.youtrack.cloud/issue/YARG-187/Rework-calibration-tool-to-allow-calibrating-video-and-input-delay i.e. to add video calibration. I intend to improve the calibrator further, this is just the "dipping my toes in game development" PR. For the further improvements, I'll open a proposal on Discord first.

As for the changes here, they are fairly simplistic, mainly copy/pasting what has been done for audio calibration. The latency calculation (median) is also reused. I used invoke instead of a coroutine because apparently it is more accurate (https://www.tantzygames.com/blog/most-accurate-timer-in-unity/)

### DIFF
--- a/Assets/Scenes/CalibrationScene.unity
+++ b/Assets/Scenes/CalibrationScene.unity
@@ -123,6 +123,142 @@ NavMeshSettings:
     debug:
       m_Flags: 0
   m_NavMeshData: {fileID: 0}
+--- !u!1 &51138883
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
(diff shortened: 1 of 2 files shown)


sonicfind:
Sorry for the absolutely massive delay on this.

First immediate thing that pops out at me: disconnecting the video calibration from actual in-game visual won't really work. The part of video calibration that matters the most is how it ties in to actual gameplay, and a circle popping on screen doesn't really achieve that effect.

Additionally, having it be a circle that disappears and reappears makes calibration inconsistent. Video calibration requires the player to be able to visible see the rhythm as it approaches the point when they have to supply an input. If they can't see when they need to press a button, it becomes more of a game of reaction speed than a calibration test.

A good example is CH's tool, where it uses a separate scene with a special highway for calibration. It'd be good to aim for something like that.

EliteAsian123:
Yes, I agree with sonicfind here, so unfortunately, I'm going to close the PR. If you ever make a new revision of this PR, then feel free to open a new one any time!

PR #964, opened 2025-02-06
Use GPU instancing for notes

theli-ua:
Posting as a draft to get early feedback.
As of posting GPU instancing will only be used for rectangular normal note.
It'll be one instanced draw per (mesh, material) tuple basically. So if only normal rectangular notes are rendered it'll be 3 total for them, like this (frame debugger)

image


Assets/Script/Gameplay/Visuals/MaterialPropertyInstance.cs
```diff
@@ -0,0 +1,11 @@
+using UnityEngine;
+
+namespace YARG.Gameplay.Visuals
+{
+
+    public static class MaterialPropertyInstance
+    {
+        private static MaterialPropertyBlock _instance;
+        public static MaterialPropertyBlock Instance => _instance ??= new MaterialPropertyBlock();
```
sonicfind, line 9: Would it not be better suited to just have `public static readonly MaterialPropertyBlock Instance = new();`?

theli-ua: This part is actually from a "batch vocal notes rendering pr", this draft is just on top of that. And that piece of code is really just a copy-paste of code that Nate suggested

TheNathannator: I suggested the delayed initialization because I'm not certain whether MaterialPropertyBlock is thread-safe to construct, nor whether the static initializer here will be run on a separate thread or simply on first access.

RileyTheFox:
I am assuming that https://github.com/YARC-Official/YARG/pull/1046 is the grand successor to this. If not feel free to reopen/make a new PR with this specific change in the future

PR #966, opened 2025-02-07
Replace Native STBImage Plugins With Custom-Tuned STBImageSharp

sonicfind:
Aimed to address images not loading on certain platforms without forcing a user to manually install the required c++ dist


sonicfind:
Unfortunately, after testing, I've come to find that every image seems to incur a pretty substantial performance regression in comparison to before.

Now, on my own machine, I can't notice it in-game to a substantial degree, but that's just because my rig is JACKED. We can already see reasonable delay in image loads as it is on other machines (although that may be Unity burst, can't tell), so I'd rather not take the chance right now.

Maybe when Unity gets off its stupid-ass mono implementation.

I'll do some more testing, but for now, I'm putting this on the backburner.

PR #967, opened 2025-02-07
Cache combo meter's text meshes

theli-ua:
Currently text onto meter is set every frame and every frame it (re) generates the mesh. Its not that expensive but on my low power old mac it adds up to 1ms with 4 players

Pre-generate all meter meshes since there is only a few and display the needed one.

I suspect my changes in TrackPlayer classes are wrong so hopefully someone will give a nudge in the right direction for that part.

### DIFF
--- a/Assets/Script/Gameplay/Visuals/ComboMeter.cs
+++ b/Assets/Script/Gameplay/Visuals/ComboMeter.cs
@@ -25,8 +25,20 @@ namespace YARG.Gameplay.Visuals
         [SerializeField]
         private Material _noFcRingMaterial;
 
-        public void Initialize(EnginePreset preset)
+        private TextMeshPro[] _textCache;
+
+        public void Initialize(EnginePreset preset, int maxMultiplier)
         {
+            _multiplierText.enabled = false;
+            _multiplierText.text = string.Empty;
+            _textCache = new TextMeshPro[maxMultiplier * 2 - 1];
+            _textCache[0] = _multiplierText;
+            for(int i = 0; i < _textCache.Length; ++i)
+            {
+                _textCache[i] = Instantiate(_multiplierText, _multiplierText.transform.parent, true);
+                _textCache[i].SetTextFormat("{0}<sub>x</sub>", i + 2);
+            }
+
             // Skip if the preset is a default one
             if (EnginePreset.Defaults.Contains(preset)) return;
 
@@ -36,10 +48,12 @@ namespace YARG.Gameplay.Visuals
 
         public void SetCombo(int multiplier, int maxMultiplier, int combo)
         {
-            if (multiplier != 1)
-                _multiplierText.SetTextFormat("{0}<sub>x</sub>", multiplier);
-            else
-                _multiplierText.text = string.Empty;
+            _multiplierText.enabled = false;
+            if (multiplier > 1)
+            {
+                _multiplierText = _textCache[multiplier - 2];
+                _multiplierText.enabled = true;
+            }
 
             int index = combo % 10;
             if (combo != 0 && index == 0)
@@ -56,7 +70,7 @@ namespace YARG.Gameplay.Visuals
 
         public void SetFullCombo(bool isFc)
         {
-            _ringMesh.material = isFc ? _fcRingMaterial : _noFcRingMaterial;
+            _ringMesh.sharedMaterial = isFc ? _fcRingMaterial : _noFcRingMaterial;
         }
     }
 }
--- a/Assets/Script/Gameplay/Player/TrackPlayer.cs
+++ b/Assets/Script/Gameplay/Player/TrackPlayer.cs
@@ -88,7 +88,6 @@ namespace YARG.Gameplay.Player
 
             var preset = player.EnginePreset;
             IndicatorStripes.Initialize(preset);
-            ComboMeter.Initialize(preset);
 
             // Set fade information and highway length
             ZeroFadePosition = DEFAULT_ZERO_FADE_POS * Player.Profile.HighwayLength;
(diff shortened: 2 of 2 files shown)


Assets/Script/Gameplay/Visuals/ComboMeter.cs
```diff
@@ -25,8 +26,27 @@ public class ComboMeter : MonoBehaviour
...
+                {
+                    _multiplierText.SetTextFormat("{0}<sub>x</sub>", multiplier + 2);
+                }
+                else
+                {
+                    _multiplierText.text = string.Empty;
+                }
+                return _multiplierText;
+            })
+            .ToArray();
```
RileyTheFox, line 47: Try to avoid using LINQ especially in gameplay code. While this does only run once, this will allocate quite a lot and will eventually be garbage collected during gameplay which could cause a stutter.

Try to pre-calculate the required number of items for the array and allocate the array up front then assign by index instead of doing LINQ Enumerable stuff then ToArray (which is another allocation)

RileyTheFox (changes requested):

Assets/Script/Gameplay/Visuals/ComboMeter.cs
theli-ua: :+1:

Assets/Script/Gameplay/Visuals/ComboMeter.cs
```diff
@@ -25,8 +25,20 @@ public class ComboMeter : MonoBehaviour
...
+            _multiplierText.gameObject.SetActive(false);
+            _multiplierText.text = string.Empty;
+            _textCache = new TextMeshPro[maxMultiplier * 2 - 2];
+            _textCache[0] = _multiplierText;
+            for(int i = 0; i < _textCache.Length; ++i)
+            {
+                _textCache[i] = Instantiate(_multiplierText, _multiplierText.transform.parent, true);
+                _textCache[i].SetTextFormat("{0}<sub>x</sub>", i + 2);
+            }
+
```
sonicfind, line 41: Instead of cloning the entire unity object, you could instead just hold an array of all the strings. Then, just call `SetText` on the object to assign the string.

Assets/Script/Gameplay/Visuals/ComboMeter.cs
sonicfind, line 1: If we don't want duplicate actions to incur all the time, we should track in this object the current combo. Then on subsequent calls, compare the combo value passed to the `SetCombo` function and determine whether we need to update the multiplier display.

```
if (combo > _maxCombo)
{
    combo = _maxCombo;
}

if (_comboTracker != combo)
{
    // Update
}
_comboTracker = combo;
```
As an example

Assets/Script/Gameplay/Visuals/ComboMeter.cs
```diff
@@ -56,7 +70,7 @@ public void SetCombo(int multiplier, int maxMultiplier, int combo)
 
         public void SetFullCombo(bool isFc)
         {
-            _ringMesh.material = isFc ? _fcRingMaterial : _noFcRingMaterial;
+            _ringMesh.sharedMaterial = isFc ? _fcRingMaterial : _noFcRingMaterial;
```
sonicfind, line 73: To accompany the "array of strings" comment, this would go unchanged.

Assets/Script/Gameplay/Visuals/ComboMeter.cs
sonicfind: We only used SetTextFormat because the numerical value for the multiplier was always supplied directly to the object. However, if we cache all the strings beforehand, then the formatting for the number already happened, so `SetText` would work fine.

You'd just have to change the string to `{i + 2}<sub>x</sub>"`.

sonicfind: Also, in doing so, we could cache this array of strings saved statically for all instances of the object since we've predefined the max it could go to across the various instruments (that being "6x").

sonicfind (changes requested):

Assets/Script/Gameplay/Visuals/ComboMeter.cs
theli-ua: This has nothing to do with it being SetTextFormat, whichever you use, what this avoids is generating a mesh from a text every time either SetText or SetTextFormat is called

Assets/Script/Gameplay/Visuals/ComboMeter.cs
theli-ua: it would make sense if you would call SetText in the hot path, but with meshes already pre-generated this doesn't actually save anything

Assets/Script/Gameplay/Visuals/ComboMeter.cs
theli-ua: This is actually not related to rest changes. This simply avoids instantiating materials unnecessary here.

Assets/Script/Gameplay/Visuals/ComboMeter.cs
sonicfind: Ah, I see. Fair enough (which also means no static array, DAMN YOU UNITY).

sonicfind:
This breaks once you activate star power. 
{A15AE478-FB26-4340-BAC4-D55700C1372A}
Drums tries to get `8x`.

sonicfind:
But with how it's structured, you'd end up generating x5, & x7, which go completely unused.
And with bass, x7, x9, & x11.

sonicfind:
Granted, that's a separate issue really, stemming from the engines themselves, but it does complicate the current situation.

theli-ua:
> But with how it's structured, you'd end up generating x5, & x7, which go completely unused. And with bass, x7, x9, & x11.
That's fine, they're not that expensive to keep around and it does keep code simple.


> Drums tries to get 8x.

oops, fixed this off-by-one

PR #968, opened 2025-02-07
Fix Circular HOPO Note Flares not scaling with FOV (Approximate scaling math to target Circular preset/60°)

MaddieMewmews:
Currently, on both Nightly and Stable, HOPO note lens flare effects on the Circular note shape do not scale with the perceptual distance change that comes with a change of FOV. The current code for Note Flares only accounts for the *physical distance* to the camera object. This makes them tiny on low FOV, and notably, Michael Bay levels of flare on high FOV.

Attached are some screenshots to demonstrate the intended effect, a 120fov profile, and after the linear function patch is applied
### 60FOV Circular Preset, target appearance
60FOV Circular Preset, target appearance
### 120FOV profile, pre-patch
120FOV profile, pre-patch
### The same profile, after the patch
sameprofilepatch

The function might need some tweaking, but it does make higher FOV camera presets usable with the Circular note shape option.

### DIFF
--- a/Assets/Script/Gameplay/Visuals/TrackElements/NoteFlare.cs
+++ b/Assets/Script/Gameplay/Visuals/TrackElements/NoteFlare.cs
@@ -11,7 +11,6 @@ namespace YARG.Gameplay.Visuals
 
         [SerializeField]
         private float _scaleConstant = 4.25f;
-
         private LensFlareComponentSRP _flare;
 
         private float _fadeFullPosition;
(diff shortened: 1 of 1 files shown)


sonicfind:
I will let you take the credit (being the catalyst for this fix after all), but I will note a few things.

While it does work decently enough for higher FOVs, the result for lower FOVs doesn't really match. Low FOVs provide much too minimal of an effect in comparison to the base & higher FOVs. This basically stems from the solution you provided being based on a linear progression.

If you look up the most basic formula for calculating an FOV, you'll notice that it includes a tangent calculation. If you combine that with a plot for a best-fit graph in something like desmos, you'd find that the best matching formula for our needs involves a cotangent.

In my case, the best one I got to work was
```
float fovScale = (2.6f / Mathf.Tan(0.0126f * TrackPlayer.TrackCamera.fieldOfView - .046f)) + 1.223f;
_flare.scale = fovScale / Vector3.Distance(transform.position, TrackPlayer.TrackCamera.transform.position);
```

Feel free to test it out and/or modify it. If you want to use it, just update your code and we'll push that through.
You will notice though that it does away with the original `4.25` scale value. That's just a consequence.

MaddieMewmews:
Commit amended, thanks so much for the excellent proper FOV scaling code. The linear function was absolutely just a quick-fix shortcut on my end. Not sure if _scaleConstant is needed anymore, but I'd rather not break anything.

PR #973, opened 2025-02-15
Fix some text sometimes not visible in opengl mode

theli-ua:
By changing shader used to an `overlay` variant that will always render on top. The Distance Field and Distance Field Overlay shaders are two nearly-identical variants of the TextMesh Pro signed distance field (SDF)shader. The difference between the two is that the Distance Field Overlay variant always renders the TextMesh Pro object on top of everything else in the Scene, while the Distance Field variant renders the Scene normally—objects in front of the TextMesh Pro object are rendered on top of the text.

There is no case currently where we need non-overlay variant and apparently according to user reports it causes some text not being visible in certain situations. Likely due to some depth collision/z-fighting.


sonicfind:
Screw it, balls to the wall

PR #977, opened 2025-02-22
Render highway/notes as opaques

theli-ua:
This is in a state where I believe  everything is working as intended. Everything is rendered as opaques (mostly, other than hit flash/smoke effects obviously). And fade is applied afterwards as a post effect. Fade is applied as part of UICurve shader right now but could be applied separately. its just cheaper to do it right there.  Fade/start position is calculated by projecting highway's coordinates onto the screen viewpoint. As far as I can see it looks the same as current code.
Biggest benefit is that no highway-related  materials have to carry whole fading around and not care about it at all. 
Second is fixing weird issues like  kick note peeking through cymbals and whatnot. or the kick note being incorrect color when its in the fade  zone because it doesn't have tonemapping applied correctly to it


theli-ua:
This does fix some quirks like notes/kicknote blending weirdly where you see kick note through a cymbal/etc, or the kick note flashing red before it turns orange at the edge of fading

theli-ua:
This doesn't seem to have any effect on performance whatsoever. 
I do like the fact that it simplifies things in a sense that no materials on the highway need to care about fade anymore. its done centralized elsewhere

theli-ua:
Forgot to mention, but this also removes the hack to clear z-depth between rendering various parts. So this also should significantly reduce the overdraw when rendering highways

theli-ua:
This will also make fade a gradient within an object, as in - today object has the same opacity throughout as its calculated as object's distance from the strikeline. 
With this change it will be gradual - more distant parts will have less opacity (assuming they are in the fade zone obviously)

Dansla116:
From a QA/end-user perspective, everything appears to be working exactly as before with no visual regressions or behavioral changes that I can see.

That said, I see this PR has gone through a number of thoughtful revisions, and the current commit history reflects a well-contained, cleaner approach to rendering and fading.  I appreciate the potential performance and maintainability wins (like dropping the depth buffer clear hack, reducing overdraw, and fixing some long-standing quirks).

Looks solid from my end!

RileyTheFox:
For simplicity, can you re-add anything in this PR not currently in https://github.com/YARC-Official/YARG/pull/1046 to that PR instead as you said yourself that 1046 supersedes this

PR #979, opened 2025-02-23
Do not generate garbage every frame in scorebox update

theli-ua:
TextMesh.text getter generates garbage on each call


wyrdough (approved):
Playtesting shows that this does not break anything and it's presumably better to allocate a double off the stack that doesn't need to be GCed than a string on the heap that does, so I think this is good.

PR #980, opened 2025-02-24
Split most frequently updated elements into own ui canvas

theli-ua:
Whenever they update they trigger rebuilding whole UI layout so it is the best practice to isolate them.

### DIFF
--- a/Assets/Prefabs/Gameplay/HUD/ScoreDisplay.prefab
+++ b/Assets/Prefabs/Gameplay/HUD/ScoreDisplay.prefab
@@ -394,7 +394,6 @@ GameObject:
   m_Component:
   - component: {fileID: 1682100933276730677}
   - component: {fileID: 3580836438431718176}
-  - component: {fileID: 1427118048}
   m_Layer: 5
   m_Name: ScoreDisplay
   m_TagString: Untagged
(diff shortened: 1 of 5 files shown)


theli-ua:
This does improve performance a bit on my test low-end device

Dansla116:
I've been testing this PR against the current dev build, switching back and forth.  Even on a relatively powerful gaming PC, I can anecdotally confirm that performance feels improved (fewer FPS dips), at least from what I can catch out of the corner of my eye.  So it seems like the PR is doing what it's meant to.

That said, I did notice a change in lyric color that might be a regression.  After a lyric is played, the highlight color is noticeably more faded, which makes it harder to track your place peripherally.  Even if this was by design, I personally prefer the current dev build.

Lyric display before (dev build):
image

After (with this PR):
image


This only seems to affect the lyric bar. The gameplay view for vocals appears unchanged.
Vocal view before (dev build):
image

After (with this PR):
image

Dansla116:
I pulled the latest change and have been playing on this build.  I can confirm we've got the full benefit of the performance improvements and the lyric highlight now has that nice blue again instead of being faded.

Thanks for following up on it!

wyrdough (approved):
Looks good to me

PR #981, opened 2025-02-25
Fix fret flash/smoke making holes in the track

### LINKED ISSUE #790: [BUG]: note flames alpha layer transparency problem
I just noticed that the note flames are creating transparent effect and it's going tro the "alpha layer?" - maybe. 
I attached a picture. I can clearly see the cable and the table edge behind the flames.
image

theli-ua:
change to premultiply blending mode
fixes #790


theli-ua:
I've tested this by using a graph paper as background 

Before:
image


After:

image

theli-ua:
oops, forgot the meta for material. fixed now.  Added ParticlesUnlit material is just a full copy of one from urp package. Just so that we could change blending mode

TheNathannator (approved):
tested and confirmed. thank you!

Issue #860, opened 2024-08-29
[FR] Built in Web Server for Song Queing/Search from external device(tablet/phone etc)


aenertia:
Please add support to enable remote Library Search/Queue management from a Webpage whilst YARG is running. This would facilitate i.e kareoke / couch participation and events.

Nyxyxylyth:
Similar proposal on Discord:
https://discord.com/channels/1086048856678084609/1337991395503505439

PR #984, opened 2025-02-27
Add very basic ability to select song via http to music library

theli-ua:


theli-ua:
Very basic version of https://github.com/YARC-Official/YARG/issues/860

theli-ua:
To clarify: this allows to open "yarg_host:9090" , be presented with basic list of songs in the library, upon clicking on them they will become selected in yarg

theli-ua:
This is more of PoC to get some general feedback on the idea  I guess. If this is something that we want to actually do then probably will need more thought put into this

joewestcott:
Hey @theli-ua 👋 

I've just checked this out and it works really well! It's quite surreal to see the YARG UI change itself, and browsing the song list whilst not using the main YARG instance is very neat. I think this PR could lay the foundation for a YARG app, which would be an excellent addition to parties where everyone can view the song list via their own phone/tablet.

I'm sure this goes without saying (it's a PoC, after all!), but it's probably better for YARG itself to serve a JSON API over HTTP, rather than generating HTML within csharp. This would make the csharp side much easier to work with, and allow for a separate frontend team to pickup the HTML/JS/CSS side. Perhaps myself or the launcher contributors could then get involved with the UI for this project!

If we go down the route of separating the app/api, I can see two options of how to serve the app:
1. Bundle the app code along with YARG, and serve the app on a local port along with the JSON API. This is the most straightforward approach, but does require the user to be on the same wifi network, which is a small disadvantage of this.
2. Host the app online, and relay the API of each running YARG instance through a web server hosted by us. This would be a more seamless approach for users, but does involve maintaining a server.

To get this project started - perhaps we could go with option 1, and just bundle a html file along with YARG for now?

Questions for you:
- Do you think a JSON API is a good idea?
- Could we get more track information? Artist, album, year, instrument difficulties, and album art would be a great start.
- If possible - it would be really cool to show a QR-code of the URL on the YARG home screen. Something needs to go there - it's a bit empty at the moment!

Thanks :)

theli-ua:
All of those suggestions are great. Though I don't know if anyone is actually interested in working on this or of any plans to do anything in this area.

joewestcott:
I'm interested in getting a frontend put together! Do you think you could mould this PR into a backend JSON server? It's halfway there already. Otherwise I'll tackle both if it's likely to be accepted. :)

Essentially, `yarg_host:9090` would serve the HTML file from the frontend team, and `yarg_host:9090/api` would have a GET endpoint for a song list and a POST endpoint to change the selected song. Easy right?!

PR #986, opened 2025-03-01
Fix stutter caused by score star popping

theli-ua:
Activating object does it recursively and forces gui canvas rebuild

Before:
image

After:

image

### DIFF
--- a/Assets/Prefabs/Gameplay/HUD/ScoreDisplay.prefab
+++ b/Assets/Prefabs/Gameplay/HUD/ScoreDisplay.prefab
@@ -1574,6 +1574,21 @@ PrefabInstance:
       propertyPath: m_FillAmount
       value: 0
       objectReference: {fileID: 0}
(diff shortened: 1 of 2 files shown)


theli-ua:
what this does is make all stars active from the get go, but sets image to fill from simple, and sets fill to 0 for start, sets to 1 when we need to display

sonicfind:
Just wanna make sure: are *all* the prefab changes necessary for this to behave or just the one with the altered `m_IsActive` values?

theli-ua:
> Just wanna make sure: are _all_ the prefab changes necessary for this to behave or just the ones with the altered `m_IsActive` values?

Other than changing active state to true it changes type of the sprite from simple to fill, and sets an initial fill amount to 0 (so none of it is visible). Those are the other changes. 

They are required here. (well, unless you say this is all bs)

Basically this changes showing stars from active false->true, into fill amount 0->1

sonicfind:
sounds good to me

PR #987, opened 2025-03-01
Update Active Player List instrument icons on device connect/disconnect

thomeval:
This PR refreshes the active player (or more specifically the instrument icon) list whenever an input device is connected or disconnected. This fixes an issue where the list would display the "device not connected" red instrument icon for a player even after that player's device is connected after the game starts up.


sonicfind:
Sounds fair enough

PR #988, opened 2025-03-01
Use much more aggresive prewarming for track element pools

theli-ua:
When a pool is depleted creating new elements is expensive and total memory used is only couple mb anyway
Cost to create them is pretty significant (profiling on steam deck)
image


sonicfind:
@EliteAsian123 any complaints? seems like a harmless change

PR #990, opened 2025-03-10
Fix bug with score database high score query.

wyrdough:
Songs that have only been played on one difficulty currently always show the first score record when score display is set to "highest difficulty only". This PR fixes the issue by using a different SQL query.

The new query [provided by goulart](https://discord.com/channels/1086048856678084609/1086052453801283624/1333068169551810560) when the current query was originally implemented uses a subselect, but returns the correct values. 

On my score database, goulart's query is about 0.7ms slower averaged over 1000 repetitions of the high score query. This shouldn't be an issue since all songs are loaded at once and the results are cached.

Relevant bug reports are [here](https://discord.com/channels/1086048856678084609/1345827528501100637) and [here](https://discord.com/channels/1086048856678084609/1347039075420340297)

### DIFF
--- a/Assets/Script/Scores/ScoreDatabase.cs
+++ b/Assets/Script/Scores/ScoreDatabase.cs
@@ -219,18 +219,21 @@ namespace YARG.Scores
             bool highestDifficultyOnly
         )
         {
-            string query =
-                @"SELECT *, MAX(Score) FROM PlayerScores
-                INNER JOIN GameRecords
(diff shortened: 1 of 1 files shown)


wyrdough:
Also, just to address the concern sonicfind brought up about garbage when this was originally implemented: 

Though the query uses a subselect, it is still a single query and is passed to SQLite that way by the C# SQLite library, so it should create no more and no less garbage than any other single query that returns the same number of rows.

PR #991, opened 2025-03-11
Fix issues caused by material name collisions when exporting some venues

theli-ua:


theli-ua:
This basically fixes failures to export venues which uses multiple materials with the same name

wyrdough (approved):
Simple and does what it says on the tin. I'd merge if I could. ;)

PR #992, opened 2025-03-11
Fix early wait countdown obscuring player name display

wyrdough:
Currently, if there is a wait countdown active at the beginning of a song it will render over the player name display. This wouldn't be so terrible except that the instrument icon shows behind the transparent parts of the countdown "pie". Not great.

This PR reverses the order of the elements in the TrackView prefab so that the player name display renders over the wait countdown. Rather than just completely obscuring the wait countdown I thought it would be nice to have the ring meter show around the outside of the instrument icon, so the scale of the wait countdown canvas is increased to 1.7, which matches the size of the inner part of the wait countdown to the size of the instrument icon. This leaves the ring meter on the wait countdown showing around the instrument icon.

Before: 
before

After:
after

### DIFF
--- a/Assets/Prefabs/Gameplay/HUD/TrackView.prefab
+++ b/Assets/Prefabs/Gameplay/HUD/TrackView.prefab
@@ -204,7 +204,7 @@ RectTransform:
   m_GameObject: {fileID: 881258039788376651}
   m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
   m_LocalPosition: {x: 0, y: 0, z: 0}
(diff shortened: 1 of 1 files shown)


theli-ua (approved):

sonicfind:
Does this still happen? I can't seem to get that to proc

wyrdough:
Yes, this is from latest nightly. Not the most obvious example, but it's the one song I can think of off the top of my head that has an early wait countdown.


https://github.com/user-attachments/assets/7179fb0f-b8b2-4fac-a922-9ca67b98d2f0

sonicfind:
yea, I got what you meant a bit later

Issue #242, opened 2023-05-02
Start button for Drums


EmperorJacoba:
When mapping a drum controller, there should be an option for a "Start" bind. Would not be required when mapping, but would be a helpful option for those with an extra drum switch.

(For personal context, I use my hi-hat pedal to pause the game mid-song in Clone Hero and a similar option here would be useful)

TheNathannator:
I implemented this recently, it should be available in the next release. (I added it after the v0.10 beta was made so it's not currently in any released builds.)

EmperorJacoba:
Coolio!

PR #993, opened 2025-03-12
Limit delay before score screen at end of song

wyrdough:
There is currently an excessive delay before showing the score screen at the end of many songs. This is because there is a fixed two second delay after the end of the greater of the chart end time, the audio end time, or the end event time. In many cases this can lead to 5+ seconds of sitting in silence awaiting the score screen.

This PR changes the delay to be adaptive, with a minimum of two seconds wait after the last charted note for any instrument before showing the score screen, thus eliminating the excessive delay.

SongLength is set depending on whether an end event is charted (endTime is the time of the end event):
- With an end event SongLength = Max(lastNoteTime, endTime)
- With no end event SongLength = Max(lastNoteTime, audioLength)

The song ending again depends on whether an end event is charted:
- With an end event the song ends at the greater of lastNoteTime + SONG_END_DELAY and endTime
- With no end event the song ends at the greater of lastNoteTime + SONG_END_DELAY and audioLength

~~The only exception is if the end_events tag is set in song.ini and an end event exists. In that case, the end event is respected precisely with no delay, regardless of any other considerations.~~

Additionally, silence detection is run on the StemMixer output once lastNoteTime + SONG_END_DELAY has elapsed when there is no end event and SongLength is being controlled by audioLength. This is to avoid long waits for audio tracks with excessive trailing silence.

Requires [YARG.Core PR #242](https://github.com/YARC-Official/YARG.Core/pull/242)

### DIFF
--- a/Assets/Script/Gameplay/GameManager.Loading.cs
+++ b/Assets/Script/Gameplay/GameManager.Loading.cs
@@ -286,32 +286,33 @@ namespace YARG.Gameplay
             _chartLoaded?.Invoke(Chart);
 
             double audioLength = _mixer.Length;
-            double chartLength = Chart.GetEndTime();
             double endTime = Chart.GetEndEvent()?.Time ?? -1;
-            double lastNoteEndTime = Chart.GetLastNoteEndTime();
+            _lastNoteEndTime = Chart.GetLastNoteEndTime();
 
             // The time we want to end the song depends on whether or not there is an end event.
-            // If there is, then Max(lostNoteEndTime, endTime) (in case the end event is bogus)
-            // If not, Max(chartLength, audioLength)
+            // If there is, then Max(lastNoteEndTime, endTime) (in case the end event is bogus)
+            // If not, Max(lastNoteEndTime, audioLength)
             // Then we add a delay to ensure that we don't go to the score screen until
             // SongLength plus SONG_END_DELAY
-            if (endTime > -1)
+
+            // Disregard the end event if it comes after the end of the audio track
+            if (endTime > -1 && endTime <= audioLength)
             {
-                SongLength = Math.Max(lastNoteEndTime, endTime);
-                _realEndDelay = Math.Max(lastNoteEndTime + SONG_END_DELAY, endTime) - SongLength;
+                SongLength = Math.Max(_lastNoteEndTime, endTime);
+                SongLength += Math.Max(_lastNoteEndTime + SONG_END_DELAY, endTime) - SongLength;
             }
             else
             {
-                SongLength = Math.Max(chartLength, audioLength);
-                _realEndDelay = Math.Max(lastNoteEndTime + SONG_END_DELAY, SongLength) - SongLength;
-            }
-
-            // If the end_events ini flag is set and the end event actually exists, ignore all previous instructions,
-            // respect the charter's intention, and do exactly what they say even if it results in stupid
-            if (Song.EndEvents && endTime > -1)
-            {
-                SongLength = endTime;
-                _realEndDelay = 0;
+                if (audioLength > _lastNoteEndTime)
+                {
+                    SongLength = audioLength;
+                    _endIsAudioEnd = true;
+                }
+                else
+                {
+                    SongLength = _lastNoteEndTime;
+                }
+                SongLength += Math.Max(_lastNoteEndTime + SONG_END_DELAY, SongLength) - SongLength;
             }
 
             _songLoaded?.Invoke();
--- a/Assets/Script/Gameplay/GameManager.Loading.cs
+++ b/Assets/Script/Gameplay/GameManager.Loading.cs
@@ -288,23 +288,32 @@ namespace YARG.Gameplay
             double audioLength = _mixer.Length;
             double chartLength = Chart.GetEndTime();
             double endTime = Chart.GetEndEvent()?.Time ?? -1;
+            double lastNoteEndTime = Chart.GetLastNoteEndTime();
 
-            // - Chart < Audio < [end] -> Audio
-            // - Chart < [end] < Audio -> [end]
-            // - [end] < Chart < Audio -> Audio
-            // - Audio < Chart         -> Chart
-            if (audioLength <= chartLength)
+            // The time we want to end the song depends on whether or not there is an end event.
+            // If there is, then Max(lostNoteEndTime, endTime) (in case the end event is bogus)
+            // If not, Max(chartLength, audioLength)
+            // Then we add a delay to ensure that we don't go to the score screen until
+            // SongLength plus SONG_END_DELAY
+            if (endTime > -1)
             {
-                SongLength = chartLength;
-            }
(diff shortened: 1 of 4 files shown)


Assets/Script/Gameplay/GameManager.Loading.cs
```diff
@@ -288,23 +288,32 @@ private void FinalizeChart()
...
+            else
             {
-                SongLength = audioLength;
+                SongLength = Math.Max(chartLength, audioLength);
+                _realEndDelay = Math.Max(lastNoteEndTime + SONG_END_DELAY, SongLength) - SongLength;
             }
-            else
+
+            // If the end_events ini flag is set and the end event actually exists, ignore all previous instructions,
+            // respect the charter's intention, and do exactly what they say even if it results in stupid
```
TheNathannator, line 310: > The only exception is if the end_events tag is set in song.ini

The primary purpose of `end_events` is to *disable* end events, not to force them to be used unconditionally. YARG doesn't respect it because we already account for the problems that tag tries to address (charts that have end events placed before the last note) in a way that doesn't require modification of charts.

Assets/Script/Gameplay/GameManager.cs
```diff
@@ -335,7 +338,7 @@ public void ResumeCore()
             }
 
             _pauseMenu.PopAllMenus();
-            if (_songRunner.SongTime >= SongLength + SONG_END_DELAY)
+            if (_songRunner.SongTime >= SongLength + _realEndDelay)
```
TheNathannator, line 341: As I think about it now, it would probably be best to just bake the end delay into the song length. It creates the least amount of surprises, and fixes the song progress timer stopping two seconds before the song actually ends.

Assets/Script/Gameplay/GameManager.Loading.cs
```diff
@@ -288,23 +288,32 @@ private void FinalizeChart()
             double audioLength = _mixer.Length;
             double chartLength = Chart.GetEndTime();
             double endTime = Chart.GetEndEvent()?.Time ?? -1;
-
-            // - Chart < Audio < [end] -> Audio
-            // - Chart < [end] < Audio -> [end]
-            // - [end] < Chart < Audio -> Audio
-            // - Audio < Chart         -> Chart
-            if (audioLength <= chartLength)
+            double lastNoteEndTime = Chart.GetLastNoteEndTime();
```
TheNathannator, line 291: Perhaps it was short-sighted of me to say it should respect the time of the last note lol, there might be other important events within the chart that should be respected as well (though with global events being ignored in `GetEndTime`, I suppose that point doesn't hold too much water). Having refreshed myself of the context of this function, I don't see a need to specifically track a new time instead of using the existing `chartLength` value; that was the value I had in mind when I said "the last note".

TheNathannator (changes requested):

Assets/Script/Gameplay/GameManager.Loading.cs
wyrdough: The way it was explained to me is that end_events being set is to say "no, I really mean it, the end event is where it's supposed to be".

```Then, when CH added it back in, we had this awful problem where people were using templates with the [end] event already in them at timestamp 120.1.000 or something 
And if they didn't move or remove that, then customs would just randomly end at that time
hence the inclusion of this new metadata field

Assets/Script/Gameplay/GameManager.Loading.cs
wyrdough: It isn't clear to me that we should actually care about text events or phrases after a (valid) end event?

Certainly in its original meaning, the end event was intended to be a go directly to jail, do not pass go, do not collect $100 kind of thing. Holding off a bit in the case where the end event is a bit earlier than is desirable from a user experience standpoint makes sense to me, but possibly delaying due to other chart events does not.

Assets/Script/Gameplay/GameManager.Loading.cs
TheNathannator: That's not what they were added for in CH whatsoever, they were added simply to make it possible to disable them in bad charts that don't use them properly. CH just blindly accepts the end event with no checks to whatever else is happening in the chart, the devs never did anything more than adding the tag to make it possible to account for this.

Assets/Script/Gameplay/GameManager.Loading.cs
TheNathannator: I'm probably just being overzealous on future-proofing with my concerns here lol, the point is from a consistency standpoint I would prefer for `GetEndTime` to be used (its behavior can be modified to exclude text events; there's not much reason we should be storing raw text events in the chart data anyways in retrospect). But if you're adamant on this, make sure you do it in the `else` case as well (or document why specifically it's not used there), as currently it's using `lastNoteEndTime` in one but `chartLength` in the other.

wyrdough:
During testing I ran into a bunch of charts with no end event and several seconds of trailing silence (in one case over 10 seconds!). This greatly annoyed me, so I implemented some rudimentary silence detection in GameManager.Audio that gets triggered beginning two seconds after the last note that will end the song before SongLength if a third of a second of continuous silence is detected. 

I'm not wedded to the threshold value being used or the third of a second duration, so suggestions are more than welcome.

Dansla116:
I've been doing a fair bit of testing across a wide range of charts (mostly legacy rips), and I wanted to share some results that highlight how this PR behaves in the wild.  Out of everything I tested, these four RBN songs stood out as good examples:

**Buckner and Garcia – *"The Defender"***
Ends with ~2 seconds of silence remaining.  From what I can tell, this change kicks in right around the natural end anyway, so nothing feels different.

**Cetan Clawson and the Soul Side – *"White Heat"***
Has a short silence after the last note, and it transitions cleanly to the score screen at ~2:15/2:16.  This one feels like a perfect example of the intended behavior.

**Fallen Angel – *"The One Who Walks Alone"*** and **Askari Nari - *"Mercury Blooming"***
The former has a few seconds of wind ambiance after the last note that gets cut short.  It's noticeable, but not necessarily a problem.  The latter has ~10 seconds of beach/wave ambiance that also gets cut off, but it feels more abrupt and jarring.

All that in mind, I can definitely see this being a subjective issue.  Some folks in the community - especially the more competitive "play and FC as many songs as possible" crowd - will probably appreciate the faster transitions and reduced downtime.  But others - particularly more casual or immersion-focused players - might miss the extra ambiance and breathing room at the end of a track.

Even so, I think this PR is a great solution for the vast number of charts without end events.  Since there's no practical way to manually "fix" all of them, silence detection makes sense as a fallback and generally improves the flow of most songs.

If there's ever a way to make this behavior configurable (maybe a menu toggle), that might help satisfy both camps.  But that's likely beyond the scope of what this PR set out to do.  As it stands, this is a solid QoL improvement.

Ape:
Idea: Move to the score screen soon after the last note, but keep playing the song until the end in that scree so to music is not cut out ever.

PR #994, opened 2025-03-15
Update BassStemMixer to support YARC-Official/YARG.Core#243

theli-ua:


sonicfind:
Waiting for the core submodule update commit

PR #995, opened 2025-03-16
[Visualizer venue support / Sound texture] Introduce Gameplay.TextureManager

theli-ua:
This serves for integrating game state related data into venue materials. Currently supported textures are:
 * Sound spectrum and waveform
 * Chart source icon
 * Chart album cover

### DIFF
--- a/Assets/Script/Gameplay/TextureManager.cs
+++ b/Assets/Script/Gameplay/TextureManager.cs
@@ -1,0 +1,110 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Cysharp.Threading.Tasks;
+using UnityEngine;
+using UnityEngine.UI;
+using YARG.Helpers.Extensions;
+using YARG.Song;
+
+namespace YARG.Gameplay
+{
+    /// <summary>
+    /// A point of integration between GamePlay state
+    /// and venue materials
+    /// </summary>
+    public class TextureManager : GameplayBehaviour
+    {
+        private Texture2D _sourceIcon;
+        private Texture2D _albumCover = null;
+        private Texture2D _soundTexture = null;
+        private float[] fft = new float[1024];
+        private float[] wave = new float[1024];
+
+        private static int _soundTexId = Shader.PropertyToID("_Yarg_SoundTex");
+        private static int _sourceIconId = Shader.PropertyToID("_Yarg_SourceIcon");
+        private static int _albumCoverId = Shader.PropertyToID("_Yarg_AlbumCover");
+
+
+        private void Start()
+        {
+            _sourceIcon = SongSources.SourceToIcon(GameManager.Song.Source).texture;
+        }
+
+        protected Texture2D getAlbumArt()
+        {
+            if (_albumCover == null)
+            {
+                var image = GameManager.Song.LoadAlbumData();
+                _albumCover = image.LoadTexture(false);
+            }
+            return _albumCover;
+        }
+
+        protected Texture2D getSoundTexture()
+        {
+            if (_soundTexture == null)
+            {
+                _soundTexture = new Texture2D(512, 2, TextureFormat.R8, false);
+            }
+            return _soundTexture;
+        }
+
+        public void processMaterial(Material m)
+        {
+            if (m.HasTexture(_sourceIconId))
+            {
+                m.SetTexture(_sourceIconId, _sourceIcon);
+            }
+            if (m.HasTexture(_soundTexId))
+            {
+                m.SetTexture(_soundTexId, getSoundTexture());
+            }
+            if (m.HasTexture(_albumCoverId))
+            {
+                m.SetTexture(_albumCoverId, getAlbumArt());
+            }
+        }
+
+        public async void FixedUpdate()
+        {
+            if (_soundTexture != null)
+            {
+                var pd = _soundTexture.GetPixelData<Byte>(0);
+
+                await UniTask.RunOnThreadPool(() =>
+                {
+                    GameManager.GetMixerFFTData(fft, 9 /* 512 */, false);
+                    GameManager.GetMixerSampleData(wave);
+
+                    double minDb = -100.0;
+                    double maxDb = -30.0;
+                    double range = maxDb - minDb;
+
+
+                    for (int i = 0; i < 512; ++i)
+                    {
+                        var fft_value = fft[i];
+                        // Avoid 0
+                        double magnitude = fft_value + 1e-20;
+                        // logarithmic scale
+                        double db = 20.0 * Math.Log10(magnitude);
+                        // clamp to range
+                        db = Math.Max(minDb, Math.Min(db, maxDb));
+                        // normalize
+                        double normalized = ((db - minDb) / range) * 255;
+
+                        // set spectrum data in the first row
+                        pd[i] = (byte) Math.Round(normalized);
+                        // waveform data in the second row
+                        pd[512 + i] = (byte) (255.0f * (1.0f + wave[i]) / 2.0f);
+
+                    }
+                });
+
+                _soundTexture.Apply(false, false);
+
+            }
+        }
+    }
+}
--- a/Assets/Script/Gameplay/TextureManager.cs
+++ b/Assets/Script/Gameplay/TextureManager.cs
@@ -132,10 +132,10 @@ namespace YARG.Gameplay
 
             for (int i = 0; i < FFT_TEXTURE_WIDTH; ++i)
             {
-                var fft_value = _fft[i] * (1.0f - fftSmoothingFactor) + _prevFft[i] * fftSmoothingFactor;
+                // var fft_value = _fft[i] * (1.0f - fftSmoothingFactor) + _prevFft[i] * fftSmoothingFactor;
                 _prevFft[i] = _fft[i];
                 // Avoid 0
-                double magnitude = fft_value + 1e-20;
+                double magnitude = _fft[i] + 1e-20;
                 // logarithmic scale
                 double db = 20.0 * Math.Log10(magnitude);
                 // clamp to range
--- a/Assets/Script/Gameplay/TextureManager.cs
+++ b/Assets/Script/Gameplay/TextureManager.cs
@@ -33,6 +33,8 @@ namespace YARG.Gameplay
         private static int _soundTexId = Shader.PropertyToID("_Yarg_SoundTex");
         private static int _sourceIconId = Shader.PropertyToID("_Yarg_SourceIcon");
         private static int _albumCoverId = Shader.PropertyToID("_Yarg_AlbumCover");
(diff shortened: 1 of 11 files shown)


theli-ua:
I'm not fully happy as how the sound spectrum looks like and I do intend to update it at some point but that won't change any publicly facing APIs

wyrdough (commented):
Be careful what you wish for, because the genie thought it sounded like you were wishing for nitpicks.

theli-ua:
[asd.zip](https://github.com/user-attachments/files/19723265/asd.zip)
Here's sample visualizer venue , its in a zip because otherwise github won't let me attach it

theli-ua:
[aa.zip](https://github.com/user-attachments/files/19723488/aa.zip)
Here is sample venue that uses album/cover art

theli-ua:
And here is test of how those 2 are supposed to look https://youtu.be/xk_ueNFsQLs

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -0,0 +1,110 @@
...
+                        pd[i] = (byte) Math.Round(normalized);
+                        // waveform data in the second row
+                        pd[512 + i] = (byte) (255.0f * (1.0f + wave[i]) / 2.0f);
+
+                    }
+                });
+
+                _soundTexture.Apply(false, false);
+
+            }
```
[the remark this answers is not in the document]
theli-ua: I don't like inverting it in `FixedUpdate` specifically as there may be more there later. But I'll factor it out into `UpdateFFT` and invert it there

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -0,0 +1,111 @@
...
+            if (m.HasTexture(_albumCoverId))
+            {
+                m.SetTexture(_albumCoverId, getAlbumArt());
+            }
+        }
+
+        private async void UpdateFFT()
+        {
+            if (_soundTexture == null)
+            { return; }
```
wyrdough, line 76: Super nitpicky, but this doesn't quite hit the style guidelines as is.

```suggestion
            if (_soundTexture == null)
            { 
                return; 
            }
```

wyrdough (approved):
Other than the one tiny nitpick, this looks good to me.

And just for the public record, I've been dogfooding this for a while now, and it works great. I'm super excited for the wide array of audio reactive shader-based venues this enables. I just need to get better at translating GLSL to HLSL so my success rate at porting them is better than 25% ;)

Seriously, thanks a ton for your work on this!

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -0,0 +1,125 @@
...
+        private float[] _prevWave = new float[FFT_SIZE * 2];
+
+        private static int _soundTexId = Shader.PropertyToID("_Yarg_SoundTex");
+        private static int _sourceIconId = Shader.PropertyToID("_Yarg_SourceIcon");
+        private static int _albumCoverId = Shader.PropertyToID("_Yarg_AlbumCover");
+
+        private const double MIN_DB = -100.0;
+        private const double MAX_DB = -30.0;
+        private const double DB_RANGE = MAX_DB - MIN_DB;
+        private const int FFT_SIZE_LOG = 9 /* aka log2(512) */;
```
wyrdough, line 37: Firefox uses a default FFT size of 2048. If we want to ease the porting of ShaderToy shaders, this needs to change.

```suggestion
        private const int FFT_SIZE_LOG = 11 /* aka log2(2048) */;
```

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -0,0 +1,125 @@
...
+                return; 
+            }
+            var pd = _soundTexture.GetPixelData<Byte>(0);
+
+            await UniTask.RunOnThreadPool(() =>
+            {
+                GameManager.GetMixerFFTData(_fft, FFT_SIZE_LOG, false);
+                GameManager.GetMixerSampleData(_wave);
+
+                for (int i = 0; i < FFT_SIZE; ++i)
```
wyrdough, line 95: I believe the ShaderToy JS actually only uses the first 512 bins of the 1024 returned for an FFT2048

```suggestion
                for (int i = 0; i < FFT_SIZE / 2; ++i)
```

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -0,0 +1,125 @@
...
+        // Smoothing factor (adjust to taste)
+        [Range(0.0f, 0.95f)]
+        public float fftSmoothingFactor = 0.8f;
+        [Range(0.0f, 0.95f)]
+        public float waveSmoothingFactor = 0.5f;
+
+        private Texture2D _sourceIcon;
+        private Texture2D _albumCover = null;
+        private Texture2D _soundTexture = null;
+        private float[] _fft = new float[FFT_SIZE * 2];
```
wyrdough, line 25: Can you elaborate (in a comment here, not necessarily in the code) on why you're using ```FFT_SIZE * 2``` for the size of all the buffers?

My understanding is that for the FFT data only ```FFT_SIZE / 2``` is required, although the fft buffer and wave buffers would need to be the same size, so maybe wave needs to be ```FFT_SIZE * 2```?

wyrdough (changes requested):
While this has always worked (and continues to work in this latest revision), the suggested changes make SoundshaderLines match the original ShaderToy version much better and eliminate the artifacts I mentioned on Discord without having to mess with MIN_DB and MAX_DB.

What I can't say is whether using FFT2048 instead of FFT512 is too much of a performance impact, as I have only made sure that it compiles and runs OK on my system. It may be burning a bunch more CPU without me having noticed since I didn't even check.

Assets/Script/Gameplay/TextureManager.cs
theli-ua: I actually think it used to be 2048 at some point and part of it was cut off, I think the whole / * by two is somewhere from there. It got lost somewhere in translation between revisions. 
like, fft_size used to be related to texture size but then you'd have buffers as double as you'd have twice more data from bass fft call. Thanks for looking into this. I'll clean it up

wyrdough (approved):
I think this is in good shape now. Certainly works great for me. (aside from some of the shaders I've ported being complete pigs and like halving my framerate, but that's not an issue with the PR)

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -0,0 +1,153 @@
...
+                _wave[i] = (_rawWave[i * AUDIO_CHANNELS] + _rawWave[(i * AUDIO_CHANNELS) + 1]) * PER_CHANNEL_MULTIPLIER;
+
+                // set spectrum data in the first row
+                pixelData[i] = (byte) Math.Round(normalized);
+                // waveform data in the second row
+                pixelData[FFT_TEXTURE_WIDTH + i] = (byte) Math.Max(0, Math.Min(255, 128 * (_wave[i] + 1)));
+            }
+        }
+
+        public async void FixedUpdate()
```
EliteAsian123, line 140: I'd discourage the use of `FixedUpdate` as it's meant for Unity's physics system. `Update` with a timer is probably a better option.

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -0,0 +1,153 @@
...
+        [Range(0.0f, 0.95f)]
+        public float waveSmoothingFactor = 0.5f;
+
+        private Texture2D _sourceIcon;
+        private Texture2D _albumCover = null;
+        private Texture2D _soundTexture = null;
+        private float[] _fft = new float[FFT_SIZE / 2];
+        private float[] _wave = new float[FFT_TEXTURE_WIDTH];
+        private float[] _prevFft = new float[FFT_SIZE / 2];
+        private float[]   _rawFft       = new float[FFT_SIZE * 2];
```
EliteAsian123, line 29: Some weird spacing here.

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -0,0 +1,153 @@
...
+                };
+            }
+            return _soundTexture;
+        }
+
+        public void processMaterial(Material m)
+        {
+            if (m.HasTexture(_sourceIconId))
+            {
+                m.SetTexture(_sourceIconId, _sourceIcon);
```
EliteAsian123, line 86: Is there a specific reason why the source icon is being computed in `Start` but everything else isn't? I'd either move the source icon getter here or move everything to `Start` to make it consistent.

EliteAsian123 (changes requested):
Hello, thanks for the PR and sorry for the long wait!

Firstly, just make sure you're using the correct name styling. For example public fields such as `fftSmoothingFactor` should be in `PascalCase`, methods should also always be in `PascalCase`, etc.

I've also left some comments below!

Thanks, and have a good day.

Assets/Script/Gameplay/TextureManager.cs
theli-ua: It's not necessarily for physics while it is the main usecase. 
I disagree here as Update with a timer seems silly if there is already an exposed fixed update trigger here. 
If you'd prefer I could make it *just* `Update` but not `Update` + timer, though imo `FixedUpdate` is just about perfect fit here

Assets/Script/Gameplay/TextureManager.cs
theli-ua: > Is there a specific reason why the source icon is being computed in `Start` but everything else isn't? I'd either move the source icon getter here or move everything to `Start` to make it consistent.

A reason would be that _sourceIconId is sort of readily available while the rest needs to be created/loaded. But I agree that it would be more consistent to move that code from Start into a getter.

Assets/Script/Gameplay/TextureManager.cs
RileyTheFox: I think the concern is more why does it specifically need to be the rate of FixedUpdate (default to 50 times per second) and if so, that should be communicated somewhere.

An Update + Timer allows easy change of the rate at which it is called which is better understood. FixedUpdate is *traditionally* used for physics components so there is potential confusion when it is used with no physics and the reason is not clear.

EliteAsian123: I forgot to reply to this, but @theli-ua, to expand on the reasoning as to why you shouldn't use `FixedUpdate`: it's because Unity tries to run it as consistently as possible. This means that if, for example, the `FixedUpdate` rate is 0.02s and the frame lasts 0.04s, `FixedUpdate` will be called twice in that frame, which is more of a waste. Rendering-related stuff should always be in `Update`, as this scenario would not happen (you only *need* to update rendering-related stuff once per frame because changes made in between are not necessary). Using a timer in `Update` would fix the issue while keeping the intended behaviour you want here, as it's still guaranteed it'll only be run once per frame at most.

PR #1005, opened 2025-04-01
Add background to text notifications

wyrdough:
This PR adds a background to text notifications to make them more visible when using venues. The background is colored based on the type of notification. At present, only bass groove, star power ready, and full combo use colors.

### Default
hot start

### Bass Groove
bass groove

### Star Power
star power

Reference to Discord suggestion where this was discussed: https://discord.com/channels/1086048856678084609/1355248369269739732

### DIFF
--- a/Assets/Art/Textures/Gameplay/NotificationBackground.png
+++ b/Assets/Art/Textures/Gameplay/NotificationBackground.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:fa67e81d0a5a87a98f3adeba4efd852800262efe782bbd395662577aaf11b03e
+size 23525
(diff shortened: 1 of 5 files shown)


TheNathannator (changes requested):
While I like the design for the background, trying it out against various backgrounds tells me it needs some adjustments to work in practice. There's too much transparency in the regions that the text actually occupies, and so the text background itself tends to blend in with similar colors, and doesn't do much to help with legibility on bright backgrounds.

I recorded a showcase to demonstrate how the text background currently interacts with different shades and colors, using various backgrounds from my current collection: https://www.youtube.com/watch?v=wWwrpNy-Y3M

wyrdough:
> While I like the design for the background, trying it out against various backgrounds tells me it needs some adjustments to work in practice. There's too much transparency in the regions that the text actually occupies, and so the text background itself tends to blend in with similar colors, and doesn't do much to help with legibility on bright backgrounds.
> 
> I recorded a showcase to demonstrate how the text background currently interacts with different shades and colors, using various backgrounds from my current collection: https://www.youtube.com/watch?v=wWwrpNy-Y3M

Thanks for taking the time to record that, it definitely helps clarify the situation.

As I mentioned on Discord, the difficulty is balancing the look on dark/no backgrounds with light ones. Something will always be suboptimal unless the alpha of the background can be adjusted on the fly based on what is rendering behind it at any given time.

My feeling on it was that it does solve the problem it was trying to solve, in that it makes the text much more visible against light backgrounds, and doesn't look any worse than the previous behavior in any case.

PR #257, opened 2023-05-06
Random Names from Team

nevespt:
[EDITED]
- Added a function to read from Credits.txt and add a random name for the "Empty" Player Name
- Change the "BOT" text to blue color

[CHANGED]
- Changed Location of Credits.txt from '/Assets' to '/Assets/Resources'

PR #1007, opened 2025-04-04
Add support for shift indicators for five fret instruments

wyrdough:
This PR adds support for shift indicators on five fret instruments and frontend support for a "no range shifts" modifier that compresses all notes in a song into the starting range, using range shift text events found in some of the YARG setlist songs.

During initialization the list of text events attached to the note track is scanned for events with the form `[ld_range_shift d r s]` where d is the difficulty, r is the 1 indexed range, and s is the number of frets in the range. s is now optional. If it is not present, it defaults to 3 on Easy difficulty and 4 on Hard. Any frets not in the initial range are grayed out.

inactive fret

When a range shift event occurs, range shift markers are shown at the four preceding beat lines on the track and the currently inactive frets coming into use flash at each range shift marker.

Shift marker, currently identical to the pro keys range shift marker:
shift marker

Fading out from the beat line flash:
mid flash

Once the event has passed, the now active frets are colored and the now inactive frets fade to gray.

inactive flipped

This PR supersedes #959 and requires YARG.Core PR [#257](https://github.com/YARC-Official/YARG.Core/pull/247).

(Marked as draft because the Core PR still needs a modifier icon)


Assets/Script/Gameplay/Player/FiveFretPlayer.cs
```diff
@@ -337,5 +501,151 @@ public override (ReplayFrame Frame, ReplayStats Stats) ConstructReplayData()
...
+
+            foreach (var shift in _rangeShiftEventQueue.ToList())
+            {
+                if (shift.Range == lastShiftRange)
+                {
+                    continue;
+                }
+
+                var shiftLeft = LeftyFlip ? shift.Range < lastShiftRange : shift.Range > lastShiftRange;
+                lastShiftRange = shift.Range;
```
grishhung, line 590: An edge case: This is technically not proper charting, but if we had a range going from one size to another, there would always be a direction. To be thorough

| start range | end range | resulting direction |
| --- | --- | --- |
| `GRY--` | `GRY--` | none |
| `GRY--` | `GRYB-` | right |
| `GRY--` | `-RYB-` | right |
| `GRYB-` | `GRY--` | left |
| `GRYB-` | `GRYB-` | none |
| `GRYB-` | `-RYB-` | right |
| `-RYB-` | `GRY--` | left |
| `-RYB-` | `GRYB-` | left |


It might be better to instead change the conditions to be something along the lines the following:

```cs
var lastShift = mostRecentEvent;

foreach (var shift in _rangeShiftEventQueue.ToList())
{
    // The "value" represents the general placement
    var lastShiftValue = lastShift.Range + lastShift.Size;
    var shiftValue = shift.Range + shift.Size;

    if (shiftValue != lastShiftValue)
    {
        // Not a fan of continue, but this is just a general idea
        continue;
    }

    var shiftLeft = shiftValue < lastShiftValue;

    if (leftyFlip)
    {
        shiftLeft = !shiftLeft;
    }

    lastShift = shift;

    // etc.
```

Assets/Script/Gameplay/Visuals/Fret/FretArray.cs
```diff
@@ -158,5 +171,52 @@ public void ResetAll()
...
+                if (frets[i])
+                {
+                    _frets[i].ResetColor(true);
+                }
+                else
+                {
+                    _frets[i].DimColor(true);
+                }
+                _activeFrets[i] = frets[i];
+            }
```
grishhung, line 219: nit. This condition and other surrounding ones making use of `continue` are little bit hard to read. Could be simplified into:

```cs
for (int i = 0; frets < _frets.Count; i++)
{
    if (_activeFrets[i] != frets[i])
    {
        if (frets[i])
        {
            _frets[i].ResetColor(true);
        }
        else
        {
            _frets[i].DimColor(true);
        }

        _activeFrets[i] = frets[i];
    }
}
```

Assets/Script/Helpers/FiveFretRangeShift.cs
```diff
@@ -0,0 +1,38 @@
...
+using YARG.Core.Chart;
+using YARG.Core.Logging;
+
+namespace YARG.Helpers
+{
+    // TODO: Figure out a better place to put this
+    public class FiveFretRangeShift
+    {
+        public double Time         { get; private set; }
+        public int    Range        { get; private set; }
```
grishhung, line 12: nit. When I first read this, I thought this was referring to the size, not the position. This isn't super important; it's just something I noticed that threw me off for a second.

Assets/Script/Gameplay/Player/FiveFretPlayer.cs
grishhung: In hindsight, `-RYB-` to `GRYBO` and vice versa probably shouldn't be no range shift, but the point that we need to handle range shifts from one size to another still stands.

Assets/StreamingAssets/lang/en-US.json
```diff
@@ -60,7 +60,9 @@
             "TapsToHopos": "Taps to HOPOs",
             "UnpitchedOnly": "Unpitched Only",
             "NoDynamics": "No Dynamics",
-            "NoVocalPercussion" : "No Percussion"
+            "NoVocalPercussion" : "No Percussion",
+            "NoDynamics": "No Dynamics",
```
grishhung, line 64: Should this be here twice?

Assets/Script/Gameplay/Player/FiveFretPlayer.cs
grishhung: Discussed offline; the above could give the incorrect impression that the position (with respect to the index finger) would need to change when it wouldn't. This is such a niche edge case that it's probably fine as-is.

sonicfind (changes requested):
There are more things I wish I could say, but *those* problems are more a consequence of YARG's rendering design. I can't really comment much without requiring a complete redo the of the entire rendering system to use instancing instead of pools.

So for now, so long as no explicit errors show up, the current comments are all I'll require.

Assets/StreamingAssets/lang/en-US.json
wyrdough: Oops, must have messed up when rebasing after the no vocal percussion thing was merged... Thanks for catching that.

Assets/Script/Helpers/FiveFretRangeShift.cs
```diff
@@ -0,0 +1,38 @@
...
+            var shiftEvents = new List<FiveFretRangeShift>();
+
+            foreach (var shift in instrumentDifficulty.RangeShiftEvents)
+            {
+                shiftEvents.Add(new FiveFretRangeShift(shift.Time, shift.Range, shift.Size));
+            }
+
+            return shiftEvents;
+        }
+    }
```
sonicfind, line 37: Use an array, not a list. We already know the exact number of elements.

Assets/Script/Gameplay/Player/FiveFretPlayer.cs
```diff
@@ -122,26 +156,156 @@ protected override void FinishInitialization()
...
+            {
+                var poolable = _shiftIndicatorPool.AllSpawned[i];
+                _shiftIndicatorPool.Return(poolable);
+            }
+
+            for (var i = 0; i < _rangeIndicatorPool.AllSpawned.Count; i++)
+            {
+                var poolable = _rangeIndicatorPool.AllSpawned[i];
+                _rangeIndicatorPool.Return(poolable);
+            }
```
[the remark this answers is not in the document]
wyrdough: Coincidentally, I just noticed the existence of that method earlier today when working on something else...

wyrdough:
Should be good to merge now unless I misunderstood some comments. I did play test after the changes to make sure everything still works.

Assets/Script/Gameplay/Player/FiveFretPlayer.cs
```diff
@@ -505,32 +490,36 @@ public override (ReplayFrame Frame, ReplayStats Stats) ConstructReplayData()
 
         private void InitializeRangeShift(double time = 0)
         {
+            if (!Player.Profile.RangeEnabled)
+            {
+                return;
+            }
+
```
sonicfind, line 497: Move the checks for `RangeEnabled` to the callers. It'd allow you to early-exit `ResetRangeShift` immediately and wrap the initialization of `_allRangeShiftEvents` and `_activeFrets` inside `Initialize()`, making it so you don't need to do any null checks on either variable.

PR #1009, opened 2025-04-09
Add venue rendering scaling + FSR3.1 and AA controls

theli-ua:
Adds ability to render venue at lower resolution

Adds setting to control anti aliasing for venue camera, selection of None/FXAA/MSAA

It also allows FSR3 AA/upscaling, however it doesn't seem like its performance is sufficient to use for upscaling. it does work great as temporal anti aliasing though

### DIFF
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -121,6 +121,7 @@ namespace YARG.Gameplay
                     var bundleBackgroundManager = bgInstance.GetComponent<BundleBackgroundManager>();
                     bundleBackgroundManager.Bundle = bundle;
                     bundleBackgroundManager.ShaderBundle = shaderBundle;
+                    bundleBackgroundManager.SetupVenueCamera(bgInstance);
 
                     // Destroy the default camera (venue has its own)
                     Destroy(_videoPlayer.targetCamera.gameObject);
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -121,6 +121,7 @@ namespace YARG.Gameplay
                     var bundleBackgroundManager = bgInstance.GetComponent<BundleBackgroundManager>();
                     bundleBackgroundManager.Bundle = bundle;
                     bundleBackgroundManager.ShaderBundle = shaderBundle;
+                    bgInstance.GetComponentInChildren<Camera>().gameObject.AddComponent<VenueCameraManager>();
 
                     // Destroy the default camera (venue has its own)
                     Destroy(_videoPlayer.targetCamera.gameObject);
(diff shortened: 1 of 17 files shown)


wyrdough:
Works a treat! I can run YARG in 4k with the shader venues now using the balanced upscaling setting.

Two comments, though.

First, do we really want this to be imported as a submodule rather than copied into the tree like the other dependencies that aren't unity packages?

Second, and I don't think this is really an issue, but just want to point it out for the sake of completeness, it does happen that sometimes the framerate is low at the very beginning of the song, but by the time the two second lead-in is done it stabilizes at 60fps.

Assets/Script/Gameplay/BackgroundManager.cs
```diff
@@ -121,6 +121,8 @@ private async UniTaskVoid Start()
                     var bundleBackgroundManager = bgInstance.GetComponent<BundleBackgroundManager>();
                     bundleBackgroundManager.Bundle = bundle;
                     bundleBackgroundManager.ShaderBundle = shaderBundle;
+                    var camManager = bgInstance.GetComponentInChildren<Camera>().gameObject.AddComponent<VenueCameraManager>();
```
TheNathannator, line 124: I'd suggest moving this component addition (and future venue initialization code) to inside of `BundleBackgroundManager` instead, there's a `mainCamera` field there which can be used to ensure it always gets applied to the same camera every time the venue is loaded.

TheNathannator (changes requested):
The only reason other dependencies are copied in is because it's either not feasible as a submodule, or we made modifications specific to YARG in it. I wouldn't mind it being a submodule, I'd just nitpick its placement and prefer it go into the `Packages` folder (or into a new `Submodules` folder if it's not a Unity package).

However, in this case, it would be more appropriate to use a Git URL in the packages.json instead of a submodule, since it is structured as a Unity package already. You can see an example of using a Git URL directly below the new entry: we use a custom Minis fork to fix various issues with it (and because the developer doesn't seem active).

I went through everything else and left some other comments, mostly just nitpicks lol

Assets/StreamingAssets/lang/en-US.json
```diff
@@ -853,6 +857,28 @@
...
+                    "FXAA": "FXAA",
+                    "MSAA": "MSAA",
+                    "FSR3": "FSR3"
+                }
+            },
+            "VenueRenderingQuality": {
+                "Name": "Venue rendering quality",
+                "Description": "Changes rendering resolution for venues",
+                "Dropdown": {
+                    "NativeAA": "Native",
```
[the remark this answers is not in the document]
theli-ua: I thought it has to match enum variant for the enum this setting is associated with, it doesn't?

Assets/Script/Gameplay/BackgroundManager.cs
theli-ua: I did that, though ,frankly, `BundleBackgroundManager` implies it handles asset bundles (and thats what it does) while `BackgroundManager` seems like an entity that manages things about backgrounds.

However if we rename `BundleBackgroundManager` into something like `VenueBackgroundManager` then it would make much more sense ... though that would break compatibility ... hm ...

PR #1016, opened 2025-04-19
Force GC right before song starts

theli-ua:
It's a common PR feedback to avoid generating garbage even in a library view or scene setup as it "may be collected during gameplay and cause stutter" To alleviate those concerns it seems natural to simply force garbage collection before song start so that any garbage generated during setup would be collected then.

### DIFF
--- a/Assets/Script/Persistent/GlobalVariables.cs
+++ b/Assets/Script/Persistent/GlobalVariables.cs
@@ -163,6 +163,7 @@ namespace YARG
             {
                 LoadSceneAdditive(scene);
             }
+            GC.Collect();
         }
 
         // Due to the preprocessor, it doesn't know that an instance variable is being used
--- a/Assets/Script/Persistent/LoadingScreen.cs
+++ b/Assets/Script/Persistent/LoadingScreen.cs
@@ -128,6 +128,7 @@ namespace YARG
                     YargLogger.LogException(ex);
                 }
             }
+            GC.Collect();
         }
 
         public async void Dispose()


Assets/Script/Gameplay/GameManager.Loading.cs
```diff
@@ -211,6 +211,8 @@ private async void Start()
             // Loaded, enable updates
             enabled = true;
             IsSongStarted = true;
+            // Collect all garbage potentially generated  in library/venue setup, etc
+            System.GC.Collect();
```
TheNathannator, line 215: Might just be a nitpick on my part lol, but this particular placement of the GC call feels a little awkward/shoehorned. I'd prefer for this to be done in `LoadingContext.Wait`, right after all queued loading tasks are finished up. (Would also be good to put it in `GlobalVariables.LoadScene` in addition to this.)

Would also prefer not using the fully-qualified name here, add a `using System;` directive instead and change it to:

```suggestion
            // Collect all garbage potentially generated  in library/venue setup, etc
            GC.Collect();
```

TheNathannator (commented):
Curious whether this actually helps much in practice, I've had a similar thought in the past but haven't taken the time to look into it.

Even with this though, it's generally not ideal to have spurious allocation happening throughout any scene, gameplay or not. I wouldn't allow this addition to be used as an excuse to be lax elsewhere; GC stutters still affect other scenes (especially on lower-end devices), and there's almost always a better way to handle things in scenarios that frequently allocate new objects.

I think the most this change would actually help with in practice is more quickly clearing out data that ends up getting moved to long-lived/large-object heaps, such as chart data (which Unity seems to be reluctant to clear out, most noticeable with large songs like Crash Test 5). Still worth having if it does help clear that out, but this otherwise doesn't relax the need to avoid allocation for frequently-called code.

Assets/Script/Gameplay/GameManager.Loading.cs
theli-ua: >but this otherwise doesn't relax the need to avoid allocation for frequently-called code.  

This is more relevant to "do not use linq when fixing up shaders when loading yarground because it can later be collected when song is already running" comments rather than something happening often.

and fwiw there are already things  that generate garbage literally every frame. see #979 for  an example

PR #1017, opened 2025-04-22
Highway animation 2025

grishhung:
> This is a continuation of this older PR https://github.com/YARC-Official/YARG/pull/823, with some minor differences. The below has been copied and pasted from that PR and tweaked to reflect changes.

### Notes

- Removed existing `PerformanceTextScaler` class and replaced with `DOTween` implementations
- Added highway raise animation; takes 0.5 seconds w/ keyframes on 1/3s and 1/2s
- Added strikeline zoom aniamtion; takes 0.5 seconds w/ keyframes on 1/6s and 1/4s (and 1/2s of delay)
- Separated multiplayer highway animations by 0.1 seconds
  - 1/3s ended up looking too busy with the two-part animation
  - Longer durations cause problems if there are too many players
- Shortened vocal performance text to 1 second to match all Rock Band games
  - Was previously 2 seconds; text lingers on screen for too long for some short phrases 

### Testing

https://github.com/user-attachments/assets/ac5b7ce7-103f-49ed-8501-c2b4b44f9c4b

### Action items

- Add a highway raise sound; cc: @kaduwaengertner @bbrsk
- Make the vocals highway drop in from the top and the other UI elements come in from the side
  - May need to disable this in the case that the UI elements have custom positioning
  - Alternatively, make the animation dynamic based on the placement (could look very bad)
- Add a matching highway lowering animation for failing out and completing the chart (lower on `[end]` events)
- Discussed with @kaduwaengertner: after backgrounds are in, delay the highway raise to on the first tick of the chart so that it raises when you hear the count-in sound, just like in Rock Band (following the visual of the band on stage)
  - In order to account for poorly made charts where the first note is on tick 0, handle this animation so that the highway always raises at least 2 seconds before the first charted note (this is the whole reason we have the delay in the first place)


wyrdough:
I like the look of the animation, but the delay at the beginning almost makes me feel like the game has hung if I'm not using a background. Perhaps the first track should begin its animation sooner?

hennessyevan:
I like the delay FWIW, I think it adds some slight anticipation which I appreciate

Assets/Script/Gameplay/Visuals/CameraPositioner.cs
```diff
@@ -40,6 +56,12 @@ public void Initialize(CameraPreset preset)
 
             // Rotation
             transform.localRotation = Quaternion.Euler(preset.Rotation, 0f, 0f);
+
+            // Set camera preset
+            _preset = preset;
+
+            // Animate the highway raise
+            _coroutine = StartCoroutine(RaiseHighway(_preset, true));
```
wyrdough, line 64: Maintainers probably have better ideas about how to do this, but this should be protected by a check that we're actually in gameplay before running the coroutine, otherwise this breaks the preview track in settings. The suggestion works well enough for testing, though.

```suggestion
            if (FindObjectOfType<GameManager>() != null)
            {
                _coroutine = StartCoroutine(RaiseHighway(_preset, true));
            }
```

Assets/Script/Gameplay/Visuals/StrikelinePositioner.cs
```diff
@@ -0,0 +1,56 @@
...
+        private const float ANIM_INIT_Z_OFFSET           = -1.2f;
+        private const float ANIM_PEAK_Z_OFFSET           = 0.1f;
+
+        private float     _zOffset;
+        private Coroutine _coroutine;
+
+        private void Start()
+        {
+            _zOffset = transform.localPosition.z;
+            _coroutine = StartCoroutine(RaiseStrikeline(true));
```
wyrdough, line 28: Here as well, with the same caveat about this probably not being the best way of doing it, just a way that works.

```suggestion
            if (FindObjectOfType<GameManager>() != null)
            {
                _coroutine = StartCoroutine(RaiseStrikeline(true));
            }
```

wyrdough (approved):
Tested to work, so if you're happy with the delay before the first highway raises, I'm fine with all of this.

wyrdough:
> Added animations for star power activation:

Having playtested it some, I think that the SP activation animation looks kinda out of place until we get some kind of track effect going. Otherwise I'm super happy with the visual aspects of this.

(expect a couple more review comments at some point, as there does seem to be an issue with the tweens still trying to run as the song ends)

PR #1018, opened 2025-04-23
Allow to set fps cap to 0 to disable it

theli-ua:
More straightforward then to set it to 9000


Dansla116:
I verified that setting the FPS cap to 0 now disables the cap and lets the game render at my monitor’s refresh rate (tested on 1440p@144Hz and 1080p@60Hz).  Previously, setting 0 defaulted to 1, but with this change, it stays 0 as expected.

Everything below is more or less also what you'd expect.  Some different settings I toyed around with:
- VSync on/off didn’t change the behavior - works the same.
- Fullscreen and Borderless Fullscreen both worked identically.
- Windowed mode made the game’s FPS match the highest monitor refresh rate on a multi-monitor setup (not just the window’s monitor).

All in all, this looks good to me!

</img>

</img>

PR #1025, opened 2025-05-10
Fix particle blending

theli-ua:
Fixes transparency issues with particle backgrounds (especially vocal percussions)

While still not allowing particles make alpha holes in the track


wyrdough (approved):
Just noting that I have playtested this some, it does what it says on the tin, and didn't lead to any obvious breakage.

PR #1027, opened 2025-05-10
Generalize and localize practice sections (v2)

Frickitickitavi:
Fresh implementation of PR #950. Differences from the last version that received a review are detailed below.

## Two-step localization for section numbers and subsection letters
Previously, the code localized section names without section numbers and subsection letters (henceforth referred to as "suffixes" for conciseness), then manually concatenated the suffix afterward. This was a flawed approach because some languages might want the suffix placed somewhere other than the end of the section name, and I received a suggestion to give every section name two localizations: one with no suffix and one with a suffix, like so:

```
"verse": {
    "WithoutNumber": "Verse",
    "WithNumber": "Verse {0}",
},
"chorus": {
    "WithoutNumber": "Chorus",
    "WithNumber": "Chorus {0}",
},
//etc
```

This, of course, had the drawback of writing every section name twice, leading to file bloat and redundant information. Instead, I implemented a two-step localization process. First, we localize the section ignoring the suffix, so `verse_2b` becomes "Verse" at step 1. Then, for sections that have prefixes, we pass the first-step localized string and the suffix to a single new localization key, `"WithNumber": "{0} {1}"`, resulting in "Verse 2B".

This means that each section name only needs to be written once, and languages can change the order in which the name and suffix are presented. Other languages will still use Latin letters and Arabic numerals for suffixes (this was already the case in the previous implementation).

## Improved Tolerance and `.chart` Compatibility
Although the old implementation would parse `[section]`-based events, as typically found in `.chart` files, it would only accept them if they exactly matched the content of a `[prc]`-based section. In other words, `[section gtr_solo_2]` would work, but the more likely case of `[section Guitar Solo 2]` would not.

To maximize compatibility with existing charts and minimize charter effort, this new implementation includes a step that sanitizes the event text to increase the likelihood of it matching a standardized event. We convert to lowercase and replace all spaces with underscores. Additionally, because the standard events always use `gtr` instead of the full word, it also converts `guitar` to `gtr`.

I've also expanded the list of accepted name variations to allow for things like `fade-in` instead of `fade_in`, further improving compatibility. I could have also accounted for `guitar->gtr` here, but because that spelling appears in so many sections, it was easier to replace it up front.

## Support for Letter-Based Sections With Subsection Letters but No Section Numbers

Previously, the implementation accepted the format `a` for "A section", `a1` for "A section 1", and `a1b` for "A section 1B", but had no way to render "A section B". We can't naively accept the format `ab`, because it would collide with the standardized section `ah` ("Ah!"). Instead, this implementation accepts the format `a_b`.

## New Sections

I've added a few new sections that were not part of the original specification, such as `violin_solo` and `drop`.


Frickitickitavi:
@Dansla116 thanks for the detailed review! I've removed `using UnityEditor.Build.Pipeline;` - not sure how it got there, but good catch.

It's cool to see that this actually has some beneficial effects on GH rips (mostly by removing underscores). Since, as you mention, they used freeform section names in each song, I didn't really factor them in at all.

You're correct that this is mostly future groundwork, as is currently the case with all of YARG's localization. The biggest immediate benefit will be for RB-standardized sections whose tags differ significantly from their display text, such as `bre` for "Big Rock Ending!", `oohs` for "Oohs and Ahs", and all of the letter-based sections like `a1` for "A section 1". We use a lot of letter-based sections in the public domain DLC packs, so that's something to look forward to.

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
```diff
@@ -101,5 +105,157 @@ public void UpdateView()
...
+            {
+                return false;
+            }
+
+            if (text.Length == 1)
+            {
+                return text[0].IsAsciiLetterLower() || text[0].IsAsciiDigit();
+            }
+
+            return text[..^2].All(c => c.IsAsciiDigit()) && (text[^1].IsAsciiLetterLower() || text[^1].IsAsciiDigit());
```
wyrdough, line 231: Since this runs during gameplay, I suspect some will be unhappy with the use of LINQ here.

You could instead do this with a for loop, though it wouldn't be as concise.

```suggestion
            for (int i = 0;i < text.Length - 2; i++) 
            {
                if (!text[i].IsAsciiDigit()) {
                    return false;
                }
            }
            return text[^1].IsAsciiLetterLower() || text[^1].IsAsciiDigit();
```

(not tested, just typed into GitHub's terrible little text box)

Assets/Script/Gameplay/HUD/Practice/PracticeSectionView.cs
```diff
@@ -101,5 +105,157 @@ public void UpdateView()
...
+                return text[1].IsAsciiDigit();
+            }
+
+            if (text[1] == '_')
+            {
+                // e.g. "a_b" for "A section B"
+                return text.Length == 3 && text[2].IsAsciiLetterLower();
+            }
+
+            if (text[1..^1].All(c => c.IsAsciiDigit()))
```
wyrdough, line 181: Also using LINQ here. In this case the alternative would be more like:
```suggestion
            bool isDigit = true;
            for (int i = 1;i < text.Length - 1; i++)
            {
                if (!text[i].IsAsciiDigit()) 
                {
                    isDigit = false;
                    break;
                }
            }
            if (isDigit) {
```

wyrdough (commented):
For my part, I'd be a bit more lax about LINQ in most cases and just make people throw in a comment like `// LINQ` to ensure that any such usages could be easily found in the future when it actually became a problem.

(Indeed, these particular calls may not even allocate on the heap, IDK, just wanted to call them out for whoever does the final review/merge and offer an alternative strategy in case it does get flagged as an actual problem)

Otherwise, I think this is about as good as processing free form text gets and would approve if it were up to me.

EliteAsian123 (dismissed):
Hello, thanks for the PR!

I've left some comments below, but I'd advise moving all of this to a static helper class so it can be used elsewhere in the future if needed.

I agree with @wyrdough's review as well, so once these changes are made, this should be good to merge!

Frickitickitavi:
> Hello, thanks for the PR!
> 
> I've left some comments below, but I'd advise moving all of this to a static helper class so it can be used elsewhere in the future if needed.
> 
> I agree with @wyrdough's review as well, so once these changes are made, this should be good to merge!

All addressed! I also added the little `// LINQ` comments like wyrdough suggested

PR #1030, opened 2025-05-18
Add support for user-supplied song sources

raphaelgoulart:
- Custom sources are prioritized, so user-provided replacements supersede the official ones;
- The location is `persistent data path / custom / icons`. To avoid duplicate code, the `CustomContentManager.CustomizationDirectory` variable is used;
- If there's an `index.json` there, that is used (similar to OpenSource sourced sources). It's up to the user to maintain that file;
- Otherwise, `.png`s are read individually. If it's a new source, the filename is used as its id/name/etc.
  - If a source already exists for a given PNG's filename, the existing source will be modified to use the custom icon instead. This is done so the metadata for existing sources aren't lost.

### DIFF
--- a/Assets/Script/Song/SongSources.cs
+++ b/Assets/Script/Song/SongSources.cs
@@ -14,6 +14,7 @@ using YARG.Core.Logging;
 using YARG.Core.Song;
 using YARG.Helpers;
 using YARG.Helpers.Extensions;
+using YARG.Settings.Customization;
 
 namespace YARG.Song
 {
@@ -50,7 +51,7 @@ namespace YARG.Song
 
         public class ParsedSource
         {
-            private readonly string _icon;
+            public string _icon;
             private readonly Dictionary<string, string> _names;
 #nullable enable
             private Sprite? _sprite;
@@ -122,6 +123,7 @@ namespace YARG.Song
 #else
         public static readonly string SourcesFolder = Path.Combine(PathHelper.StreamingAssetsPath, "sources");
 #endif
+        public static readonly string CustomSourcesFolder = Path.Combine(CustomContentManager.CustomizationDirectory, "icons");
 
         private static readonly string[] SourceTypes =
         {
@@ -130,6 +132,7 @@ namespace YARG.Song
 
         private static readonly string[] SourceRoots =
         {
+            CustomSourcesFolder, // Prioritize user-supplied sources
             Path.Combine(SourcesFolder, SOURCE_REPO_FOLDER, "base", "icons"),
             Path.Combine(SourcesFolder, SOURCE_REPO_FOLDER, "extra", "icons"),
         };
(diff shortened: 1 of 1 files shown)


Assets/Script/Song/SongSources.cs
```diff
@@ -50,7 +51,7 @@ public class Source
 
         public class ParsedSource
         {
-            private readonly string _icon;
+            public string _icon;
```
TheNathannator, line 54: Small nitpick about the encapsulation here, is there a way to do this without requiring modification of the field?

Assets/Script/Song/SongSources.cs
```diff
@@ -322,6 +318,56 @@ private static void ReadSources()
                     }
                 }
             }
+
+            // Read individual PNGs (this is done after reading regular sources, so their icons can be replaced if needed)
+            if (!customSourcesRead) {
```
TheNathannator, line 323: Code style nitpick:

```suggestion
            if (!customSourcesRead)
            {
```

Assets/Script/Song/SongSources.cs
```diff
@@ -322,6 +318,56 @@ private static void ReadSources()
...
+                    {
+                        if (icon == DEFAULT_KEY)
+                        {
+                            _default = parsed;
+                        }
+                    }
+                    else // Source already exists; override its icon
+                    {
+                        _sources[icon]._icon = icon;
+                    }
```
TheNathannator, line 342: I believe this can be simplified to just unconditionally inserting the icon, regardless of if it existed previously:

```suggestion
                    _sources[icon] = parsed;
                    if (icon == DEFAULT_KEY)
                    {
                        _default = parsed;
                    }
```

Assets/Script/Song/SongSources.cs
```diff
@@ -282,32 +285,25 @@ private static async UniTask DownloadSources(LoadingContext context)
 
         private static void ReadSources()
         {
+            // Read custom sources - they are read first so user replacements are prioritized
+            // Create the folder if it doesn't exist
+            Directory.CreateDirectory(CustomSourcesFolder);
+            // Read index.json if it exists; otherwise fallback to reading PNGs individually later on
+            string customIndexPath = Path.Combine(CustomSourcesFolder, "index.json");
+            var customSourcesRead = false;
```
TheNathannator, line 293: Nitpick: use explicit type for built-in types:

```suggestion
            bool customSourcesRead = false;
```

Assets/Script/Song/SongSources.cs
```diff
@@ -322,6 +318,56 @@ private static void ReadSources()
                     }
                 }
             }
+
+            // Read individual PNGs (this is done after reading regular sources, so their icons can be replaced if needed)
+            if (!customSourcesRead) {
+                PathHelper.SafeEnumerateFiles(CustomSourcesFolder, "*.png", true, (path) =>
+                {
+                    // Filename will be used for all values (id, name, icon path).
+                    var icon = Path.GetFileNameWithoutExtension(path);
```
TheNathannator, line 327:
```suggestion
                    string icon = Path.GetFileNameWithoutExtension(path);
```

Assets/Script/Song/SongSources.cs
```diff
@@ -322,6 +318,56 @@ private static void ReadSources()
...
+                var parsed = new ParsedSource(source.icon, source.names, source.type switch
+                {
+                    "game"    => SourceType.Game,
+                    "charter" => SourceType.Charter,
+                    "rb"      => SourceType.RB,
+                    "gh"      => SourceType.GH,
+                    _         => SourceType.Custom
+                });
+
+                foreach (var id in source.ids)
```
TheNathannator, line 363:
```suggestion
                foreach (string id in source.ids)
```

TheNathannator (changes requested):
Just a few nitpicks, otherwise looks good!

Assets/Script/Song/SongSources.cs
raphaelgoulart: Not really because, if the source already exists, I want to preserve its _metadata_ and replace only the icon itself.

For instance, if the user supplies a custom `gh.png`, surely they still want to keep the source name "Guitar Hero" intact. But if I do `_sources[icon] = parsed;`, the source name will be changed to `gh` (the filename) as well. `_sources[icon]._icon = icon;` will change the icon whilst preserving the metadata.

(This is also why I had to change `_icon` from `private` to `public` as well)

Assets/Script/Song/SongSources.cs
raphaelgoulart: The change itself is needed so a .png replacement of an existing source can replace its icon without affecting metadata.

However, I can think of a few alt. solutions:
- `internal` keyword (it's still changing the field, but better than `public`);
- Setter method for the `_icon` variable (still needs `readonly` to be removed);
- Constructor that takes `string icon, ParsedSource source` as arguments, and constructs a new ParsedSource with the updated icon and existing metadata (cumbersome but works with `_icon` being `private readonly`).

I've tested all of these and they work, just let me know which solution you prefer (including leaving it as is since it works)

Assets/Script/Song/SongSources.cs
TheNathannator: Ah I didn't think that through all the way lol, all good here then.

I would have used `TryGetValue` to probe for existence instead of `TryAdd`, but the difference is minimal for a one-and-done operation, and so nitpicking the allocation efficiency of new `ParsedSource` instances isn't really necessary.

Assets/Script/Song/SongSources.cs
TheNathannator: `internal` wouldn't really have an effect since all of the code is in the same assembly, it's not relative to namespaces.

After thinking about it I don't think I really care about the encapsulation that much anymore lol, a simple `public string IconName { get; set; }` will do. Habits from the perspective of writing a library are hard to put aside ^^;
They can certainly be helpful, but it's also easy to forget the bigger picture of how the code is actually used; there's no real reason this will get modified outside of `SongSources`.

TheNathannator (approved):

PR #1032, opened 2025-05-22
Add multiplier glow effects

wyrdough:
This PR adds visual effects when the score multiplier changes.

- When multiplier increases, the sunburst effect "flashes" to more clearly indicate that the multiplier has changed
- When multiplier drops, the track light flashes red to more clearly indicate that the player did something wrong
- When at maximum multiplier, the sunburst pulses in size at every beat
- When transitioning to/from max multiplier, the sunburst and track light effects are eased instead of instantly changing state

https://github.com/user-attachments/assets/e0a57a28-94a5-447d-9876-5ac7c1f43db6

~~Currently marked as draft because, while it was discussed on Discord, Kadu has not yet given specific approval.~~

### DIFF
--- a/Assets/Art/Materials/Gameplay/Track/TrackComboSunburst.mat
+++ b/Assets/Art/Materials/Gameplay/Track/TrackComboSunburst.mat
@@ -112,7 +112,7 @@ Material:
     - _ZWrite: 1
     m_Colors:
     - _BaseColor: {r: 1, g: 1, b: 1, a: 1}
-    - _Color: {r: 1, g: 1, b: 1, a: 1}
+    - _Color: {r: 0, g: 0.8509804, b: 1, a: 1}
     - _EmissionColor: {r: 0, g: 0, b: 0, a: 1}
     - _Flip: {r: 1, g: 1, b: 1, a: 1}
     - _RendererColor: {r: 1, g: 1, b: 1, a: 1}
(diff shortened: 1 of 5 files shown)


Dansla116:
I've found a minor regression that is very hard to recreate naturally.  It happened to me once, and I was only able to reproduce it using a custom test track I set up.

When you run out of starpower at the exact moment you transition from a 3x to a 4x multiplier, the visual sunburst effect around the multiplier indicator doesn't trigger.

**To reproduce this _somewhat_ consistently:**
Download my test track here: [PR1032 Test.zip](https://github.com/user-attachments/files/20495547/PR1032.Test.zip)
Hit every green note.
Miss the one red note.
Activate starpower on the 3rd snare.

See my video demonstration here: https://youtu.be/lpZA5IGQrt0 **or** this screenshot from the end of the video:
image

Other than this minor edge case, I'd say the rest of the effects look really nice!  Once I played a few songs with this PR, it just blended in naturally.  It's one of those polished little details that makes the whole game feel more professional and complete.

wyrdough:
Wow, thanks for finding that and super thanks for putting in the effort to make it so easy to repro!

Dansla116:
I didn't want to leave you hanging.  I've downloaded the new patch and tried pretty hard to break it again (both using my test track and just playing naturally) and I can't reproduce the issue anymore.  So I'd say it's fixed, and everything looks great now!

I'll let you know if anything else pops up, but otherwise, no news is good news.  Nice work!

PR #1035, opened 2025-05-26
Play along with replay

wyrdough:
This PR adds a feature allowing users to play along with a selected replay by selecting a replay from the history screen and pressing the orange menu button. This launches into the usual difficulty select screen. Upon completion, the score screen is shown and the combined replay is saved as a new replay, allowing for a single player to build up a full band playthrough one instrument at a time.

~~Draft because this has so far only been tested with five fret and drums.~~


wyrdough:
Now tested to work with vocals and pro keys...

PR #1037, opened 2025-05-28
Handle charts with missing initial range shift

wyrdough:
It turns out that some of the YARG DLC charts have range shifts in the middle of the chart, but do not include one at the beginning.

This PR changes some logic in FiveFretPlayer so that no frets are dimmed until the first range shift if there is no range shift event prior to the first note of the chart.


wyrdough:
Since this is still open, I also added a fix for a different issue identified by @grishhung involving fret colors sometimes getting stuck in "half on". I never saw this in hours of testing, but somehow it happens reliably on his mac at the start of a song. Go figure.

PR #1046, opened 2025-06-24
Highway rendering overhaul

theli-ua:
This PR overhauls how highway rendering is done. Biggest motivator is performance.
This builds on top of #977 and supersedes it. I will close that one once/if this is fully polished (unless that one is somehow merged before that). This is now in a state where it's very usable and provides great perf benefits with multiple highways. Sharing for some feedback but keeping this as draft for now.
You will likely need to do Assets -> Refresh All if switching to/from this branch for unity to properly pickup/regenerate shadergraph shaders

Main highlights:
* What have to be opaques are rendered as opaques, no need to inject fading into everything. (So you can sample opaques texture in your transparent materials if wanted for example).
* fading is part of post processing (which was necessary to not render everything as transparents)
* curve is applied in vertex shaders, therefore has proper perspective
* depending on number of highways this doubles/triples/etc FPS, this basically minimizes overhead from multiple highways
* Vertex changes are injected automatically, there is no need to carry it into every new material
* All highways are rendered within one camera's scope. A new component sets up proper parameters and translation matrices in arrays and shaders figure out which one to use.

Quick (rough average of built in fps meter, didn't have much time for more extensive testing) testing on a steam deck. Vulkan. 2k resolution, no backgounds. Low Quality:
| Highways              | Nightly FPS | This PR FPS |
| :---------------- | -------: | ----: |
| 2          |   165   | 180 |
| 4    |  95   | 170 |
| 8 |  55   | 165 |

as far as single highway rendering, of all the people who tested this in #yarg-dev none reported regression, and on stress-test charts they reported improvements over nightly

As seen this does bring small overhead for single highway case but otherwise provides better performance as number increases.

Cons/Notes:
* Particle materials cannot use orientation relative to cameras
   * I can make this work if necessary, but for now all particles are set to orient themselves in world space
* Shadows are not rendered on highways
   * I haven't even tried to make it work since they never worked in the first place as everything was rendered as 
transparents anyway and therefore casted no shadows. I also think shadows on highways may look weird if actually implemented
* Max of 128 highways supported. (this can be changed by moving to passing params as textures instead of arrays, however it is hard to imagine anyone trying to use more until multiple monitors are supported)

TODO:
* ~~Settings preview~~

Maybe TODO:
* Open note sustain is not curved (does it matter?)
* Vocal track rendering within same camera (?)


wyrdough:
This may not be your bug (could be unity being weird with my build, could be an existing YARG bug), but I did notice that text notifications and the solo box sometimes didn't pop. In fact, they never did when running with 6-7 highways on 5 or so songs.

Just something to look out for when you're testing...

Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -93,6 +91,13 @@ public virtual void Initialize(int index, YargPlayer player, SongChart chart, Tr
             ZeroFadePosition = DEFAULT_ZERO_FADE_POS * Player.Profile.HighwayLength;
             FadeSize = Player.CameraPreset.FadeLength;
 
+            // Clip everything that was set to not be visible anyway
+            var camPosition = this.transform.position;
+            var trackZeroFadePosition = new Vector3(camPosition.x, camPosition.y, ZeroFadePosition + 3f);
+            Plane farPlane = new Plane();
+            farPlane.SetNormalAndPosition(this.TrackCamera.transform.forward, trackZeroFadePosition);
+            this.TrackCamera.farClipPlane = Mathf.Abs(farPlane.GetDistanceToPoint(this.TrackCamera.transform.position));
```
wyrdough, line 99: Don't need the `this.` here...

Assets/Script/Gameplay/Visuals/CameraPositioner.cs
```diff
@@ -27,10 +27,11 @@ private void Start()
             }
         }
 
-        public void Initialize(CameraPreset preset)
+        public void Initialize(CameraPreset preset, float fadePos, float fadeSize)
```
wyrdough, line 30: You're not using fadePos and fadeSize here.

wyrdough (commented):
Couple of teeny tiny nits here, but otherwise this looks great to me!

I do have one question, though (showing my lack of Unity knowledge here): will the copy of Varyings in the URP package folder get clobbered if we upgrade Unity?

(I assume there is a reason I don't understand for having two slightly different copies)

theli-ua:
>  In fact, they never did when running with 6-7 highways on 5 or so songs.

Indeed. and that is so weird. The countdown timers do still work though which is even weirder :-/

theli-ua:
The HUD elements scaling and positioning should be fixed now :tada:

theli-ua:
Build as of current state: https://github.com/theli-ua/YARG-BleedingEdge/releases/tag/rendering-overhaul-20250806

The remaining issue is that notes (gems?) do not go fully below starpower bar (and starpower bar frame really sucks and is not curving well imo )

PR #1049, opened 2025-06-29
Expand the light manager in general

Mekkabit:
More exposed light colors from the main manager and an expanded version of the neon manager to use them and temporarily overwrite the emission color of materials not in the existing list, then set it back when there's not a specific color being sent.

Existing material set functionality left alone for compatibility, and to still have the option of the current behavior.

(If there's a more efficient way to code this though then by all means, there's hardly an impact from my measurements but, copy-pasting was the best I could figure out)

### DIFF
--- a/Assets/Script/Venue/Lights/NeonLightManager.cs
+++ b/Assets/Script/Venue/Lights/NeonLightManager.cs
@@ -68,86 +68,37 @@ namespace YARG.Venue
                 {
                 if (_neonMaterialsFullColor[i].Location == VenueLightLocation.Generic)
                 {
-					if (lightState.Color == null)
-					{
-						_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, _neonMaterialsFullColor[i].InitialColor);
-					}
-					else
-					{
-						_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, lightState.Color.Value);
-					}
+					_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, lightState.Color ?? _neonMaterialsFullColor[i].InitialColor);
 					_neonMaterialsFullColor[i].Material.SetFloat(_emissionMultiplier, lightState.Intensity);
                 }
 				else if (_neonMaterialsFullColor[i].Location == VenueLightLocation.Left)
                 {
-					if (lightStateLeft.Color == null)
-					{
-						_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, _neonMaterialsFullColor[i].InitialColor);
-					}
-					else
-					{
-						_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, lightStateLeft.Color.Value);
-					}
+					_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, lightStateLeft.Color ?? _neonMaterialsFullColor[i].InitialColor);
 					_neonMaterialsFullColor[i].Material.SetFloat(_emissionMultiplier, lightStateLeft.Intensity);
                 }
                 else if (_neonMaterialsFullColor[i].Location == VenueLightLocation.Right)
                 {
-					if (lightStateRight.Color == null)
-					{
-						_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, _neonMaterialsFullColor[i].InitialColor);
-					}
-					else
-					{
-						_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, lightStateRight.Color.Value);
-					}
+					_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, lightStateRight.Color ?? _neonMaterialsFullColor[i].InitialColor);
 					_neonMaterialsFullColor[i].Material.SetFloat(_emissionMultiplier, lightStateRight.Intensity);
                 }
                 else if (_neonMaterialsFullColor[i].Location == VenueLightLocation.Front)
                 {
-					if (lightStateFront.Color == null)
-					{
-						_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, _neonMaterialsFullColor[i].InitialColor);
-					}
-					else
-					{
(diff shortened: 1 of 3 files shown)


Mekkabit:
Oh right something I forgot to note originally is that materials assigned to the new location lists that use HDR colors to look bright will be, not so bright when the manager takes control of them. I'm not sure how to keep the HDR color's intensity value on the manager's color so the way I was getting around it was with a modified shader to change to non-HDR color with a separate emission intensity.

EliteAsian123 (changes requested):
Hello, thank you for the PR, and sorry for the wait!

My recommendation for the `NeonLightManager` class would be to store each material as a part of a serializable struct with properties instead of creating an array for each neon light position. Here is an example of what a struct like that might contain:
```
[Serializable]
public struct NeonLightMaterial {
    public Material Material;
    public VenueLightLocation Location;

    [NonSerialized]
    public Color InitialColor;
}
```
Note that the last field in the struct would store the default color of each material and would be computed in `Awake`, much like how the default color fields are initialized now.

Then, you could have an array of these structs instead, which would simplify the code and also make the inspector side of things a little bit nicer to deal with. If you do this, however, make sure you keep the `_neonMaterials` field for compatibility purposes.

Thanks!

Mekkabit:
Ah it was my intention, actually- I hadn't even realized until yesterday that only one color per location actually got stored because I'd only ever had one color per location in my venue, until I added a new color to one field and got a little confused about why the color of some of my lights was suddenly changing when the venue was loaded in-game. This took, *quite* a while to figure out how to rewrite that way but I got it working again, finally.

Mekkabit:
Oh, hm, just discovered a new bug with this rewrite in that blackout and silhouettes_spotlight cues specifically will reset to default color and brightness, silhouettes without spotlights, and strobing both still work.

Mekkabit:
You know, I'd only just now noticed underscores got added into the light manager's lighting type names between me exposing the rest of the location light states and now. Don't know how or why but I can't remove them because the names without underscores aren't defined anymore. Absolutely baffled by that.

Mekkabit:
SHOULD be good to go now, finally, I have it running as it should from a fresh re-clone of the dev branch.

Assets/Script/Venue/Lights/NeonLightManager.cs
```diff
@@ -12,35 +12,96 @@ public class NeonLightManager : MonoBehaviour
...
+                {
+					_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, lightStateCenter.Color ?? _neonMaterialsFullColor[i].InitialColor);
+					_neonMaterialsFullColor[i].Material.SetFloat(_emissionMultiplier, lightStateCenter.Intensity);
+                }
+                else if (_neonMaterialsFullColor[i].Location == VenueLightLocation.Crowd)
+                {
+					_neonMaterialsFullColor[i].Material.SetColor(_emissionColor, lightStateCrowd.Color ?? _neonMaterialsFullColor[i].InitialColor);
+					_neonMaterialsFullColor[i].Material.SetFloat(_emissionMultiplier, lightStateCrowd.Intensity);
                 }
             }
```
wyrdough, line 104: I feel like this would be more readable as a switch statement

```suggestion
            for (int i = 0; i < _neonMaterialsFullColor.Length; i++) 
            {
                var neon = _neonMaterialsFullColor[i];
                
                switch (neon.Location)
                {
                    case VenueLightLocation.Generic:
                        neon.Material.SetColor(_emissionColor, lightState.Color ?? neon.InitialColor);
                        neon.Material.SetFloat(_emissionMultiplier, lightState.Intensity);
                        break;
                    case VenueLightLocation.Left:
                        neon.Material.SetColor(_emissionColor, lightStateLeft.Color ?? neon.InitialColor);
                        neon.Material.SetFloat(_emissionMultiplier, lightStateLeft.Intensity);
                        break;
                    case VenueLightLocation.Right:
                        neon.Material.SetColor(_emissionColor, lightStateRight.Color ?? neon.InitialColor);
                        neon.Material.SetFloat(_emissionMultiplier, lightStateRight.Intensity);
                        break;
                    case VenueLightLocation.Front:
                        neon.Material.SetColor(_emissionColor, lightStateFront.Color ?? neon.InitialColor);
                        neon.Material.SetFloat(_emissionMultiplier, lightStateFront.Intensity);
                        break;
                    case VenueLightLocation.Back:
                        neon.Material.SetColor(_emissionColor, lightStateBack.Color ?? neon.InitialColor);
                        neon.Material.SetFloat(_emissionMultiplier, lightStateBack.Intensity);
                        break;
                    case VenueLightLocation.Center:
                        neon.Material.SetColor(_emissionColor, lightStateCenter.Color ?? neon.InitialColor);
                        neon.Material.SetFloat(_emissionMultiplier, lightStateCenter.Intensity);
                        break;
                    case VenueLightLocation.Crowd:
                        neon.Material.SetColor(_emissionColor, lightStateCrowd.Color ?? neon.InitialColor);
                        neon.Material.SetFloat(_emissionMultiplier, lightStateCrowd.Intensity);
                        break;
                    default:
                        // Just in case a new location ever gets added
                        YargLogger.LogDebug("Unknown location for neon light");
                        break;
                }            
            }
```

wyrdough (changes requested):
One small readability change and I think this is good to merge

Assets/Script/Venue/Lights/NeonLightManager.cs
```diff
@@ -18,9 +19,10 @@ public class NeonLightManager : MonoBehaviour
         private Material[] _neonMaterials;
 		
 		[System.Serializable]
-		public class NeonFullColor {
+		public struct NeonFullColor {
 			public Material Material;
 			public VenueLightLocation Location;
+			public VenueSpotLightLocation SpotLocation;
```
wyrdough, line 25: Not sure how this leaked in here, but this doesn't exist in YARG yet...

wyrdough (changes requested):
Other than the one thing that shouldn't be here yet and would keep it from compiling, I'm happy with this

PR #1057, opened 2025-07-23
fix score box positioning after hud has been edited

wyrdough:
Somehow DraggableHudElement was on both the Score Canvas element and the ScoreDisplay prefab, so it has been removed from the prefab so the scorebox actually follows the canvas again.


theli-ua:
I blame yaml scene merges

PR #1061, opened 2025-07-30
Fix issue with xinput gamepad device binding

wyrdough:
This fixes the issue with CRKD Mode 1 default bindings not getting set.

We weren't waiting on the PromptGamePadMode dialog, so bindings were not getting added and the profile was not getting connected.

### DIFF
--- a/Assets/Script/Menu/ProfileList/ProfileView.cs
+++ b/Assets/Script/Menu/ProfileList/ProfileView.cs
@@ -46,9 +46,10 @@ namespace YARG.Menu.ProfileList
 
         public YargProfile Profile { get; private set; }
 
-        private ProfileListMenu _profileListMenu;
-        private ProfileSidebar _profileSidebar;
+        private        ProfileListMenu _profileListMenu;
+        private        ProfileSidebar  _profileSidebar;
 
+        private static bool            _xinputDialogShowing;
 
         public void Init(ProfileListMenu menu, YargProfile profile, ProfileSidebar sidebar)
         {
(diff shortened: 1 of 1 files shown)


Assets/Script/Menu/ProfileList/ProfileView.cs
```diff
@@ -160,6 +161,7 @@ public async UniTask<bool> PromptAddDevice()
                         // Some remappers and non-gamepad devices show up as XInput gamepads
                         if (device is XInputController xinput)
                         {
+                            xinputDialogShowing = true;
```
TheNathannator, line 164: This never gets set back to `false` when the dialog is closed.

I'd move this to `PromptGamepadMode`, and only set/unset it when the dialog actually gets opened/closed in that function, as the selected controller type *does* get cached.

TheNathannator (changes requested):

Assets/Script/Menu/ProfileList/ProfileView.cs
```diff
@@ -209,6 +221,7 @@ public async UniTask<bool> PromptAddDevice()
 
         private static async UniTask<GamepadBindingMode?> PromptGamepadMode(XInputController xinput)
         {
+            _xinputDialogShowing = true;
```
TheNathannator, line 224: This needs to be set *after* checking the cache, not before.

I'd recommend using a try-finally block for this to ensure that the flag is always unset upon leaving the function, regardless of any unforeseen exceptions or control flow:

```cs
try
{
    _xinputDialogShowing = true;
    var dialog = DialogManager.Instance.ShowList("Which kind of controller is this?");
    // the rest of the dialog code...
}
finally
{
    _xinputDialogShowing = false;
}
```

TheNathannator (dismissed):

TheNathannator:
the real fool is me lol, i mistook the local variable for a field

PR #1062, opened 2025-07-30
Added ability to collapse sort headers in the music library menu.

prgmatic:
Tried my best to follow existing code styles and put things in the correct spot. The header collapse state is stored in `SongContainer`, this could be wrong. `SortHeaderViewType` does not point back to the category it represents, so I'm using the `HeaderText` as the key for storing and looking up collapsed states. There is a fair bit of logic restoring the SelectedIndex to the entry you were on before a collapse/expand happens. Lemme know if anything needs changing. 

https://github.com/user-attachments/assets/ba5a19e3-8718-4150-9bf7-6fb7609248ef

### DIFF
--- a/Assets/Art/Menu/MusicLibrary/MusicLibraryIcons.png
+++ b/Assets/Art/Menu/MusicLibrary/MusicLibraryIcons.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:d689a1103ed1c772050aa7b54f6533b274acba81fa15f365d14c158fe75ffd92
-size 2600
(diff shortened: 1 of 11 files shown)


wyrdough:
Thanks for your PR! I think this is a great feature to have.

As you may already know, the current project leads are moving on to other things so we're in a state of transition at the moment and feature PRs are not being merged until after 0.13 is released in a couple of weeks. Once that happens I will get back to you on this.

Thanks for your patience.

wyrdough (changes requested):
I still need to think some more on whether or not having some of the logic in SongContainer is the right way to go here, but it should be easy enough to move that if necessary...

polson:
Closing this in favor of https://github.com/YARC-Official/YARG/pull/1152

Issue #1063, opened 2025-07-31
Request for localization and support for more practice sections


shadowflower64:
## Missing practice sections
I have recently noticed that not all documented sections are properly localized by YARG. Here is a list of them:

### From RBN Docs:
* `fade_in` - "Fade in" (note: [prc_fade-in] works, but the documented [prc_fade_in] does not)
* `kick_it` - "Kick It!"
* `spacey` - "Spacey part" (note: [prc_spacey_part] works, but [prc_spacey] does not; both are documented)
* `perc_break` - "Percussion break" (note: [prc_percussion_break] works, but the documented [prc_perc_break] does not)
* `build_up` - "Build up" (note: [prc_buildup] works, but the documented [prc_build_up] does not)
* `speedup` - "Speed up"
* `speed_up` - "Speed up" (note: both spellings of "speed up" appear in the docs, neither one is supported by YARG)

### From GuitarGame_ChartFormats:
* `aco_gtr_intro` - "Aco Gtr intro"
* `kybd_intro` - "Keyboard intro" (sic)
* `foreplay_intro` - "Foreplay intro"
* `longtome_intro` - "Long Time intro" (sic)
* `very_spacey` - "Very spacey"
* `noise_build` - "Noise build"
* `gtr_ostinato` - "Gtr ostinato"
* `speed_picking` - "Fast picking" (sic)

I think all of the above should be supported - if a charter uses the documentation to make a chart, they should assume that the documentation is correct, even if the section name is misspelled. Therefore, I think it should also work properly in YARG.

### Alternative names for existing sections
Apart from the documented sections, I would also like to request some more alternative spellings for existing sections. Note that these are technically not documented in RBN Docs or the GuitarGame_ChartFormats repo, but I think that they make sense, or are common enough to be supported.
* `orchestral_intro` - same as `orch_intro`
* `rhythm_gtr_enters` - same as `rhy_enters`
* `post-verse` - same as `postverse`, `pre-verse` is already implemented
* `post-chorus` - same as `postverse`, `pre-chorus` is already implemented as well

If the previously mentioned missing documented sections get added, I would also like to request the following aliases, as "more correct" spellings of the sections:
* `acoustic_gtr_intro` - same as `aco_gtr_intro`
* `long_time_intro` - same as `longtome_intro`
* `fast_picking` - same as `speed_picking`

### `ugc` practice sections
[...]

PR #1065, opened 2025-07-31
Further improvements to practice section parsing and localization

Frickitickitavi:
Addresses Issue #1063. Summary of changes below:

# Bug fixes in `IsSectionNumber`
The `IsSectionNumber` method had an off-by-one error that caused YARG to fail to localize practice sections that had an underscore as their third-to-last character (`fade_in`, `kick_it`, `build_up`, etc.). It was also missing a `break` statement in its `foreach` loop, causing unnecessary iterations. This changelist fixes both of these issues.

# Removal of `guitar->gtr` shortcut
Previously, the code substituted all instances of `guitar` with `gtr` early in `ParseSectionName`. This was meant to improve compatibility with `[section]`-based sections that are written in plain English, without needing excessive entires in the `_alternateSectionNames` dictionary due to the high number of sections that include `gtr`.

However, this resulted in `guitar` being substituted even in nonstandard section names, which should be left alone (e.g. `[section Look at my guitar!]` should not be rendered as "Look at my gtr!"). I've removed the upfront substitution and instead added entires for each `gtr` section to `_alternateSectionNames`. As a bonus, removing the `.Replace()` call improves the performance of the code, since we now handle its use case via a single dictionary lookup.

# Support for percentage-based `ugc_section` markers from RBN
Magma autogenerates practice sections in increments of 5, 10, or 20% (based on the length of the song) for songs without explicit sections. These sections take the form `prc_ugc_section_<size>_<start>`, where `size` is the interval and `start` is the low end of the range (e.g. `40%-60%` has a `size` of 20 and a`start` of 40).

While YARG performs a similar autogeneration process on charts that do not have any practice sections, it doesn't gracefully interpret these section markers in Magma-generated charts. These changes add generalized support for such sections, supporting any one- or two-digit values for `size` and `start`.

# Other new sections and alternate names
This PR adds support for various section names and alternate names that are requested in the issue, mostly those that are documented in GuitarGame_ChartFormats but not the RBN docs. It also adds a few new sections that are extrapolated from the new "Very spacey" counterpart to the "Spacey part" section, adding sections such as "Very heavy", "Very quiet", etc. to complement the "<_adjective_> part" sections.

# Bug fix for "Kick it!"
`_alternateSectionNames` previously incorrectly substituted `kick_it!` to `kick it` (with a space), preventing it from being properly localized. This PR fixes the substitute value to `kick_it`.


Assets/StreamingAssets/lang/en-US.json
```diff
@@ -486,11 +496,15 @@
...
+				"fade_out": "Fade out",
+				"speed_picking": "Fast picking",
+				"foreplay_intro": "Foreplay intro",
+				"longtome_intro": "Long Time intro",
 			},
 			"SectionFormats": {
 				"LetterSection": "{0} section",
-				"WithNumber": "{0} {1}"
+				"WithNumber": "{0} {1}",
+				"PercentageBased": "{0}% - {1}%",
```
shadowflower64, line 507: The JSON structure is broken here - commas should not be placed after the last element of an object

Frickitickitavi:
> I've tested all of the sections - including the new percentage sections. It looks like the bug mentioned before is fixed now and the sections are translatable. I have two more notes:
> 
>     1. `[prc_rhy_guitar_enters]` should be added as an alias to `[prc_rhy_enters]`.
>        
>        * for context, these are already working aliases: `[prc_rhy_gtr_enters]`, `[prc_rhythm_gtr_enters]`, and `[prc_rhythm_guitar_enters]`. `[prc_rhy_guitar_enters]` seems like an obvious one to add.
> 
>     2. Currently the section `[prc_speed_picking]` uses the translation key `speed_picking`. `[prc_fast_picking]` also works, but is aliased to `speed_picking`. Maybe the `[prc_fast_picking]` should become the "default" section name, and then `[prc_speed_picking]` should be an alias to that? I think that makes sense. The displayed name is "Fast Picking" after all. In my mind, this is similar to other documented sections like `[prc_sctrach_break]` being aliased to `scratch_break` or `[prc_perc_break]` being aliased to `percussion_break`, and not the other way around.
> 
> 
> I also like the addition of the new "very [blank]" sections. I added those to the spreadsheet, and also added all new aliases.

`rhy_guitar_enters` is a good catch; that one slipped past me because I was only referencing the JSON file when creating all the explicit `guitar` substitutions.

Agreed on `speed` vs. `fast`. All feedback should now be addressed

PR #1066, opened 2025-07-31
use unique names when creating new profile

mikikiv:
When creating multiple profiles, it would be helpful for each account to not be created with the same name by default

### DIFF
--- a/Assets/Script/Menu/ProfileList/ProfileListMenu.cs
+++ b/Assets/Script/Menu/ProfileList/ProfileListMenu.cs
@@ -95,7 +95,7 @@ namespace YARG.Menu.ProfileList
             }
         }
         
-        private string GetUniqueProfileName(string profileName)
+        private static string GetUniqueProfileName(string profileName)
         {
             int count = 1;
             var existingNames = PlayerContainer.Profiles.Select(p => p.Name);
--- a/Assets/Script/Menu/ProfileList/ProfileListMenu.cs
+++ b/Assets/Script/Menu/ProfileList/ProfileListMenu.cs
@@ -94,12 +94,27 @@ namespace YARG.Menu.ProfileList
                 _navigationGroup.AddNavigatable(go);
             }
         }
+        
+        private string GetUniqueProfileName(string profileName)
+        {
+            int count = 1;
+            var existingNames = PlayerContainer.Profiles.Select(p => p.Name);
+            
+            string newName;
+            do
+            {
+                newName = $"{profileName} {count}";
+                count++;
+            } while (existingNames.Contains(newName));
+            
+            return newName;
+        }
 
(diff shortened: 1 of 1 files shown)


Assets/Script/Menu/ProfileList/ProfileListMenu.cs
```diff
@@ -94,12 +94,27 @@ private void AddListGroup(string header, IEnumerable<YargProfile> profiles)
                 _navigationGroup.AddNavigatable(go);
             }
         }
+        
+        private string GetUniqueProfileName(string profileName)
```
wyrdough, line 98: This should be made static

```suggestion
        private static string GetUniqueProfileName(string profileName)
```

wyrdough (commented):
You may want to consider whether a new profile without any conflict should be called "New Profile 1" or just "New Profile". Personally, I'd rather only have the postfix on profiles when there is actually a conflict, but that's just a matter of taste, and there is a good argument to be made for keeping the code simpler, so it's really up to you whether you want to keep it as-is or change it.

mikikiv (commented):
only append a number if the profile name already exists

Assets/Script/Menu/ProfileList/ProfileListMenu.cs
```diff
@@ -95,11 +95,31 @@ private void AddListGroup(string header, IEnumerable<YargProfile> profiles)
             }
         }
 
+        private static string GetUniqueProfileName(string profileName)
+        {
+            var existingNames = PlayerContainer.Profiles.Select(p => p.Name);
```
wyrdough, line 100: We normally try to avoid LINQ because of garbage.

Assets/Script/Menu/ProfileList/ProfileListMenu.cs
```diff
@@ -95,11 +95,31 @@ private void AddListGroup(string header, IEnumerable<YargProfile> profiles)
...
+
+            if (!existingNames.Contains(profileName))
```
wyrdough, line 102: LINQ returns an IEnumerable that gets fully enumerated every time you call `Contains()` on it, so generally speaking if you're going to use it multiple times you should do something like `var names = existingNames.ToArray()` and call `Contains()` on `names` instead.

wyrdough (approved):
I noted a couple of things to keep in mind for the future, but since they're pointlessly minor in this specific context I'm going to go ahead and merge this as-is.

Specifically, LINQ isn't really a huge issue outside of gameplay since we force the garbage collector to run before loading in to gameplay and the multiple enumeration isn't much of an issue because it's only happening twice and the list of profiles is all but certain to remain sufficiently small that the extra time spent evaluating it twice is small enough to make zero difference in the menu.

PR #1076, opened 2025-08-06
Adjust rectangle theme note size, position, and rotation

grishhung:
Resized all rectangle theme notes slightly. See the following changes:

- Cymbal notes now maintain their triangular shape while moving towards the strikeline.
- Most note types are now vertically to cover more of the fret when at the strikeline.
- The front edges of kick notes and tom notes now align for unified judgement.
- Tom notes now sit on top of the kick note rather than being skewered by them.
- 5L cymbal ghost note is no longer randomly misaligned to the left.
- Small notes are standardized as 4/5 of the normal note height for better readability.
- Cymbal notes no longer have metallic shine (fixes graying and discoloration).
- Kick note is now slightly thinner to allow note to visibly sit on top.

https://github.com/user-attachments/assets/1813336b-d627-4958-91ed-22a0e3b57243

https://github.com/user-attachments/assets/ccdd102b-d1eb-494c-8cf0-55c14a3f4797

### DIFF
--- a/Assets/Art/Materials/Gameplay/Notes/Rectangular/CymbalMetal.mat
+++ b/Assets/Art/Materials/Gameplay/Notes/Rectangular/CymbalMetal.mat
@@ -107,7 +107,7 @@ Material:
     - _ReceiveShadows: 1
     - _SHINE: 0
     - _ShineAmount: 0
-    - _Smoothness: 0.5
+    - _Smoothness: 0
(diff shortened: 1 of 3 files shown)


grishhung:
Added flame effect to cymbals (was there before but got removed at some point)


https://github.com/user-attachments/assets/6662d517-64a8-450f-8905-e47414cdc041

PR #1079, opened 2025-08-07
add button to add/remove song from favorites to score screen

4dispel:
The button changes based on if the song is already a favorite. This is my first time working with C#/Unity, but it seems to work.

image
image

### DIFF
--- a/Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
+++ b/Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
@@ -19,6 +19,7 @@ using YARG.Menu.Navigation;
 using YARG.Menu.Persistent;
 using YARG.Scores;
 using YARG.Song;
+using YARG.Playlists;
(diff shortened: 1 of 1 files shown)


4dispel:
remove/add from favorites not playlist, sorry

wyrdough:
Thanks for the PR! 

As you may already know, the project is in a state of flux at the moment, so we're not merging PRs with new features just at the moment, but we will get back to you when the merge window opens again in a couple of weeks.

I did take a quick look and I didn't see anything at a glance that would make this unmergeable, so I don't think any significant changes will be required at that time, but we'll let you know when the time comes.

Thanks again for the PR and your patience.

wyrdough (changes requested):
Just a minor formatting nitpick. Once you fix that, I can merge this.

(The reason I/we care is just that breaking it up makes it easier to see what's happening at a glance)

wyrdough (approved):

PR #1080, opened 2025-08-07
Fix practice accuracy not resetting when song speed is changed

### LINKED ISSUE #640: [Nightly] In practice, speed up or down doesn't reset your accuracy percentage.
[Video Example](https://streamable.com/5y5wug) on the latest nightly build.

In practice mode, if you select a section, set the speed low, play it, and then right before the section resets, speed it up to full speed, the game won't reset your current accuracy percentage, and so that stat will then be set as your "best" accuracy percent after the section resets.

This is concerning as this could easily be used to fake section FC screenshots. Really current accuracy and best accuracy should both be reset on speedup/slowdown, not sure why it's just best accuracy.

santorfo:
Fixes #640 

In PracticeHud.cs, I added a flag _speedChanged, that is temporarily set to true when the speed is changed in the middle of the practice section.
When the speed change happens, in ResetStats(), the current player.NotesHit value is stored in the _offsetNotesHit variable, and as long as the _speedChanged flag is set to true, this variable is used in the Update() function to calculate the accuracy percentage correctly, by only using the amount of notes hit after the speed was changed.

### DIFF
--- a/Assets/Script/Gameplay/HUD/Practice/PracticeHud.cs
+++ b/Assets/Script/Gameplay/HUD/Practice/PracticeHud.cs
@@ -35,6 +35,9 @@ namespace YARG.Gameplay.HUD
 
         private int _currentSectionIndex;
 
+        private bool _speedChanged = false;
+        private int  _offsetNotesHit = 0;
+
         protected override void GameplayAwake()
         {
             _sections = Array.Empty<Section>();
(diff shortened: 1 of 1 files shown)


wyrdough:
Genuine question since I don't really use practice mode and so don't know what exactly the desired behavior here actually is, but should we also be applying an offset to `TotalNotes` as well?

PR #1083, opened 2025-08-10
Save "AverageMultiplier" and "NumPauses" to Replay Stats

raphaelgoulart:
What it says on the tin; it'll depend on the similarly named YARG.Core PR


raphaelgoulart:
Closed for the reason mentioned here: https://github.com/YARC-Official/YARG.Core/pull/268

PR #1085, opened 2025-08-12
[perf] Use highways as stencil when rendering venue

theli-ua:
This makes it so that venue does not render parts that are obscured by highways.
This bumps my fps at 2560x1440 when using Fractal77 with 2  highways from 88 to 117 fps
It doesn't have such a pronounced effect with less demanding venues though obviously

### DIFF
--- a/ProjectSettings/GraphicsSettings.asset
+++ b/ProjectSettings/GraphicsSettings.asset
@@ -37,7 +37,7 @@ GraphicsSettings:
   - {fileID: 10753, guid: 0000000000000000f000000000000000, type: 0}
   - {fileID: 10770, guid: 0000000000000000f000000000000000, type: 0}
   - {fileID: 10783, guid: 0000000000000000f000000000000000, type: 0}
-  - {fileID: 4800000, guid: 358932d403fb17de0b26c0a868118cb0, type: 3}
+  - {fileID: 4800000, guid: d401e57306d63ad8c95aebbec69297e9, type: 3}
   m_PreloadedShaders: []
   m_PreloadShadersBatchTimeLimit: -1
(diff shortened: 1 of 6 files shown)


ProjectSettings/GraphicsSettings.asset
```diff
@@ -37,7 +37,7 @@ GraphicsSettings:
   - {fileID: 10753, guid: 0000000000000000f000000000000000, type: 0}
   - {fileID: 10770, guid: 0000000000000000f000000000000000, type: 0}
   - {fileID: 10783, guid: 0000000000000000f000000000000000, type: 0}
-  - {fileID: 4800000, guid: 358932d403fb17de0b26c0a868118cb0, type: 3}
+  - {fileID: 4800000, guid: d401e57306d63ad8c95aebbec69297e9, type: 3}
```
wyrdough, line 40: Did you intend to remove HighwayBlit?

theli-ua: Yes, since its not actually used. I replaced it with the shader that creates a stencil

wyrdough: Sorry, had a braino, I thought that HighwayBlit was one of your new shaders, lol

theli-ua: > Sorry, had a braino, I thought that HighwayBlit was one of your new shaders, lol

It actually was at some point when I moved fade and curve into separate pass. But I stopped using it when I moved it into PP and later into the complete overhaul

theli-ua: The history was: 
* getting things to render as opaque - for this had to move fading into UI curve shader
* extracted into a separate pass/shader (highway blit)
* improvements
* ignore/stagnation for half a year
* complete rework at which point use of that shader dropped

PR #1089, opened 2025-08-14
Five-lane keys

Frickitickitavi:
[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/273)

Implements the Five-Lane Keys engine, accessible via the `Keys (Pro and Five-Lane)` profile type (previously called `Pro Keys`).

This change means that there is no longer a strict mapping from `Instrument` to `GameMode`, because the five 5L instruments (lead, bass, rhythm, coop, and keys) can now each be associated with either the `FiveFretGuitar` or `Keys` (formerly `ProKeys`) game mode. This has the following impacts:
* `Instrument.ToGameMode()` is now nonsensical. The function has been renamed to `ToNativeGameMode` and is used only in cases where the game mode is irrelevant (such as in the `SongLoader`, where they're all loaded as `GuitarNote`s no matter what; the distinction comes into play later, in the engines)
* We previously serialized a profile's instrument, but not its game mode, and derived the latter from the former on deserialization. We now explicitly serialize the game mode in addition to the instrument, which has caused a serialization version bump

The 5L Keys engine is implemented as a copy-paste of everything in the Pro Keys engine except glissandos. This includes subtleties like chord staggering and fat-finger forgiveness; I'm not sure if these should be retuned for 5L, or if fat-finger forgiveness is even supposed to be a thing there, but for now everything is 1:1 with Pro.

The relationship between engines, players, and visual prefabs forced me to create several 5LK-specific duplicates of 5F Guitar prefabs. As it stands, this feels unfortunate because 5LK's visuals are a strict subset of 5F's and could theoretically be single-sourced. However, I think it's worth looking into whether we want to give 5L Keys any special visual treatment such as unique gems, so that a watcher can tell which engine is being used without waiting to see if there are any HOPOs or taps on the highway.

Open notes in the 5L Keys engine are treated as just another note, with its own binding and everything. Open chords function just like any other chord. We might want to offer the ability to treat opens as "omni-notes" in the 5L Keys engine and/or the ability to show a dedicated purple lane on the left side of the 5LK highway (either always, or only for charts that contain opens), but I consider those to be post-MVP goals.


wyrdough:
So... I tried to test this and a couple of problems popped out.

1. Previous PK and 5 fret bindings are lost when connecting a profile again. It's like setting everything up from scratch, down to having to select the device again
2. This may just be a "me" problem, but some change here is causing exceptions to be thrown in the input system. When binding the 5 lane actions, I get the following:

```Exception 'MissingReferenceException' thrown from state change monitor 'ButtonBinding' on 'MidiNote:/MidiDevice/note065'```

```MissingReferenceException: The object of type 'Image' has been destroyed but you are still trying to access it.
Your script should either check if it is null or you should not destroy the object.
UnityEngine.EventSystems.UIBehaviour.IsActive () (at Library/PackageCache/com.unity.ugui@1.0.0/Runtime/EventSystem/UIBehaviour.cs:28)
UnityEngine.UI.Graphic.SetVerticesDirty () (at Library/PackageCache/com.unity.ugui@1.0.0/Runtime/UI/Core/Graphic.cs:295)
UnityEngine.UI.Graphic.set_color (UnityEngine.Color value) (at Library/PackageCache/com.unity.ugui@1.0.0/Runtime/UI/Core/Graphic.cs:156)
YARG.Menu.ProfileInfo.ButtonDisplay.set_IsPressed (System.Boolean value) (at Assets/Script/Menu/ProfileInfo/EditBinds/ButtonDisplay.cs:25)
YARG.Menu.ProfileInfo.SingleMidiNoteBindView.OnStateChanged (System.Single state) (at Assets/Script/Menu/ProfileInfo/EditBinds/SingleBindViews/SingleMidiNoteBindView.cs:49)
YARG.Input.SingleBinding`1[TState].InvokeStateChanged (TState state) (at Assets/Script/Input/Bindings/ControlBinding.cs:204)
YARG.Input.SingleButtonBinding.UpdateState (System.Double time) (at Assets/Script/Input/Bindings/ButtonBinding.cs:164)
YARG.Input.ControlBinding`2[TState,TBinding].UnityEngine.InputSystem.LowLevel.IInputStateChangeMonitor.NotifyControlStateChanged (UnityEngine.InputSystem.InputControl control, System.Double time, UnityEngine.InputSystem.LowLevel.InputEventPtr eventPtr, System.Int64 monitorIndex) (at Assets/Script/Input/Bindings/ControlBinding.cs:530)
UnityEngine.InputSystem.InputManager.FireStateChangeNotifications (System.Int32 deviceIndex, System.Double internalTime, UnityEngine.InputSystem.LowLevel.InputEvent* eventPtr) (at Library/PackageCache/com.unity.inputsystem@1.7.0/InputSystem/InputManagerStateMonitors.cs:380)
UnityEngine.DebugLogHandler:Internal_LogException(Exception, Object)
YARG.Logging.Unity.CustomUnityLogger:LogException(Exception, Object) (at Assets/Script/Persistent/Logging/Unity/CustomUnityLogger.cs:120)
UnityEngine.InputSystem.LowLevel.<>c__DisplayClass7_0:<set_onUpdate>b__0(NativeInputUpdateType, NativeInputEventBuffer*)
UnityEngineInternal.Input.NativeInputSystem:NotifyUpdate(NativeInputUpdateType, IntPtr)
```

That said, it did play and the basic functionality seemed to work perfectly, assuming that we want to allow anchoring. (aside from the annoying assert spam that was pre-existing in the PK engine since the engine-fixes merge, but that's not a you problem)

wyrdough:
> On the other hand, if you mean that you could hit a G note, then continue holding the G while you hit a later R, then that _is_ correct and intended behavior. The keys engine shouldn't care what else you're already holding when you hit a note. The distinction is that, unlike with proper anchoring on a guitar engine, it _also_ doesn't care what you're holding when you _release_ a key, because there's no pull-off behavior.

That is indeed what I was referring to. Doesn't seem like it should do that to my simple mind, but if that's the intended behavior, ok...
 
> Separately, I'm surprised that profiles didn't carry forward. I'm guessing it's due to one of these three changes, although I wouldn't have expected any of them to have this effect. Any hunch what it could have been?
> 
> 1. The fact that I added new bindings to the profile type
> 2. The fact that the `GameMode.ProKeys` enum value was renamed to `GameMode.Keys` (actual integer value shouldn't have changed though)
> 3. The fact that I changed the UI text name of the profile from `Pro Keys` to `Keys (Five-Lane and Pro)` (just a localization file change, so this seems least likely of all)

I don't think it's option 1, because the need to reselect the device also happens for guitar profiles. Option 3 is all but impossible since the bindings JSON doesn't refer to profile names at all, it identifies them by UUID. It's possible it has nothing to do with your PR at all and it happened due to something else I was testing this morning. I'll try to recreate the problem and better nail down what triggered it.

Frickitickitavi:
> That is indeed what I was referring to. Doesn't seem like it should do that to my simple mind, but if that's the intended behavior, ok...

Short answer: Yes, that's 100% correct and intentional

Long answer: This is consistent with the PK engine, and both the Pro and 5L Keys engines in RB3. It's also just kind of a necessary bit of leniency - if someone is playing some fast shredding, they shouldn't need to meticulously lift the previous note in order for the next one to be valid. 

You might think that we could solve that by just introducing a short grace period, but even then, it's not the type of thing we want to punish players for. We don't punish them for holding their sustains too long on guitar, unless they re-strum them. Likewise, we shouldn't punish players for just holding down a note too long - it's when they _play_ it again that we have a problem. Much of the time, keys is playing an instrument that has a limited sustain time (like a piano, xylophone, harpsichord, etc), so it's perfectly plausible that a player who's still holding down an old note wouldn't actually create any dissonance by continuing to do so under a new note.

wyrdough (changes requested):
I don't see any issues with any of the things you changed. Just need to deal with the error in deserializing bindings when old PK profiles are connected that I mentioned on Discord and I think we'll be good on the YARG side.

PR #1091, opened 2025-08-15
Add modifier to disallow star power overlap

thearst3rd:
YARG.Core PR: https://github.com/YARC-Official/YARG.Core/pull/274

> Hey! I'm not sure if this is a feature that anybody wants, but since YARG is very flexible and supports features from a variety of GH+RB games, I thought it'd be nice to add to that.
>
> This PR adds a modifier to prevent earning a star power phrase while star power is already active, just like it works in the early Guitar Hero games. Hitting a SP note while SP is active will kill that phrase, but if SP ends before the first note of the phrase is hit (even if it's already on the fretboard) then SP works like normal.

Some things that need to be considered:

* All the stuff in my YARG.Core PR
* The list of modifiers on Guitar is now too long to see all at once. Strumming through it should make it automatically scroll the scrollbar
* The modifier will need an icon, right now it's just showing up on the endscreen as a pure white square :P 

Video demonstrating functionality: Left is normal and right has the modifier

https://github.com/user-attachments/assets/e3ccf02c-00ff-4c3c-9ef6-9418aa304f59

> Regardless of if this is something we want in the game, this was a nice warmup for me to get used to using Unity and navigating the codebase. Thank you all for your hard work developing YARG, I've been loving it so far :)

### DIFF
--- a/Assets/Script/Gameplay/Player/DrumsPlayer.cs
+++ b/Assets/Script/Gameplay/Player/DrumsPlayer.cs
@@ -69,8 +69,10 @@ namespace YARG.Gameplay.Player
 
             if (!Player.IsReplay)
             {
+                bool noStarPowerOverlap = (Player.Profile.CurrentModifiers & Core.Game.Modifier.NoStarPowerOverlap) != 0;
+
                 // Create the engine params from the engine preset
-                EngineParams = Player.EnginePreset.Drums.Create(StarMultiplierThresholds, mode);
+                EngineParams = Player.EnginePreset.Drums.Create(StarMultiplierThresholds, noStarPowerOverlap, mode);
             }
             else
             {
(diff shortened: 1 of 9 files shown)


wyrdough:
Thanks for the PR! I haven't had time to look at it fully in depth yet, but will get to it soon. My initial read only had a couple of things stick out at me that should probably be addressed.

* Please make this an engine preset parameter rather than a modifier. (I could be swayed on this point if you have a good argument for keeping it as a modifier, but my preference would be engine preset)
* What happens visually in the case where the player activates SP after the notes have already been spawned? I did not see anything that addresses this, but it was a pretty quick scan, so I may have missed something.

thearst3rd:
Thanks for the review :) 

* I wasn't sure which was better and just went with modifier, I have no strong argument for keeping it so I will gladly change it 👍
* Visually it does change the color, I adjusted `UpdateElement` to also detect changes in the star power activation state and call `OnStarPowerUpdated`. The comment above the `OnStarPowerUpdated` seems to indicate that this is possible for some classes but I couldn't find any instances of it

On that second point actually, I didn't consider the circular theme which changes models, I just tested it and it is broken, will need to fix this as well

PR #1096, opened 2025-08-22
Add activation customization

grishhung:
## **Depends on https://github.com/YARC-Official/YARG.Core/pull/276!**

---

## Changes

- Makes the new default SP activation CH-style (rightmost-only required).
- Adds a setting to revert to the old (all notes required) activation style.
- Moves activation note flagging to `DrumPlayer`.
- Adds a fill strip up to the rightmost gem for CH-style activations.
- Makes activation notes blink to the beat.

https://github.com/user-attachments/assets/c95a12c9-006b-448c-beca-5a7963a8dbbe

## Known Issues

- Kick activation fill lanes currently do not respect the curve; this will be handled by Anton.
- In pro drums split view, notes and activation fill lanes get chopped up. Cause is same as the above.
- Green is currently used for all activations, which may be confusing (parity with RB4, but not RB3).


Assets/Script/Gameplay/Visuals/TrackElements/Drums/FourLaneDrumsNoteElement.cs
```diff
@@ -115,5 +128,11 @@ protected override void UpdateColor()
             // Set the metal color
             NoteGroup.SetMetalColor(colors.GetMetalColor(NoteRef.IsStarPower).ToUnityColor());
         }
+
+        private static int GetColorFromPulse(int color, float pulse)
```
wyrdough, line 132: See comment on same method in `FiveLaneDrumsNoteElement`

Assets/Script/Gameplay/Visuals/TrackElements/TrackEffectElement.cs
```diff
@@ -376,6 +390,51 @@ protected void RescaleForZ()
...
+        {
+            // Hide lead up if the chosen star power activation type does not need it
+            if (Player.Player.Profile.StarPowerActivationType is StarPowerActivationType.AllNotes
+                or StarPowerActivationType.Freestyle)
+            {
+                return 0f;
+            }
+
+            // Kick lead-ups and unknown indices should not be shrunken
+            return EffectRef.FillLane <= 0 || EffectRef.FillLane > EffectRef.TotalLanes
```
wyrdough, line 433: I think this would be better as an explicit if..else. It isn't immediately obvious what it's doing the way it is now.

Assets/Script/Gameplay/Visuals/TrackElements/TrackEffectElement.cs
```diff
@@ -1,13 +1,19 @@
 using System;
+using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Serialization;
+using YARG.Core;
+using YARG.Core.Game;
 using YARG.Gameplay.Player;
 using YARG.Helpers.Extensions;
+using Color = System.Drawing.Color;
```
wyrdough, line 9: Pot calling the kettle black here given the number of other unused `using`s here, but this isn't needed.

wyrdough (changes requested):

Assets/Script/Gameplay/Visuals/TrackElements/TrackEffectElement.cs
grishhung: That would be a lot more readable. Fixed.

PR #1112, opened 2025-08-29
Re-add fret glow effects

grishhung:
# Notes

https://github.com/user-attachments/assets/0b175ccc-6bbf-42d6-934a-94c1c70e487f

- Add fret glow effects, as shown above, with hit flames now using the normal setting.
- Changed drum hit animations to wobble for cymbals but bounce for pads (like they'd act in real life).
- All instruments have a fade time of 100ms, except for drums which is doubled to 200ms.
  - Drums lacks sustains and looks hollow otherwise. 

## Known issues

- Open and kick flames are not adjust and will need to be replaced.
- The cymbal animation has not been touched, but because of the sharper flame drop off, the wobble appears somewhat lengthy. This may need to be reduced in the future.


grishhung:
Yeah, it cutting off sharply was the intention. I can probably experiment with adding a slight drop off as well, though.

Assets/Script/Helpers/Authoring/EffectLight.cs
```diff
@@ -78,12 +91,13 @@ public void SetColor(Color c)
...
         {
             _light.intensity = _initialIntensity;
+            _timeRemaining = _totalDuration;
             _playing = true;
         }
 
         public void Stop()
         {
-            if (_mode != Mode.FadeOut)
+            if (_mode == Mode.Wavy)
```
wyrdough, line 100: I assume this is some bug you're fixing?

Assets/Script/Helpers/Authoring/EffectLight.cs
```diff
@@ -52,6 +58,13 @@ private void Start()
 
         private void Update()
         {
+            // Normal mode
+            if (_mode == Mode.Normal)
```
wyrdough, line 62: probably want to take a cue from FadeOut below and add `&& _light.intensity > 0f` here. (Or maybe `&& _playing`? I'm not actually sure)

wyrdough (changes requested):
I haven't actually looked at the visual aspect yet, but here's a few minor code changes you may want to consider...

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -449,11 +367,11 @@ private void OnPadHit(DrumsAction action, bool wasNoteHit, float velocity)
             {
                 if (isDrumFreestyle)
                 {
-                    _fretArray.PlayHitAnimation(fret - 1);
+                    _fretArray.PlayHitAnimation(fret);
```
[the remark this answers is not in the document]
grishhung: Huh. I had done this before but it must have gotten lost when I did my merge earlier. Will add this back.

Assets/Script/Helpers/Authoring/EffectLight.cs
grishhung: Added this in.

wyrdough:
I'm going to go ahead and merge this, but at some point we probably do need to bump the brightness of the SP bar up a bit when it is in the less than half full state.

PR #1113, opened 2025-08-30
add vocal character animation states

4dispel:
adds MouthOpen and MouthClose animation events to the venue character script

this may very well be incomplete as I'm not sure what everything in the character code is actually for.
i didn't put in Start/Stop Animation in process vocals because i didn't see what the actual use of those is considering the tempo gets updated anyways in the character manager update method and i didn't see _isAnimating used for anything else

### DIFF
--- a/Assets/Script/Venue/Characters/CharacterManager.cs
+++ b/Assets/Script/Venue/Characters/CharacterManager.cs
@@ -20,7 +20,7 @@ namespace YARG.Venue.Characters
         private readonly Dictionary<VenueCharacter.CharacterType, VenueCharacter> _characters = new();
 
         // Ugh, the different note types ruin me again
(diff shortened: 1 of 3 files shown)


wyrdough:
Hey, this looks like it should do the business, but presently lacking a character to test it with I'm not sure I want to merge it just yet. 

And yeah, there's some stuff in the character animation code that isn't really there for any good reason any more that just hasn't yet been stripped out and some other stuff that's only still there because I didn't want to break earlier versions of the venues with characters until people have had a chance to update. It remains a work in progress...the hope is that having something to work against will get some more venue authors looking at it so they can provide feedback on what more should be added, what can be done differently to make it easier on them, etc.

4dispel:
closing this in favor of #1330

PR #1114, opened 2025-08-31
Chart/metadata-determined vocal scroll speed and MVP static vocals

Frickitickitavi:
[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/280)

Various changes to vocal parsing and visual mechanics to support dynamic vocal scroll speed and static vocals. 

# Dynamic Vocal Scroll Speed
Vocal scroll speed is no longer a vocal profile setting, and is now determined by either chart metadata or the chart itself. For INI-based charts, the charter can supply a new `vocal_scroll_speed` flag to change the chart's scroll speed. The field is interpreted as a percentage, so `100` represents the default scroll speed (equivalent to the `5.0` profile setting value).

RB CON files also contain vocal scroll speed values. The default is `2300`; I'm not clear on what the units are, but it appears to be some kind of unit of time, because lower values create higher scroll speeds. If a value other than `2300` is present, the parser records the scroll speed as the ratio between the provided value and `2300`. Because Magma automatically includes the value `2300` unless otherwise specified, I've designed the parser to treat `2300` as if no value were provided.

If no value is provided (including CONs with `2300`), the game runs a check at song startup to see whether the song's lyrics are so fast that they'll get woefully out of sync with the notes. The calculation is fairly rough, but it should kick into gear if the song's lyrics get more than about half a screen's width out of sync. It calculates the single largest misalignment in the song, and the severity of that alignment informs how much it increases the speed by (in discrete chunks).

Metadata-provided values (ignoring CONs with `2300`) are always honored, even if the chart ends up suffering from lyric misalignments. They are also honored in Static Vocals Mode, although Static Vocals doesn't bother to run the misalignment calculation if no value is provided.

# Static Vocals Mode
Besides spacing out fast lyrics, the other benefit of the old vocal profile track speed setting was as an accessibility option for people who have difficulty reading fast-moving words. With that option removed, we need a different solution. As such, I've implemented a basic version of RB3's Static Vocals mode (not to be confused with the fully-paginated version used in RB2)

Static Vocals is available as a new global setting under Graphics, right above the Use Three Lyric Lanes option. When enabled, lyrics queue up along the lyric lane and shift phrase-wise rather than scrolling in sync with their respective tubes. To support this, I implemented a new `VocalsStaticPhraseElement`, which represents an entire phrase at once.

Because of some subtleties regarding how Static Vocals mode behaves in Harmonies mode, we need to separately track _scoring_ phrases from _lyric_ phrases. This resulted in some parsing changes, and the song loader now needs to separately compile lists of each.

In two-lyric-lane mode, this implementation correctly handles the HARM2/3 merging process, including the `$` modifier and automatically merging simultaneous-and-identical lyrics.

Because this is implemented in support of the primary goal of dynamic vocal scroll speed, it glosses over some of the finer details of Static Vocals behavior. The following aspects are _not_ implemented:
* Showing vocal activation zones on the lyric lane (not implemented for scrolling mode either)
* Supporting mid-phrase lyric shifts placed by the charter
* Attempting to align simultaneous upcoming phrases with each other

# Removed Vocal Profile Settings
In addition to replacing the Note Speed setting for vocal profiles, this changelist also removes a few other settings that had no effect in the first place. Some, like Highway Length, will likely never come back, while others, like Color Profile, can be reintroduced once they have a function.


Assets/Script/Gameplay/Player/Vocals/VocalTrack.Spawning.cs
```diff
@@ -162,6 +116,103 @@ private void SpawnLyricsInPhrase(PhraseNoteTracker tracker, int harmonyIndex)
...
+                case StaticLyricShiftType.None:
+                    break;
+                case StaticLyricShiftType.PhraseToGap:
+                {
+                    var leftmostPhraseElement = queue.Dequeue();
+                    var leftShift = leftmostPhraseElement.Width;
+
+                    foreach (var remainingPhrase in queue)
+                    {
+                        remainingPhrase.transform.DOLocalMoveX(remainingPhrase.transform.localPosition.x - leftShift, STATIC_LYRIC_SHIFT_DURATION);
```
Frickitickitavi, line 140: This, and other `DOLocalMoveX` calls in this method, are at the heart of the debugger behavior I was talking about in the team chat.

_Scrolling_ lyrics, like all other track elements (vocals and otherwise), derive their position every update based on the current time, which means that they instantly snap into place upon unpausing the debugger.

For static lyrics though, we can't determine position strictly as a function of time, because static lyrics shunt each other around. In theory, we could still derive the position as a function of both time _and_ the full list of phrases, but it's much easier to just cue the phrases to shuffle down the line when they're told to. The only unfortunate side effect (afaik) is that, if you pause the debugger, phrases start missing their cues. When you unpause the debugger, they don't have any sense of what they missed, so everything goes out of alignment

PR #1115, opened 2025-09-01
Allow scrolling the score screen results without a mouse

thomeval:
This PR adds support for horizontal scrolling in the Score Screen without having to use a mouse. This is typically needed when 5 or more players are present. To scroll the score card list, any player can use their Menu Left/Menu Right inputs (typically the D-Pad). 

This PR also adds support for *vertical scrolling* on the same screen. When pressing the Menu Up / Menu Down inputs, this will scroll the stats on that player's score card. Note that this isn't typically useful at the moment, since under normal circumstances, all stats already fit on the available space with no scrolling needed. However, should more stats be added in future, this will become essential.

~Note to the reviewer: The code in ScrollScoreCard() is far from ideal, but it works. If you can think of a better way of handling this, please let me know.~ Resolved. Thanks!


wyrdough (changes requested):
So I haven't actually tested the suggested changes, but it ought to work. (assuming I didn't miss a spot, lol)

wyrdough:
Oh, one more todo: Add your new string(s) to `en-US.json`.

I'd probably make them all `Menu.Common.Scroll` and have that translate to `Scroll` or `Scroll Left/Right/Up/Down` or whatever. All the directions get coalesced into a single item on the help bar.

PR #1117, opened 2025-09-03
Use mesh for the circular hopo halo

theli-ua:


theli-ua:
image
screenshot with bloom disabled (still seem to be perfectly readable)

theli-ua:
Just for comparison this is nightly 
image

theli-ua:
Just to clarify - mesh is generated at build time, not at runtime

wyrdough:
So the only obstacle to merging this is that in motion it looks more like there's a glowing pancake on top of the note than a halo.

Well, that and the tiny nitpick of pls PascalCase the filenames so they're consistent with everything else.

If there's a reasonably simple way to only render the halo mesh if bloom is disabled, I'll merge that so that people with bloom disabled aren't completely screwed and then we can fix up the art at some later point so bloom and no bloom can both use the mesh.

theli-ua:
I'd say only enabling it for non-bloom case would unnecessarily complicate this. Perhaps there is a way to make it not look like pancake? 

Do you think adding an outline to the halo could help here? (I'm not exactly sure what exactly a pancake concern is tbh)

PR #1127, opened 2025-09-12
Use Harmony Instrument Icons on Fail Meter / Active Player List

thomeval:
This PR adds support for Harmony Player Icons, both to the Active Players List (at the top of the screen), and the new Fail Meter. This means that the game will display unique instrument icons for each Harmony vocalist based on which harmony part they're playing, instead of the generic harmony instrument icon, making it much easier to tell players apart. These are the same icons that are already displayed at the start of the song, below the vocal highway.

image

In addition, players that are sitting out will now be displayed with a grayed out icon on the Active Players List.

Requires YARG.Core PR: https://github.com/YARC-Official/YARG.Core/pull/287


wyrdough:
Just for future reference, please don't commit core pointers unless it's a last second thing immediately before a merge. They are pretty much guaranteed to be wrong by the time a PR is merged.

PR #1134, opened 2025-09-17
Configurable Rock Meter Presets / Improve Rock Meter for Vocalists

thomeval:
This PR introduces configurable Rock Meter Presets that allows each player to change the behaviour of the Rock Meter. As with other presets, players can each set their own Rock Meter Preset using Profiles, either from the list of default presets provided, or create their own custom one. The following options are available for custom presets:

- Starting Happiness
- Miss Damage Multiplier
- Overhit Damage Multiplier
- Hit Recovery Multiplier
- Star Power Multiplier
- Vocals Miss Damage Multiplier
- Vocals Hit Recovery Multiplier

Five presets are offered available out of the box, in increasing order of difficulty: Casual, Easy, Normal, Hard, and Unfair. Easier presets set the above values to make the Rock Meter drain more slowly and recover quicker. The Hard and Unfair presets also use 50% as the starting happiness value, instead of the default 83.3%, while also reducing the effectiveness of Star Power.

In addition, this PR also dramatically improves the Rock Meter behaviour for vocalists. Happiness will increase much more for each AWESOME phrase, drop much more for an AWFUL one, with results in between having a smaller effect based on the points earned for that phrase.

Finally, some default values have been tweaked as well, due to feedback received from other YARG contributors.


Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -87,7 +87,7 @@ protected override DrumsEngine CreateEngine()
             }
 
             var engine = new YargDrumsEngine(NoteTrack, SyncTrack, EngineParams, Player.Profile.IsBot);
-            EngineContainer = GameManager.EngineManager.Register(engine, NoteTrack.Instrument, Chart);
+            EngineContainer = GameManager.EngineManager.Register(engine, NoteTrack.Instrument, Chart, Player.RockMeterPreset);
```
wyrdough, line 90: If we keep adding parameters, we're going to need to use a struct at some point, but this is fine for now..

wyrdough (approved):

PR #1136, opened 2025-09-17
Add option for note height scaling

poliwhirlge:
Goes with YARG.Core PR https://github.com/YARC-Official/YARG.Core/pull/290

Scales all note height by the NoteHeight parameter in the highway preset. I checked all note types and it looked visually ok to me at both extremes, but we can tweak the min and max values if needed.


wyrdough:
Sorry for letting this languish for so long. All looks good except that changes to note height don't seem to be reflected in the preview track.

If you can fix that and merge dev into this branch and resolve the conflict (take everything from both dev and your branch) I can go ahead and merge this and the core PR.

PR #1140, opened 2025-09-21
Fix issues where stem files were getting locked by os

polson:
**Corresponding YARG.Core PR:** https://github.com/YARC-Official/YARG.Core/pull/293

The issue is we were freeing everything except the **source handle**s of the stems. This broke previously in this pr https://github.com/YARC-Official/YARG/pull/1126

This broke because I mistakenly thought there was a single **source handle** that the mixer was freeing.

## The Original Bug

1. Launch, select quickplay
2. Highlight a song A, then highlight a different song B
3. Try to rename one of the source audio files for Song A

**Expected:** Rename works  
**Actual:** File was locked by OS

## Background Notes

- The **source handle** is created from the C# io `Stream` class, which accesses the file on disk.
- If the source handle is not freed, it will probably keep some portion of the file in memory, and the OS will lock the file. So when scrolling previews, we were leaking memory.
- Previously, a Mixer had an optional source handle, and the channels of the mixer also had a reference to a source handle.  This was causing the double free issue.  With these changes only the channels of the mixer have a source handle.

## Solution

We need to free the source handle when the mixer is disposed.  This is going to be the responsibility of channels added to the mixer.  The mixer will only free its own handle when disposed.

Since source stream -> channel is one-to-many, the channel might have already freed its source stream in `onDispose()`.  It will try to free it again, and if it was already freed it will return a `Handle` error.  We ignore that error, and log any other error.

This does some refactoring to simplify the mixer logic and source handle management:

The main changes are:
* 3 `StemMixer.addChannel` methods -> `addChannel` and `addChannels` methods
  * `addChannel` is a convenience method for `addChannels` 
    * One-to-one `Stream` to stem
  * `addChannels` takes in a stream and list of ChannelInfo structs, which contain stem and optional indices information
    * One-to-many mapping of `Stream` to stems, and prevents creating multiple source streams from the same `Stream`
    * Used for MOGG sources

### DIFF
--- a/Assets/Script/Audio/Bass/BassAudioManager.cs
+++ b/Assets/Script/Audio/Bass/BassAudioManager.cs
@@ -174,26 +174,7 @@ namespace YARG.Audio.BASS
             {
                 return null;
             }
-            return new BassStemMixer(name, this, speed, mixerVolume, handle, 0, clampStemVolume);
-        }
-
-        protected override StemMixer? CreateMixer(string name, Stream stream, float speed, double mixerVolume, bool clampStemVolume)
-        {
-            if (GlobalAudioHandler.LogMixerStatus)
-            {
-                YargLogger.LogDebug("Loading song");
-            }
-
-            if (!CreateMixerHandle(out int handle))
-            {
-                return null;
-            }
-
-            if (!CreateSourceStream(stream, out int sourceStream))
-            {
-                return null;
-            }
-            return new BassStemMixer(name, this, speed, mixerVolume, handle, sourceStream, clampStemVolume);
+            return new BassStemMixer(name, this, speed, mixerVolume, handle, clampStemVolume);
         }
 
         protected override MicDevice? GetInputDevice(string name)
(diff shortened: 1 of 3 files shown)


wyrdough:
So this breaks CONs...badly.

I have not tracked down why it is happening, but `CONFileStream::Read` gets stuck in an infinite loop. It calls `UpdateBuffer`, which exits early and doesn't read more of the audio.

Interestingly, since that's all happening in its own thread, the game continues to work if you then select a song that isn't a CON, it just wastes a core's worth of CPU. (and I guess would eventually get all cores stuck if you kept scrolling through CONs in the library)

polson:
OK I had my file stream and BASS streams confused.  We were creating a source stream multiple times for the same file in the case of CON files which caused bad things to happen.

So to reiterate, we went from 3 tightly coupled `addChannel` methods on our `BassStemMixer` to 1 generic method

Previously the three methods were called depending on which type of song you are loading (CON, ini, etc)

`addChannel` should take in BASS source streams instead of file `Stream`s, because for each stem, it can have the same source stream (in the case of a MOGG file) or a different source stream.  The mixer should not have to care about this, which reduces some complexity.

PR #1145, opened 2025-09-26
Add MIDI Drumkit profile type

Frickitickitavi:
# `MIDI Drumkit` profile type

[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/294)

## Overview
This PR adds a new profile type, called `MIDI Drumkit`. This profile type will eventually be used for Elite Drums, but it's already useful as an all-purpose drums profile that allows 4L, 4L Pro, and 5L charts to all be played in their native formats without changing profiles or performing chart conversions.

The profile contains separate bindings for each 4L Pro and 5L input, except for kick, which is assumed to be consistent across all formats. There are also commented-out placeholder bindings for Elite Drums, which will eventually be listed at the top of the bindings menu since that will be this profile type's native format.

For guidance on implementing the onboarding workflow for this profile type, see the Discord conversation spanning from [here](https://discord.com/channels/1086048856678084609/1098652752873279558/1408832767948619828) to [here](https://discord.com/channels/1086048856678084609/1098652752873279558/1411056345712693379).

## Changes to other profile types
For clarity and consistency, this PR also introduces a few other changes to the names of profile types:
* `Standard Drums` is now `RB Drumkit (4-Lane and Pro)`
* `Five-Lane Drums` is now `GH Drumkit`
* `Keys (Pro and Five-Lane)` is now `Keys (Pro and 5-Lane)`

These changes affect only the localization strings, not enum names or any other internal value. The motivations behind these changes are:
* Replacing "Drums" with "Drumkit" helps clarify the often-confusing distinction between profile types and playable parts; "drumkit" refers to a controller, while "drums" refers to a chart
* `MIDI Drumkit` is now the most preferable drums profile type, so we don't want to call RB stock controllers "standard" anymore
* Using numerals instead of spelled-out numbers is more concise; we should make this change in other contexts as well, but that's outside the scope of this PR
* I omitted "(5-Lane)" from `GH Drumkit` because I don't want players making the classic assumption that 5L must be better than 4L, and because - unlike with RB - there is only one unambiguous drum format associated with GH

## Caveats and shortcomings
The current implementation of setlists in YARG is already flawed in its treatment of setlists that support an instrument on some songs but not all. The `MIDI Drumkit` profile suffers from similar issues related to setlists that have mixed support for certain drum formats.

I've included several examples below, but **tl;dr: I'm considering this acceptable for now, since it's effectively an extension of the existing issues with mixed-support setlists. Feel free to disagree and insist that we find graceful solutions to these issues before merging this PR.**

### Examples
Imagine a setlist consisting of the following in this order:
1) A natively-4LP chart
2) A natively-5L chart

Based on the first song, the player is free to select either 4L or 4LP. However, for the second song, they will end up playing the 4L or 4LP conversion of that chart, rather than the native 5L chart. This isn't really what a `MIDI Drumkit` profile user wants, and it's extra weird because there's otherwise no way for such a player to access a 5L-to-4L conversion.

This scenario also applies in reverse:
1) A natively-5L chart
2) A natively-4LP chart

The player must select 5L Drums, Then, for the second song, they receive the 4LP-to-5L conversion, rather than the native 4LP chart that they likely prefer. Just like before, this is also the only way that a player could access such a conversion.

However, things get even weirder once we account for songs that are explicitly marked as 4L non-Pro. Imagine this setlist:
1) A natively-4L non-Pro chart
2) A natively-5L chart

Based on the first song, the player must select non-Pro 4L Drums. For the second song, they receive the 5L-to-4LNP conversion. So far, this is the same as before, but now imagine the reverse:
1) A natively-5L chart
2) A natively-4L non-Pro chart

The game treats the player as if they have a 5L Drumkit. However, it then looks ahead and determines that, because 4LNP cannot be converted to 5L, the second song has _no playable part_ for the player, and the player is forced to sit out of the setlist entirely. We have the same exact set of songs, but the native format of the first song is determining whether the player has to play chart conversions (not ideal) or sit out completely (even worse).

Along the same lines, any playlist that contains a mix of 4LP and 4LNP charts (with no 5L anywhere) will only allow players to select 4LNP, regardless of what the first song is.


wyrdough:
I'm going to go ahead and merge this, but there is some homework for us both:

1. Difficulty select screen should specify "5L Drums" or "5-Lane Drums" when playing a 5 lane native chart (and maybe "4-Lane Drums" on 4 lane native chart)
2. I need to implement a friendly binding dialog for 5L drums. (@grishhung pls make art for 5L kit when you get a chance)

PR #1146, opened 2025-09-26
Implement normalization

polson:
**Corresponding YARG.Core PR:** https://github.com/YARC-Official/YARG.Core/pull/295

## Overview

This PR implements normalization across songs as an optional setting.

## Problem

Songs tend to have different volumes, which is an annoyance to users.

## Solution

We add a new `BassNormalizer` class that handles asynchronous normalization.

### BassNormalizer Class

The `BassNormalizer` class has public:
- `AddStream` method 
- `Gain` value
- `OnGainAdjusted` action

### Summary

- Any time we add a channel to `BassStemMixer`, we also call `AddStream` on the `BassNormalizer`
- Once stems are added, the normalizer will begin async processing the input streams to calculate a gain factor
- The gain factor is then applied to the playback mixer to normalize the audio

## How It Works

The processing is done in chunks. Each time we add a stream to the normalizer:
* The input stream is first copied to memory
  * We need to do this because otherwise we would have a parallel disk read in the file, which makes both operations slower (at least on HDD)
  * On my machine (a relatively beefy PC) with songs on HDD, the memory copy takes anywhere from 1-100ms.  Most of the time caching the stream happened in 1ms, even on HDD.  I believe this is because of OS level cache warmup.  It would be worth testing to see how this performs on a USB thumb drive or similarly slow device for example.
* The normalizer adds the stream to an internal mixer and starts RMS calculation
  * Each time a new channel is added the RMS calc restarts
* The mixer output is processed in chunks
  * For each chunk, we calculate RMS value by using `Bass.ChannelGetLevel()` which is faster than doing it manually
    * A bit about the math: Since we are processing in chunks, we need to take the RMS value of the chunk, square it, then multiply by the number of samples in the chunk.  This gives us just the sum of squares.   To get the actual RMS value of all audio up to this point we accumulate the total sum of squares and divide by the total samples processed, then square root.  Even with the extra multiplications and sqrt call, using the built in `Bass.ChannelGetLevel()` is still much faster than manually calculating.
    * This means we don't have to wait for the entire song to be processed before we can start normalizing, as processing the entire song can typically take 1000 ms or more.  We can obtain "good enough" RMS values usually within the first 10% of the file being processed.  On my machine this happens in like 50-100 ms.
* We implement a max gain factor to prevent very high gain values when we first start processing the stream.  If a song has a long introduction that is very quiet, then the first chunks will report high gain factors.  So we cap it.
* The target gain factor is intended to normalize songs to about -14 LUFS, based on my testing, which is industry standard normalization levels for streaming services apparently.
* We use relative updates to avoid rapid changes in gain

### Memory usage

Actual memory usage might increase anywhere from 50-200 MB typically.  This is based on the size of the files being loaded into memory.  Some CON files can be quite large.

If normalization is turned off, we will no longer be doing the memory copy. So it is important to note the memory and compute requirements of this feature.

## Demonstration

Testing was done with [MiniMeters](https://minimeters.app/)

### Normalization Off
https://github.com/user-attachments/assets/20bf42b2-8c2e-4f04-b7df-9efc140689e1

### Normalization On
https://github.com/user-attachments/assets/48b85e27-cf21-42a8-abf2-eb77f28d96d6

### Settings Screenshot
image

### DIFF
--- a/Assets/Script/Audio/Bass/BassNormalizer.cs
+++ b/Assets/Script/Audio/Bass/BassNormalizer.cs
@@ -1,0 +1,219 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+using ManagedBass;
+using ManagedBass.Fx;
+using ManagedBass.Mix;
+using YARG.Core.Audio;
+using YARG.Core.Logging;
+
+namespace YARG.Audio.BASS
+{
+    public class BassNormalizer : IDisposable
+    {
+        private const int   BUFFER_SIZE        = 512 * 1024;
+        private const float TARGET_RMS         = 0.175f;
+        private const float MAX_GAIN           = 1.5f;
+        private const int   MAX_THREADS_ATTRIB = 86017;
+        private const int   START_SECONDS      = 5;
+
+        private          int                     _mixer;
+        private readonly List<Stream>            _streams = new();
+        private readonly List<int>               _handles = new();
+        private          CancellationTokenSource _gainCalcCts = new();
+
+        public float               Gain { get; private set; } = 1.0f;
+        public event Action<float> OnGainAdjusted;
+
+        public bool AddStream(Stream stream, params StemMixer.StemInfo[] stemInfos)
+        {
+            if (_mixer == 0)
+            {
+                if (!CreateMixer(out _mixer))
+                {
+                    return false;
+                }
+            }
+
+            if (!CloneStreamToMemory(stream, out var clonedStream))
+            {
+                YargLogger.LogError("Failed to clone stream!");
+                return false;
+            }
+
+            if (!BassAudioManager.CreateSourceStream(clonedStream, out int sourceStream))
+            {
+                YargLogger.LogFormatError("Failed to load stem source stream: {0}!", Bass.LastError);
+                return false;
+            }
+            _handles.Add(sourceStream);
+
+            foreach (var stemInfo in stemInfos)
+            {
+                var volumeMatrix = stemInfo.VolumeMatrix;
+                if (volumeMatrix != null)
+                {
+                    int[] channelMap = stemInfo.Indices.Append(-1).ToArray();
+                    int streamSplit = BassMix.CreateSplitStream(sourceStream, BassFlags.Decode, channelMap);
+                    if (streamSplit == 0)
+                    {
+                        YargLogger.LogFormatError("Failed to create split stream: {0}!", Bass.LastError);
+                        return false;
+                    }
+                    _handles.Add(streamSplit);
+
+                    if (!BassMix.MixerAddChannel(_mixer, streamSplit, BassFlags.MixerChanMatrix))
+                    {
+                        Bass.StreamFree(streamSplit);
+                        YargLogger.LogFormatError("Failed to add channel {0} to mixer: {1}!", stemInfo.Stem,
+                            Bass.LastError);
+                        return false;
+                    }
+
+                    if (!BassMix.ChannelSetMatrix(streamSplit, volumeMatrix))
+                    {
+                        YargLogger.LogFormatError("Failed to set {stem} matrices: {0}!", stemInfo.Stem, Bass.LastError);
+                        return false;
+                    }
+                }
+                else
+                {
+                    if (!BassMix.MixerAddChannel(_mixer, sourceStream, BassFlags.Default))
+                    {
+                        YargLogger.LogFormatError("Failed to add channel {0} to mixer: {1}!", stemInfo.Stem,
+                            Bass.LastError);
+                        return false;
+                    }
+                }
+            }
+
+            StartGainCalculation();
+            return true;
+        }
+
+        private bool CreateMixer(out int mixerHandle)
+        {
+            mixerHandle = BassMix.CreateMixerStream(44100, 2, BassFlags.Decode);
+            if (mixerHandle == 0)
+            {
+                YargLogger.LogFormatError("Failed to create mixer: {0}!", Bass.LastError);
+                return false;
+            }
+
+            if (!Bass.ChannelSetAttribute(mixerHandle, (ChannelAttribute) MAX_THREADS_ATTRIB, GlobalAudioHandler.MAX_THREADS))
+            {
+                YargLogger.LogFormatError("Failed to set mixer processing threads: {0}!", Bass.LastError);
+            }
+
+            _handles.Add(mixerHandle);
+            return true;
+        }
+
+        private bool CloneStreamToMemory(Stream original, out MemoryStream clonedStream)
+        {
+            clonedStream = null;
+            if (!original.CanRead || !original.CanSeek)
+                return false;
+
+            var originalPosition = original.Position;
+            try
+            {
+                original.Position = 0;
+                clonedStream = new MemoryStream();
+                original.CopyTo(clonedStream);
+                clonedStream.Position = originalPosition;
+                _streams.Add(clonedStream);
+                return true;
+            }
+            catch
+            {
+                clonedStream?.Dispose();
+                clonedStream = null;
+                return false;
+            }
+            finally
+            {
+                original.Position = originalPosition;
+            }
+        }
+
+        private void StartGainCalculation()
+        {
+            _gainCalcCts.Cancel();
+            _gainCalcCts.Dispose();
+            _gainCalcCts = new CancellationTokenSource();
+
+            var progress = new Progress<double>(gain =>
+            {
+                OnGainAdjusted?.Invoke((float) gain);
+            });
+
+            Task.Run(() => CalculateRms(progress, _gainCalcCts.Token), _gainCalcCts.Token);
+        }
+
+        private void CalculateRms(IProgress<double> progress, CancellationToken token)
+        {
+            double cumulativeSumSquares = 0.0;
+            long totalSamples = 0;
+            Bass.ChannelSetPosition(_mixer, START_SECONDS);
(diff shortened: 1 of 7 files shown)


wyrdough:
I would prefer if the setting to enable this was in experimental settings for now. (Not sure if we'll ever make it on by default, tbh, given that it can eat so much memory)

Assets/Script/Audio/Bass/BassStemMixer.cs
```diff
@@ -181,17 +213,6 @@ protected override int GetSampleData_Internal(float[] buffer)
             return data;
         }
 
-        protected override int GetLevel_Internal(float[] level)
```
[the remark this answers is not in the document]
wyrdough: See #993 for the intended use I just haven't gotten around to updating since 0.13 yet...

polson:
Updated the normalization video to show latest changes (including memory leak fix)

wyrdough:
Were you ever able to quantify the extra CPU load this will cause?

polson:
I'll do that and share my results. I expect an increase for the first couple seconds of the song playing then it should be the same after that once processing is done

polson:
Here's a couple of measurements on different songs, the highest one I saw was 68%.   I did this iin the `BassNormalizer` class in the code using `StopWatch` and  `Process`.

Screenshot 2025-10-15 at 6 38 01 PM


```
  Wall time: 805.43ms
  CPU time: 2826.63ms
  CPU usage: 35.09%
  Iterations: 19
  Avg per iteration: 42.39ms

  Wall time: 902.45ms
  CPU time: 6143.87ms
  CPU usage: 68.08%
  Iterations: 18
  Avg per iteration: 50.14ms
  ```
  
  We are trying to measure RMS as fast as possible using multicore mixer processing so I would expect some kind of cpu spike when moving around the song select or clicking next in the music player

I also tried `Bass.CPUUsage` but didn't see any cpu spike with that, it sits at 1-5% which is the same as normalization off.

Unity profiler just tells me how long the operation took which I am already measuring.  But maybe I just don't know how to use it.

Let me know if you'd like me to try a different way of measuring this.

**Comment by polson** [comment]
Fixed conflicts and updated it to do relative updates, which prevents audio spikes

**Comment by Ape** [comment]
Should we store the values for the song so that we can just look up the database if we have computed the song gain before?

**Comment by Ape** [comment]
I tested this PR in practice. It's really good! Everything I wished for in #1257 before I knew this already existed. Also, I didn't notice any performance issues.

File: Assets/Script/Audio/Bass/BassNormalizer.cs [hunk 2]
```diff
@@ -0,0 +1,229 @@
...
+            }
+            _handles.Add(sourceStream);
+
+            foreach (var stemInfo in stemInfos)
+            {
+                var volumeMatrix = stemInfo.GetVolumeMatrix();
+                if (volumeMatrix != null)
+                {
+                    int[] channelMap = stemInfo.Indices.Append(-1).ToArray();
+                    int streamSplit = BassMix.CreateSplitStream(sourceStream, BassFlags.Decode, channelMap);
```
Ape, line 84: Would you need `BassFlags.Decode | BassFlags.SplitPosition` so that the streams have their own positions?

Assets/Script/Audio/Bass/BassNormalizer.cs
```diff
@@ -0,0 +1,229 @@
...
+            });
+
+            Task.Run(() => CalculateRms(progress, _gainCalcCts.Token), _gainCalcCts.Token);
+        }
+
+        private void CalculateRms(IProgress<double> progress, CancellationToken token)
+        {
+            double cumulativeSumSquares = 0.0;
+            long totalSamples = 0;
+            Bass.ChannelSetPosition(_mixer, START_SECONDS);
```
Ape: It says "seconds", but I think the parameter is in bytes.

Assets/Script/Audio/Bass/BassNormalizer.cs
```diff
@@ -0,0 +1,229 @@
...
+                    {
+                        YargLogger.LogFormatError("Failed to free stream (THIS WILL LEAK MEMORY!): {0}!",
+                            Bass.LastError);
+                    }
+                }
+            }
+            _mixer = 0;
+            _streams.Clear();
+            _handles.Clear();
+        }
```
Ape, line 245: Looks like the Task started in `StartGainCalculation` is not cancelled in `Dispose`.

polson:
> Should we store the values for the song so that we can just look up the database if we have computed the song gain before?

I would like to do this in a separate PR if we do it at all.  This is an opt-in experimental feature which should happen pretty quickly (< 2s) on modern cpus so there's not a lot of need to cache things right now, imo

Thank you for your review :)

Assets/Script/Audio/Bass/BassNormalizer.cs
polson: I don't think so, the streams shouldn't need their own positions since we just forward scan through the stream

Assets/Script/Audio/Bass/BassNormalizer.cs
polson: Good catch, we might as well just remove this

Assets/Script/Audio/Bass/BassNormalizer.cs
polson: Good catch I think we should clean up the token here

Ape:
Perhaps it would be better to target something like -17 LUFS since some songs are around there and we want to avoid positive gain, or even worse, hitting the max gain limit.

Assets/Script/Audio/Bass/BassNormalizer.cs
```diff
@@ -0,0 +1,230 @@
...
+                    return;
+                }
+
+                int bytesRead = Bass.ChannelGetData(_mixer, buffer, buffer.Length);
+                if (bytesRead <= 0)
+                    break;
+
+                var bufferSeconds = Bass.ChannelBytes2Seconds(_mixer, bytesRead);
+                float[] level = new float[1];
+                bool didGetLevel = Bass.ChannelGetLevel(_mixer, level, (float) bufferSeconds,
```
Ape: Do both `Bass.ChannelGetData` and `Bass.ChannelGetLevel` advance the stream?

Assets/Script/Audio/Bass/BassNormalizer.cs
```diff
@@ -0,0 +1,230 @@
...
+                    if (!BassMix.MixerAddChannel(_mixer, sourceStream, BassFlags.Default))
+                    {
+                        YargLogger.LogFormatError("Failed to add channel {0} to mixer: {1}!", stemInfo.Stem,
+                            Bass.LastError);
+                        return false;
+                    }
+                }
+            }
+
+            StartGainCalculation();
```
Ape, line 117: Might not be okay to start the calculation yet since `AddStream` could be called multiple times (with multiple `AddChannels`). Perhaps it could be started in `BassSteamMixer::Play_Internal`.

polson: We restart the task when a new stream is added, so that's fine.

It is better to do it this way for the case of practice mode, where we don't start playing the stream immediately.  Doing it this way we gain the benefit of a normalized stream by the time we start the practice section.  If we did this on the first call to play, we would hear normalization take effect in practice mode.  

For quickplay / music player, the vast majority of songs start out with silence or a count-in so the normalization is not very detectable.

For the music library previews, songs fade in, which makes it quite difficult to hear normalization taking place

Assets/Script/Audio/Bass/BassNormalizer.cs
polson: Good catch, you're right.  Turns out we could just call ChannelGetLevel.  I also updated this so that updates are more frequent which actually gives the cpu more breathing room from what I can tell, so lower cpu usage

polson:
> Perhaps it would be better to target something like -17 LUFS since some songs are around there and we want to avoid positive gain, or even worse, hitting the max gain limit.

-14 LUFS is what streaming services use for normalization so that's what I'm aiming for

Positive gain is fine imo, what we don't want is rapid positive gain at the beginning of normalization if a song starts with a quiet section.

Assets/Script/Audio/Bass/BassStemMixer.cs
```diff
@@ -421,6 +468,14 @@ protected override void DisposeManagedResources()
             {
                 return;
             }
+            if (_gainDspHandle != 0)
+            {
+                Bass.StreamFree(_gainDspHandle);
```
Ape: I think this should be
```cs
Bass.ChannelRemoveDSP(_mixerHandle, _gainDspHandle);
```

Ape:
> Positive gain is fine imo

Perhaps it is, but are you certain that it won't cause clipping / distortion in any case?

polson:
> Perhaps it is, but are you certain that it won't cause clipping / distortion in any case?

We have a compressor over the playback mixer to prevent clipping first of all, but normalization for most tracks is actually reducing volume from the unnormalized version for the majority of tracks.  I haven't seen any clipping or distortion personally.  I'm not worried about it

Assets/Script/Audio/Bass/BassStemMixer.cs
polson: Good catch, fixed

wyrdough:
I'm going to go ahead and merge this since the code seems fine and it's an opt-in experimental feature, but I do think there's going to be some more tuning required in the long run. 

My main issues are:
* It seems a bit too aggressive at reducing the overall loudness (at least on the setlist songs I tested)
* The adjustment throws off the relative balance of the song vs the crowd sounds (not that it was perfect before, lol)

Ape:
Thanks for implementing this! It's a huge quality-of-life upgrade for me.

PR #1151, opened 2025-09-30
Added setting to choose how to display Discord rich presence

bullseye126:
This PR adds a new setting: **Discord Rich Presence**. It adjusts how YARG shows on your current activity on Discord, also known as rich presence.

The setting has 3 options:

- **Show all details** (default)
  - The normal functionality. It remains unchanged to how it was before this PR.
- **Show without the currently playing song**
  - Still shows the activity on discord, but not the song you are currently playing. Instead it'll show the default activity by calling `SetDefaultActivity()` in `DiscordController`.
- **Hide**
  - Completely hides the activity on discord and also disposes the discord instance in YARG by calling `TryDispose()` in `DiscordController`.

image

Let me know if anything needs changing. This is my first PR.

### DIFF
--- a/Assets/Script/Integration/DiscordController.cs
+++ b/Assets/Script/Integration/DiscordController.cs
@@ -37,6 +37,7 @@ namespace YARG.Integration
         private const string LARGE_ICON_KEY = "icon_stable";
 #endif
 
+        private bool _initialized = false;
         private Discord.Discord _discord;
 
         // Keep track of this in case we want to update the value asynchronously
@@ -53,6 +54,19 @@ namespace YARG.Integration
 
         public void Initialize()
         {
+            _initialized = true;
+
+            CreateInstance();
+        }
+
+        public void CreateInstance()
+        {
+            // Skip if loading screen hasn't finished loading localization (localization is required for rich presence to function normally)
+            if (!_initialized)
+            {
+                return;
+            }
+
             // Don't create instance if Discord rich presence is turned off in settings
             if (SettingsManager.Settings.DiscordRichPresence.Value == DiscordRichPresenceMode.Hide)
             {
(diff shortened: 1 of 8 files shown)


Assets/Script/Integration/DiscordController.cs
```diff
@@ -47,7 +56,21 @@ public void Initialize()
             // Listen to the changing of states
             GameStateFetcher.GameStateChange += OnGameStateChange;
 
-            // Create the Discord instance
+            // Create the Discord instance if Discord rich presence is turned on
+            if (SettingsManager.Settings.DiscordRichPresence.Value != DiscordRichPresenceMode.Hide)
```
wyrdough, line 60: You need to make sure the `GameStateFetcher.GameStateChange += OnGameStateChange` above doesn't get executed if rich presence is completely disabled so that there's no chance that it gets done twice if the user later toggles it back on.

I'd probably just put the following at the top of `Initialize` and remove the check around `CreateInstance` 
```
if (SettingsManager.Settings.DiscordRichPresence.Value == DiscordRichPresenceMode.Hide)
{
    return;
}
```

Assets/Script/Integration/DiscordController.cs
```diff
@@ -232,7 +256,7 @@ private void OnApplicationQuit()
             TryDispose();
         }
 
-        private void TryDispose()
+        public void TryDispose()
```
wyrdough, line 259: Since this can now be called from `SettingsManager`, you can't rely on the cleanup that would normally happen in `OnDestroy`. You should be running `GameStateFetcher.GameStateChange -= OnGameStateChange;` somewhere in `TryDispose()` to deal with the case where the game started up with rich presence enabled and it later gets disabled.

The point of this is to make sure that we never get into a situation where OnGameStateChange is getting called twice for each state change, in case it wasn't clear.

wyrdough (changes requested):
Thanks for the PR, this is a useful feature to have!

I haven't done any functional testing, but I did note one small issue with the handling of subscribing/unsubscribing to/from game state change events. If you can get that fixed up I'll do some testing of my own to make sure everything works as expected and then this can be merged.

Thanks again!

wyrdough (approved):
Looks good to me, thanks for getting on those changes so quickly!

PR #1152, opened 2025-10-01
Add Expand / Collapse functionality to headers in Music Library

polson:
This adds expand and collapse functionality to headers.  

This builds off of previous work by @prgmatic here https://github.com/YARC-Official/YARG/pull/1062

Demonstration:

https://github.com/user-attachments/assets/5008b876-fa68-4d08-a0dd-0407a6ac7f0f

### DIFF
--- a/Assets/Art/Menu/MusicLibrary/MusicLibraryIcons.png
+++ b/Assets/Art/Menu/MusicLibrary/MusicLibraryIcons.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:d689a1103ed1c772050aa7b54f6533b274acba81fa15f365d14c158fe75ffd92
-size 2600
+oid sha256:82675a8033e7f21ff695859ccab3e056564e75b806c5b1ce6ef9d6fd2550570f
+size 4002
--- a/Assets/Art/Menu/MusicLibrary/MusicLibraryIcons.png.meta
+++ b/Assets/Art/Menu/MusicLibrary/MusicLibraryIcons.png.meta
@@ -3,7 +3,7 @@ guid: 6947e69a25ccfa049a22215a785e9f8a
 TextureImporter:
   internalIDToNameTable: []
   externalObjects: {}
-  serializedVersion: 12
+  serializedVersion: 13
   mipmaps:
     mipMapMode: 0
     enableMipMap: 0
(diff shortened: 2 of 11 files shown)


polson:
still an edge case issue when at the very top

wyrdough:
The first thing I notice when testing for functionality is that collapsing a section that does not contain the currently selected song still moves the selection to the header of the section it is in. The selected song should not change unless the collapsed header is the one containing the current selection. In that case, yes, moving to the newly collapsed header is the correct behavior. When collapsing, selectedindex need not change at all if the index of the header is greater than the current index. If it is, then if selectedindex is not within the collapsing section the number of songs in the collapsing should be subtracted from selectedindex to keep us on the same song.

In short, only move the user's selection if you have to.

polson:
Whoops I think I broke that.  Thanks for testing

polson:
Fixed now


https://github.com/user-attachments/assets/88f13c5f-eca1-4b9c-bb05-1cef60bfb5b4

Dansla116:
Pulled and merged with latest build.  Only merge conflicts were in these files due to the added "Expand All" / "Collapse All" strings.  These were straightforward to resolve:
`Assets/StreamingAssets/lang/es-ES.json`
`Assets/StreamingAssets/lang/ja-JP.json`
`Assets/StreamingAssets/lang/pt-BR.json`
`Assets/StreamingAssets/lang/pt-PT.json`

I've verified that expand and collapse works as expected in the music library.  And the "More Options" > "Expand All" or "Collapse All" work as expected.  I also re-validated the above issue.  Scrolling to a song and collapsing it's header with the mouse will move the scroll position to the newly-collapsed header.  Conversely, scrolling to a song and collapsing a different header with the mouse, will **not** move the scroll position.  Both are correct and consistent.

One non-blocking thing I will point out: is that the playlists (and Favorites) also have the expand/collapse chevrons despite them always only being one section and **not** being collapse-able.  This actually already exists in the latest build, so it's not introduced by this PR.  But the "Expand All" / "Collapse All" options are still in the _More Options_ menu and don't do anything.  Again, non-blocking, might be worth cleaning that menu up for playlists at some point.
image

polson:
@Dansla116 thanks for your review!

I'll fix the issue you mentioned.  This is a low priority feature so there's plenty of time to get it right

PR #1156, opened 2025-10-03
Fix instrument icons on Rock Meter overlapping

thomeval:
This PR fixes instrument icons on the Rock Meter incorrectly being positioned on top of each other instead of spacing out horizontally, when multiple players have approximately the same amount of happiness. Note that the implementation can still result in odd placements when there are a large number of players, but this PR is nonetheless an improvement.

image


Assets/Script/Gameplay/HUD/FailMeter.cs
```diff
@@ -94,14 +95,13 @@ public void Initialize(EngineManager engineManager, GameManager gameManager)
                 SetAutoKill(false);
 
 
-
-
             // attach the slider instances to the scene and apply the correct icon
             for (int i = 0; i < _players.Count; i++)
```
wyrdough, line 99: Not really something for you to do, just writing it down so it makes it more likely to stick in my brain, but it would be better if this were evaluated in reverse so that the order of the icons was the same as the profile ordering...

wyrdough (approved):
Thanks for fixing this..it's better than it was even before it somehow got broken...

PR #1160, opened 2025-10-04
Added "Disconnect" option to Difficulty Select Menu

thomeval:
This PR adds a "Disconnect" option to the Difficulty Select Menu, below the "Sit Out" option. This is useful for quickly removing profiles that are no longer playing, without having to go all the way back to the Profiles Menu to do this.


polson:
I was gonna say "why not both" but then I remembered we have this screen, and disconnecting profiles when device disconnects might mean we never see this

image

wyrdough (changes requested):
Functionally, I'm totally happy with this. I think the UI needs some work, though. It seems like we either need to position the sit out and disconnect elements at the bottom of the container to give some more visual space and/or make their font size smaller than the rest of the options. I'm going to post a screenshot on #artists-chat and see if anyone has any suggestions...

And just as a note to self, it would be nice if the difficulty select menu could handle profiles being connected during song selection so we can better take advantage of the quick reconnect feature.

thomeval:
This has been updated based on feedback from the team. The Difficulty Select menu now looks like this:
image

wyrdough (approved):

PR #1163, opened 2025-10-05
Lower track on song fail

polson:
This adds an animation to lower the track and a delay when a song is failed

It also fades out the mixer, and disables input after fail.

Previously the failed screen would pop up instantly which was jarring.

Demonstration:

https://github.com/user-attachments/assets/82720f66-b268-4d4b-a008-88a2835c6fd2

### DIFF
--- a/Assets/Script/Gameplay/GameManager.cs
+++ b/Assets/Script/Gameplay/GameManager.cs
@@ -1,6 +1,7 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using Cysharp.Threading.Tasks;
 using UnityEngine;
 using UnityEngine.InputSystem;
 using YARG.Core.Audio;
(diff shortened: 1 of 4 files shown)


wyrdough:
I like this. One small problem, though. If there is a drums player and the fail just so happens to occur when the kick animation is playing, the drum track will reappear when that animation completes. (I suspect the same thing would happen if the SP activation bump was happening for a five fret/keys player, too)

The obvious way to fix that would be to have `CameraPositioner` subscribe to `OnSongFailed` in its `Initialize` method or to have GameManager subscribe it. Then when `CameraPositioner` receives the signal it can check if `_gameManager.PlayerHasFailed` is true and kill all its tweens.

polson:
Added handling for conflicting animations

polson:
Actually I still don't love this solution since the CameraPositioner knows a little too much about what the highway should be doing.  Like if the caller tells the camera positioner you want to lower, and then scoop, it shouldn't refuse

polson:
Ok should be good now

wyrdough (approved):
looking good

PR #1164, opened 2025-10-08
Add global calibration to quick settings

polson:
This PR adds global audio and visual calibration in the quick settings

Demonstration:

https://github.com/user-attachments/assets/62eef963-c146-41b9-a499-0fe1b6ddbe4d

### DIFF
--- a/Assets/Prefabs/Gameplay/HUD/Pause/QuickSettings.prefab
+++ b/Assets/Prefabs/Gameplay/HUD/Pause/QuickSettings.prefab
@@ -8,148 +8,119 @@ PrefabInstance:
     serializedVersion: 3
     m_TransformParent: {fileID: 4220475145157243466}
     m_Modifications:
-    - target: {fileID: 375337769934742335, guid: 81863a63e0d96cf459c263ee0ea6f98e,
-        type: 3}
+    - target: {fileID: 375337769934742335, guid: 81863a63e0d96cf459c263ee0ea6f98e, type: 3}
       propertyPath: m_SizeDelta.y
       value: 0
       objectReference: {fileID: 0}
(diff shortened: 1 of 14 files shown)


wyrdough:
Hm, I wonder if there is a way to make it present as a slider setting in quick settings so it can be adjusted with a controller...

polson:
We can certainly do a slider.  Both options are supposed to allow controller but maybe it's not working right.

I think slider probably makes sense here

polson:
On second thought, slider might not be the best choice.

I think calibration values could be anywhere from 0-500ms in the real world

Most folks will have calibration of around 0-20ms and on a slider with that scale its very hard to navigate with a mouse, especially when differences of 5ms might be the adjustment one needs

So given that the input box should be already editable with controller, I don't see how a slider would really improve the ux

That said, there are some things I can change to make this look a little nicer

polson:
Here's a video of updated UI and changing values with a guitar

https://github.com/user-attachments/assets/2d03bfd1-5df6-4da8-9aa4-bd9b0a505ed2

ShawnHardwick:
I know you didn't create the debug menu, but it currently says `0.1ms` after increasing the audio delay by 100ms. The value in the debug menu needs increments by `1000`.

wyrdough (approved):
Sorry for completely forgetting about this. Looks like there has been a change to the quick settings prefab since you did this. If you can fix that conflict, this looks fine to me. (No rush, ofc, but if you could get it done by the end of the weekend that would be much appreciated)

polson:
I'll get this one in too.  This submenu will let me finish up the auto calibration feature too :)

also lol at squashed highway in the video, I forgot about that

polson:
good to go

Issue #297, opened 2023-05-10
solo marker not resetting when close to another one


Doune2Sable:
hey, when two solo marker are close it doesn't reset the note count, which will break the percentage

https://github.com/YARC-Official/YARG/assets/133128256/901edb11-e685-42da-b3b4-b4cf5c6a1570

RileyTheFox:
This is known and I doubt a proper fix will be made until the engine rewrite

ClampyStew:
The solo counter will only end if after the solo marker there is no other marker, also the solo counter will only count the first solo marker (correct me if I'm wrong).

Doune2Sable:
if by that you mean it only count the score then no, it will count the other one (i don't know if it count both) and it will give you prefect solo anyway

EliteAsian123:
This *should* be fixed. Please reopen if not

PR #1165, opened 2025-10-09
Use venue data from milo when there is no venue data in midi

wyrdough:
Updates GameManager to attempt to load milo data before doing the venue autogen stuff.

See [YARG.Core #297](https://github.com/YARC-Official/YARG.Core/pull/297)


polson (approved):
🔥

PR #1171, opened 2025-10-14
Added Spotlights and Singalongs to the UDP datastream

TheFatBastid:
While both are in the Performer venue track, I split them into their own bytes in the datastream.


wyrdough:
Note to self: Actually parse the performer and spotlight events from milo. (Pretty sure I left those out because they aren't used by any existing venues and the bit math made it hard compared to the other venue events, lol)

PR #1172, opened 2025-10-15
Angle highways in multiplayer to prevent fail meter icons from getting too close to the highways

grishhung:
## Changes

Angles the highways like in Rock Band to make the spacing between the edges of the screen and each highway roughly equal; also prevents player icons on the crowd meter from getting too close to the leftmost highway (very rare, but can happen for four players if the players are doing particularly poorly).

This is done by first shifting the camera, then shifting the viewport.

## Issues

- The HUD does not currently reposition to align with the back of the highway; must follow up on this.
- A setting to disable this needs to eventually be added as GH players may want to disable this feature.

### DIFF
--- a/Assets/Art/Animations/Gameplay/Rectangular/FretCymbalHit.anim
+++ b/Assets/Art/Animations/Gameplay/Rectangular/FretCymbalHit.anim
@@ -95,7 +95,7 @@ AnimationClip:
     m_Level: 0
     m_CycleOffset: 0
     m_HasAdditiveReferencePose: 0
-    m_LoopTime: 1
+    m_LoopTime: 0
     m_LoopBlend: 0
     m_LoopBlendOrientation: 0
     m_LoopBlendPositionY: 0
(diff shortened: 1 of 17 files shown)


theli-ua:
> A setting to disable this needs to eventually be added as GH players may want to disable this feature.

I'd say this is a blocker, please don't break existing behaviour all the time without ability to opt-out.
It may also be not a switch but a slider which would control the amount of angling perhaps

grishhung:
Added a setting to control the amount of highway tilt. 0.0 is no tilt at all, and 1.0 is fully tilted such that the leftmost highway has its right edge completely vertical. Set to a default of 0.5.

Also fixed a typo in settings which made the pattern for note streaks ambiguous.

wyrdough (changes requested):
Code seems totally fine to me. Functional testing reveals a couple of issues, though.

First, the center hud elements (player name and countdown) aren't being positioned quite correctly. This is especially noticeable on the highways closest to the edges with 4 or more highways. I believe the issue is that you are using the same X offset for both top and center, but center should really be calculated as if it were closer to the camera since the track is now angled.

~~Secondly, something in this seems to break kick bounce. Oddly, the other track movements seem to work fine, though I haven't tested the miss animation.~~ (me issue, not PR issue)

Lastly, and this may just be a me thing and is more a point of discussion than a "this needs to be fixed," the gems look flat compared to the highway being untilted. (Could be my brain seeing the lighting wrong or sth, I'm not sure)

PR #1173, opened 2025-10-15
Gracefully handle short stems when seeking

Frickitickitavi:
[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/298)

Currently, if a song contains a stem that is shorter than the full length of the song, then seeking to a position later than the end of the stem (such as when loading a preview) plays the short stem from the beginning and logs an error.

This changes the behavior to mute the stem instead. If we somehow fail to seek for some other reason, we log an error, but still mute the stem. When seeking succeeds, we unmute the stem in case it was previously muted

### DIFF
--- a/Assets/Script/Audio/Bass/BassStemMixer.cs
+++ b/Assets/Script/Audio/Bass/BassStemMixer.cs
@@ -118,11 +118,19 @@ namespace YARG.Audio.BASS
             {
                 if (channel.SetPosition(position))
                 {
+                    // Restore the volume, in case we previously failed to seek and muted the stem
                     double volume = GlobalAudioHandler.GetTrueVolume(channel.Stem);
                     channel.SetVolume(volume);
                 } else
                 {
-                    // Failed to seek, so mute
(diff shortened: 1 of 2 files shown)


Assets/Script/Audio/Bass/BassStemMixer.cs
```diff
@@ -116,7 +116,23 @@ protected override void SetPosition_Internal(double position)
 
             foreach (var channel in _channels)
             {
-                channel.SetPosition(position);
+                if (channel.SetPosition(position))
+                {
+                    // Restore the volume, in case we previously failed to seek and muted the stem
+                    double volume = GlobalAudioHandler.GetTrueVolume(channel.Stem);
```
polson: I am wondering if it would be better to just pause the channel of mute so we don't have to rely on GlobalAudioHandler at all

polson:
After some discussion we found a simpler solution that seems to work

wyrdough:
thank you for sparking the conversation that led to a solution

PR #1176, opened 2025-10-19
Use event system for player audio

polson:
YARG.Core PR: https://github.com/YARC-Official/YARG.Core/pull/322

# Audio Event System Refactor

This is a first step in reducing global state for handling per-player audio.

## Overview

This PR decouples audio from the Player using an event system. We can then consolidate audio handling in a class `PlayerAudioManager` with a lifecycle that is tied to the lifecycle of the song / players.

`PlayerAudioManager` then represents the audio manager for all players, and each player is associated internally with an `AudioHandler`.

Right now each `AudioHandler` just calls the existing global stem methods, which remain unchanged. This is so that we can do this refactor in parts without a huge PR.

## Goals

Eventually this will enable us to move the logic out of `GameManager.Audio` into this `PlayerAudioManager`, and will help us avoid having global stem states with a lifecycle tied to the application lifecycle.

## Rationale for Event System Over Composition

1. We have `Player` classes that have very deep inheritance hierarchies. For example `FiveFretGuitarPlayer -> TrackPlayer<> -> TrackPlayer -> BasePlayer`. Each level handles parts of the audio by calling the `GameManager`. Subclasses override some methods to change how the audio is handled over the default in the base class.

2. The `Player` class is already a bit of a "God class", handling Scoring Logic, Visuals, Notifications, Audio, Replays, at each level of the hierarchy.

An Event system flattens out the various player events into a single `Action` that is subscriptable, which makes it easy to understand how the audio is handled

### DIFF
--- a/Assets/Script/Audio/PlayerAudioManager.cs
+++ b/Assets/Script/Audio/PlayerAudioManager.cs
@@ -83,6 +83,9 @@ namespace YARG.Audio
                     case SustainEnded:
                         OnSustainEnded();
                         break;
+                    case StarPowerPhraseHit:
+                        OnStarPowerPhraseHit();
+                        break;
                     case WhammyDuringSustain(var whammyFactor):
                         OnWhammyDuringSustain(whammyFactor);
                         break;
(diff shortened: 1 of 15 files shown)


wyrdough:
Given that 0.14 should be releasing inside of a month, I think I'm going to hold off on this until that happens since it's a pretty major change.

polson:
@wyrdough yeah no need to merge this right now

polson:
Merging this into the epic branch

PR #1178, opened 2025-10-20
Quick binding for Drums and Pro Keys

wyrdough:
Adds a quick binding button to the profile view to allow users to bind keys and drum inputs with a visual guide instead of having to click 87,000 tiny buttons.

Todo: 
* Keys needs to also do 5LK and SP/touch effects somehow.
* Both need to do menu bindings.
* Cleanup

I'm considering going ahead and merging this despite its incompleteness because even in its current state it solves a serious pain point for using MIDI keyboards and drum kits.

Keys:
friendly keys

Drums:
friendly drums


Fieoner:
I think a system like this for binding inputs is also necessary for guitars

Assets/Script/Menu/Common/Dialogs/Onboarding/FriendlyBindingDialog.cs
```diff
@@ -0,0 +1,374 @@
...
+            // TODO: if failed, we should show an error message of some sort
+            button.interactable = true;
+            button.image.color = MenuData.Colors.ConfirmButton;
+            Message.text = _bindingMessages[gameMode].complete;
+            _startButton.Text.text = Localize.Key("Menu.Common.Close");
+            button = _cancelButton.gameObject.GetComponentInChildren<Button>();
+            button.interactable = false;
+            button.image.color = Color.gray;
+
+            // TODO: after we've run out of bindings, we need to change this to a button that says test and switch to
```
polson, line 121: Great idea, though we might want to also be thinking about a way to just jump to the input tester right away

polson:
My unsolicited thoughts on architecture:

You could have an interface that represents an instrument to bind, and pass that into the `FriendlyBindingDialog`

Each instrument you bind is going to have a different:

`Bindings to capture`
`Highlight images for bindings`
`Initial message`
`Complete message`

Not sure what else, but you could throw all that into an interface, then define classes that implement it for each instrument, pass the instance into `FriendlyBindingDialog` which will delegate to it for those parts.

Just my two cents, you do you

Nice work btw, this is super cool

wyrdough:
> My unsolicited thoughts on architecture:
> 
> You could have an interface that represents an instrument to bind, and pass that into the `FriendlyBindingDialog`

If you remember, remind me to actually do this when I implement the rest of the onboarding stuff. (figure it's less likely to get forgotten if I write this message and two people have it somewhere in their brain lol)

PR #1179, opened 2025-10-21
Show active instrument filters on Music Select Screen

thomeval:
This PR enhances the Music Select Screen to show which instrument part filters are active. These are now shown with a blue colour in the instrument difficulty rings display. Multiple filters can be active at once, in which case each affected instrument will be highlighted.

image

Known issue: It is currently not possible to _remove_ a single instrument filter easily. The only way to do this is to clear _all_ filters by pressing the Back button.


wyrdough (approved):
Sorry for leaving this on the back burner forever. Happily, it still merges cleanly, so I'll merge in a few minutes.

PR #1183, opened 2025-10-21
Make uses of lane and fret consistent in labels

grishhung:
## Changes

Made various changes to English language strings such that label names are consistent in settings.

- Changed `Guitar` and `Bass` to `Lead Guitar` and `Bass Guitar` to match `Rhythm Guitar`.
- Made it so that both the 4-lane drums and keys instruments types show that they support mapping for both normal and pro modes.
- Changed all instances of adjectives with numbers in them to now follow the format `5-Lane` and to not use any of `5 Lane`, `Five Lane`, or `Five-Lane`. We were previously using all of these; only the first and last are syntactically correct in English, and the former is the cleaner looking of the two. The word `Lane` is also always capitalized in these settings.
- Made it so that instrument names do not specify the number of lanes for drums in the part selection menu before a chart starts. Once either 4-lane drums or 5-lane drums has been selected as an instrument type, the player only needs to concern themselves with whether they are playing `Drums` or `Pro Drums`. (This is currently the case for 5-fret and 6-fret guitar, which do not specify the number of frets across different instruments.)

## Exceptions

In the pause menu's volume settings, lead guitar is still labeled as `Guitar`, rhythm guitar is still labeled as `Rhythm`, and bass guitar is still labeled as `Bass`. Extending these names makes the UI extremely cluttered due to the large font size.

## Follow-up

Make orderings of options across settings consistent. Currently, keys often comes after vocals rather than before it.


wyrdough:
So I'm guessing this will end up changing "Guitar" in the difficulty select screen, too? Not sure how I feel about that. I'm inclined to think that simplicity is better at least in that case.

PR #1186, opened 2025-10-23
Fix whammy drift in music library and player

polson:
# Sync Drift Correction Fix

Two issues being fixed here:

1. Sync drift correction was only happening in the `SongRunner` which is used only for quickplay and practice mode
2. Sync drift correction was not working right because `GetWhammyPitch` actually returns 1.0 when unpitched, whereas `SetWhammyPitch` converts a percentage, so expects 0 percent when unpitched. Our if/else was wrong for this so sync drift was not being corrected even in quickplay

This PR adds a new `Timer` class that uses `UniTask` so we can easily do something every x seconds. We then use that in the `BassStemMixer` to do sync correction while the song is playing. This fixes sync drift in quickplay and song select.

## Demonstration

**Broken:**

https://github.com/user-attachments/assets/76dc286c-35d3-4d22-b9fd-5340df9e25dc

**Fixed:**

https://github.com/user-attachments/assets/def772f3-1eaf-4915-b866-f6786cdd4e8d

Here's the test song

[Sync Drift Test.zip](https://github.com/user-attachments/files/23109393/Sync.Drift.Test.zip)

### DIFF
--- a/Assets/Script/Audio/Bass/BassStemMixer.cs
+++ b/Assets/Script/Audio/Bass/BassStemMixer.cs
@@ -8,14 +8,16 @@ using YARG.Core.Audio;
 using YARG.Core.Logging;
 using YARG.Core.Song;
 using YARG.Helpers;
+using YARG.Settings;
 
 namespace YARG.Audio.BASS
 {
     public sealed class BassStemMixer : StemMixer
     {
         private const    float WHAMMY_SYNC_INTERVAL_SECONDS = 1f;
-        private readonly int   _mixerHandle;
+        private          bool  IsWhammyEnabled => SettingsManager.Settings.UseWhammyFx.Value;
 
+        private readonly int   _mixerHandle;
         private StemChannel  _mainChannel;
         private StreamHandle _mainHandle;
         private int          _songEndHandle;
@@ -55,7 +57,6 @@ namespace YARG.Audio.BASS
             _speed = speed;
             _whammySyncTimer = new Timer();
             SetVolume_Internal(volume);
-            _BufferSetter(Settings.SettingsManager.Settings.EnablePlaybackBuffer.Value, Bass.PlaybackBufferLength);
         }
 
         protected override int Play_Internal(bool restartBuffer)
@@ -65,15 +66,10 @@ namespace YARG.Audio.BASS
                 return (int) Bass.LastError;
             }
 
-            if (Settings.SettingsManager.Settings.EnablePlaybackBuffer.Value)
+            if (IsWhammyEnabled)
             {
-                if (!Bass.ChannelUpdate(_mixerHandle, Bass.PlaybackBufferLength))
-                {
-                    YargLogger.LogFormatError("Failed to fill playback buffer: {0}!", Bass.LastError);
-                }
+                _whammySyncTimer.Start(WHAMMY_SYNC_INTERVAL_SECONDS, SyncWhammyPitch);
             }
-
-            _whammySyncTimer.Start(WHAMMY_SYNC_INTERVAL_SECONDS, SyncWhammyPitch);
 
             return 0;
         }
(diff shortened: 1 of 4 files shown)


Assets/Script/Audio/Bass/BassStemMixer.cs
```diff
@@ -83,6 +104,7 @@ protected override void FadeOut_Internal(double duration)
 
         protected override int Pause_Internal()
         {
+            _whammySyncTimer.Stop();
```
wyrdough: Since you're calling `Timer()` without passing the `ignoreTimeScale` bool, I believe pause should be handled automagically since the timer won't fire during the pause since unity time scale is set to zero in that case.

Creating and destroying the `CancellationTokenSource` will generate garbage, so it would be better to avoid killing and recreating the timer during gameplay if it can be done another way.

Assets/Script/Helpers/Timer.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class Timer
+    {
+        private CancellationTokenSource _cancellationTokenSource;
+
+        public void Start(float intervalSeconds, Action onTick)
+        {
+            Stop();
+            _cancellationTokenSource = new CancellationTokenSource();
+
+            UniTaskAsyncEnumerable.Timer(
```
wyrdough, line 21: If you do end up removing the `Stop` call in `Pause_Internal` you'll want to save a reference to this task and check if it is running before trying to run it again. I believe its `State` property should remain `Pending` while it runs since it can't complete without cancellation, but you'll want to double check that.

wyrdough (changes requested):

polson:
Something is messed up with the audio buffer too, hold on

Assets/Script/Audio/Bass/BassStemMixer.cs
polson: Just tested this and you're right, it does stop.  TIL.

I guess we can just remove the stop call for now.

polson:
Ok made the timer more efficient and fixed issues with the buffer

We were setting the buffer on playback and also in the constructor of `BassStemMixer`

This was causing pitch shifted stems to desync, most likely becaues the pitch fx has its own dsp buffer. It also causes the song to skip once the new buffer is filled.

You can repro the original issue by setting a huge buffer (5000ms) and then scrolling the music library, after a few (5?) seconds the song skips.

I have no idea why we were setting the buffer like that, since the global/mixer buffers get updated already when the setting changes, and new mixers also have the global buffer setting on creation.

In my testing I didn't see it break anything but also I'm bad at testing  🤷

Assets/Script/Helpers/Timer.cs
polson: Handling that internally in the `Timer` now

Assets/Script/Audio/Bass/BassStemMixer.cs
```diff
@@ -285,6 +302,7 @@ protected override void ToggleBuffer_Internal(bool enable)
         protected override void SetBufferLength_Internal(int length)
         {
             _BufferSetter(Settings.SettingsManager.Settings.EnablePlaybackBuffer.Value, length);
+            _BufferSetter(SettingsManager.Settings.EnablePlaybackBuffer.Value, length);
```
wyrdough, line 307: Don't think we need to set this twice?

wyrdough (changes requested):
Played a few songs on guitar and drums and nothing seemed off to me, but I am not terribly sensitive, so...

Assets/Script/Audio/Bass/BassStemMixer.cs
polson: wtf

wyrdough (approved):

PR #1188, opened 2025-10-24
Add fade effects and various fixes for vocals graphics

grishhung:
## Changes

- Made new shadergraphs for the track, vocals notes, and talkies to allow them to fade out at the edges of the screen. Works with all resolutions and is calculated as a smoothstep from 0.0 to 0.1, then 0.9 to 1.0.
- Created a new shader based on `Assets/TextMesh Pro/Shaders/TMP_SDF Overlay.shader` which allows text to fade out based on screen position. This was required as shadergraphs cannot be applied to the materials used for TMP.
- Made it so that if there are only two vocals parts in a song, only two lyric lines will be shown instead of three. Tested with normal lyrics mode and static lyrics mode, solo vocals, and all three harmony counts.
- Fixed the z-fighting with harmony talkies. Previously, talkies would all be spawned at the same coordinate, and the color that would prevail would be nondeterministic. This offsets them each by 0.25, with HARM1 in front (and all pipes on top of the talkies).

## Gallery

image

## Follow-up

There appears to be some jank going on with vocals positioning at differing resolutions. Upon fixing this, we should ensure:

1. The countdown is centered about the highway.
2. In single player, the multiplier meter is always in line with the 
3. The correct amount of space proportional to the screen resolution is applied above the highway (currently, the stats at the top are blocked in 4K).

Additionally, monitor the following bugs and things to adjust for the vocals highway:

- $ does not work consistently (see the end of "I'm a Bug").
- There is too much space between some words to the point where words can get delayed when they don't have to.
- Overlapping talkies generally look weird and gray; maybe change the math for how they overlap (art issue).
- The SP bar does not have a midway increment (art issue).


Assets/Script/Gameplay/Player/Vocals/VocalTrack.Spawning.cs
```diff
@@ -35,7 +35,7 @@ public partial class VocalTrack
 
         private bool[] _noMoreStaticPhrases = { false, false, false };
 
-        private void UpdateSpawning()
+        private void UpdateSpawning(int totalHarms)
```
wyrdough: Since this is just a partial of `VocalTrack` you don't actually need to pass totalHarms around since `_totalHarms` is already a property of the class (unless you need it to be copied to the stack for some reason)

Assets/Script/Gameplay/Visuals/VocalElements/VocalTalkieElement.cs
```diff
@@ -41,13 +41,23 @@ protected override void InitializeElement()
...
+            mesh.material.SetColor("_BaseColor", color.WithAlpha(ALPHA_VALUE));
 
             // Update the size of the talkie
             var transform = mesh.transform;
             float length = VocalTrack.GetPosForTime(NoteRef.TotalTimeEnd) - VocalTrack.GetPosForTime(NoteRef.Time);
             transform.localScale = transform.localScale.WithX(length);
             transform.localPosition = transform.localPosition.WithX(length / 2f);
+
+            // Prevent Z-fighting (Y acts as the "Z" coordinate here)
+            // All values must be under 0.1 (minimum forward offset for vocal pipes)
```
wyrdough, line 55: You say in the comment that all values must be under 0.1, but then you're setting 0.25 and 0.5 below. Error in comment or in code?

Assets/Script/Gameplay/Visuals/VocalElements/VocalTalkieElement.cs
```diff
@@ -41,13 +41,23 @@ protected override void InitializeElement()
...
+            transform.position = transform.position.WithY(NoteRef.HarmonyPart switch
```
wyrdough: I think it would be clearer if you assigned the result of the switch to a variable and used that in the `WithY` call instead of combining it into one.

Assets/TextMesh Pro/Shaders/TMP_SDF Overlay Fade.shader
```diff
@@ -0,0 +1,330 @@
...
+			clip(faceColor.a - 0.001);
+		#endif
+
+		    // Amending TMP_SDF Overlay.shader to do the same thing we're doing for all other vocals elements
+
+            // Compute normalized screen-space UV from SV_POSITION (NDC -> [0,1])
+            float2 screenUV = input.position.xy / input.position.w * 0.5 + 0.5;
+
+            // Fade in across 0.0 -> 0.1 (left edge), fade out across 0.9 -> 1.0 (right edge)
+            float fadeIn = smoothstep(0.0, 0.1, screenUV.x);
```
wyrdough, line 314: You might consider making the left/right fade size a shader property here and in your shader graph shaders so it can be more easily changed in the future if we end up needing to do that.

Assets/Script/Gameplay/Visuals/VocalElements/VocalTalkieElement.cs
```diff
@@ -41,13 +41,23 @@ protected override void InitializeElement()
 
             // Set the color
             var color = Player.VocalTrack.Colors[NoteRef.HarmonyPart];
-            mesh.material.color = color.WithAlpha(ALPHA_VALUE);
+            mesh.material.SetColor("_BaseColor", color.WithAlpha(ALPHA_VALUE));
```
wyrdough: In this case it doesn't matter since it is only ever called once for the lifetime of the instance, but in general it isn't good practice to reference shader properties by name like this because the lookup is slow. You should cache the numeric ID of the property once and use that instead:

```
private static readonly int _baseColorId = Shader.PropertyToID("_BaseColor");
...
mesh.material.SetColor(_baseColorId, color.WithAlpha(ALPHA_VALUE));
```

Actually, on second thought I wonder if the `_baseColorId` field were static if it would only ever get computed once for all instances of the object... I don't actually know how that works in C#.

Edited to add: Actual research seems to indicate that yes, the static variable would only get initialized once, so please change it so we aren't making unity recompute the reference every time we spawn a talkie.

wyrdough (changes requested):
Looks good overall, just a few changes and I think this will be good to go!

Assets/Script/Gameplay/Visuals/VocalElements/VocalTalkieElement.cs
grishhung: Whoops, code error. All should be under 0.1 indeed, so those should be something like 0.025 and 0.05. Thanks for catching that!

Assets/Script/Gameplay/Player/Vocals/VocalTrack.Spawning.cs
grishhung: Oopsie. I meant to remove this but I never got back around to it. It definitely doesn't need to be copied around like that.

Assets/Script/Gameplay/Visuals/VocalElements/VocalTalkieElement.cs
grishhung: Thanks for catching this. Changed to only compute once as shown here.

PR #1190, opened 2025-10-26
Fix lane spacing on non 16:9 screens

polson:
This PR adds handling for any screen size in QuickPlay, including dynamic resizing.

It handles both highway lane scaling and vocals scaling.

Highway lane scaling is handled by determining the base width of the lane at the very bottom of the screen, then scaling each lane accordingly.  We also consider the screen space height of the highway, and set a limit on how tall it can be in terms of percentage of screen height.

This means you can use virtually any camera preset and the lanes will not overlap each other.

We also simplify HUD positioning by providing a method in the `HighwayCameraRenderer` to get any position on the track based on a normalized coordinate system, which considers the x offset implicitly.

Additionally, the vocals track has its own scaling rules, which cap the track at an aspect ratio of 8.0 to 13.0.  8.0 is the original aspect ratio on a 16:9 screen.

Other changes:
1. The vocals hud is always positioned with first player below the vocals "strikeline" for any number of players

Demonstration:

https://github.com/user-attachments/assets/79c2c27c-2757-4bcf-ad6e-eab64102b763

Test plan for this feature:
https://docs.google.com/spreadsheets/d/14z29f7zoM768wOocg4_s-jsni6DjTHF-M2v6ijGnbwg/edit?usp=sharing


wyrdough (changes requested):
Code looks good to me, but it breaks the positioning of the center hud elements (player name and wait countdown) in non-16:9 ratios.

polson:
Calling this done before the PR gets too huge

polson:
Rebased with dev

polson:
Fixed conflicts again

polson:
something broke with vocals only in rebase

polson:
fixed now

PR #1192, opened 2025-10-29
Use mixer add channel delay to handle pitch fx syncing

polson:
Corresponding YARG.Core PR: https://github.com/YARC-Official/YARG.Core/pull/303

This fixes 3 issues:

1. **Stereo stems were not syncing correctly**
   
   This was fixed by setting the delay in terms of samples, not bytes, and letting Bass figure out how many bytes of delay that should be

2. **CON files were not syncing correctly**
   
   For some CON files, split streams cannot have separate positions. Calls to `SetPosition` were setting all stems to the same position, even if you called it on only one stem. I tried not using `SplitStreamReset` and not using `MixerReset` flag, it made no difference :-/
   
   The fix is to use the `Start` parameter in the `BassMix.MixerAddChannel` call to handle the delay.  Then, when setting position, we remove and re-add the channels to the mixer with the delay.
   
   So what we are doing now is kind of the reverse, where non-pitch shifted stems are given a delay to match the delay of pitch shifted stems.
   
   With this method positions of the stems are accurately reflected, which simplifies position handling.
   
   Seeking retains the delay as well.
   
   The trade off is that with whammy on, starting the mixer from the beginning will have a 2048 sample delay (43 ms or something) before we hear anything. Pause / Unpause does not have any delay.
   
   IMO This is unnoticeable in the music player or Quick Play. The `SongRunner` still handles syncing up in Quick Play quite easily so there's no issues there.
   
   Calibration should not have any issues since we turn off whammy beforehand
   
   

3. **Stems were going out of sync regardless of whammy setting**
    
    This was caused because we handled speed / chipmunking by creating a `TempoStream` for every channel in the mixer.  When we called `SetSpeed` on the mixer we would then set the speed of every channel.  However this does not ensure that all stems apply the speed at exactly the same time, leading to desync.  You can test this with one of the provided sync tracks, on nightly.  Just play quickplay and pause and unpause a bunch.  This causes the SongRunner to adjust speed to resync the song, which actually desyncs stems from each other.
    
    The fix is to use a single TempoStream created from the mixer itself.  In order to do this correctly we have to manually track position of the TempoStream to ensure it is reported by the mixer `GetPosition` correctly.  This should also be computationally cheaper than applying a TempoStream to every channel.
    
    We removed the `restartBuffer` param from the mixer `Play` method.  The caller should not have to care about restarting the buffer.  We can infer whether or not to do that by:
    1. If the mixer position changed with `SetPosition`, we want to restart the buffer so we dont hear stale audio
    3. Otherwise just use false, ie for the case of pause / unpause.

## Things to test:

The test plan contains the most important tests for this feature

[Test Plan (Google sheets)](https://docs.google.com/spreadsheets/d/14z29f7zoM768wOocg4_s-jsni6DjTHF-M2v6ijGnbwg/edit?usp=sharing)

Zip file containing 6 different sync drift test songs
[Sync drift tests.zip](https://github.com/user-attachments/files/23315147/Sync.drift.tests.zip)

In general, it is good to test:

CON file, ini file, yargsong, whammy enabled, whammy disabled, audio buffer on/off, chipmunking on/off

- Quickplay, playing until song end, various speeds
- Practice mode, selecting a section, pausing and unpausing, restart section, changing speed
- Calibration with whammy on/off
- Music player, pausing and unpausing, skipping songs and checking memory usage
- Music Library


Assets/Script/Audio/Bass/BassStemMixer.cs
```diff
@@ -15,7 +15,10 @@ namespace YARG.Audio.BASS
     public sealed class BassStemMixer : StemMixer
     {
         private const    float WHAMMY_SYNC_INTERVAL_SECONDS = 1f;
-        private          bool  IsWhammyEnabled => SettingsManager.Settings.UseWhammyFx.Value;
+
+        //The delay which is introduced by the pitch shift effect, in seconds
+        private const float PITCH_DELAY_SECONDS = GlobalAudioHandler.WHAMMY_FFT_DEFAULT / 44100f;
```
wyrdough: Do we not need to use the actual sample rate here? Or are we forcing it to 44100 somewhere and making audio drivers just deal with it?

wyrdough (commented):
People are going to lose their marbles because their calibration is going to be off by a bit. It might be nice to adjust calibration for the delay automatically if we can. What do you think?

polson:
Calibration should remain unchanged because:

1. We turn off whammy during calibration so that there is no delay 

2. In QuickPlay the SongRunner will sync up the track regardless of how much delay is at the beginning of the song (up to a reasonable limit before it can't handle it any more).  If you look at the debug panel you'll see the very beginning of the song speed up until it accounts for the delay.  It does this regardless for most songs anyway to sync up the visuals which are usually ahead.

Assets/Script/Audio/Bass/BassStemMixer.cs
polson: I noticed we hardcode 44100 everywhere else so I just kept that pattern.  It should probably be a constant in GlobalAudioHandler or something

~~44100 is correct because we use a `TempoStream` which resamples the input to 44100 no matter what sample rate the input is~~

Edit: That said, I would like to test this with a 48000hz stem to make sure, so good point

Edit 2: Well, I'll be darned.  It does resample to 44100, but its still out of sync.  Back to draft we go again.

polson:
And I already found a case where it stops working, so back to draft we go

Assets/Script/Audio/Bass/BassStemChannel.cs
```diff
@@ -217,11 +205,6 @@ protected override void DisposeUnmanagedResources()
         {
             _streamHandles.Dispose();
             _reverbHandles.Dispose();
-            if (_sourceHandle != 0)
```
polson, line 220: This is handled in BassStemMixer now, no more need to check for double freeing

Assets/Script/Audio/Bass/BassStemMixer.cs
polson: Yeah you were right, we need to use the actual sample rate, despite the documentation saying otherwise.  Good call!

image

polson:
Should be good now (famous last words)

Assets/Script/Audio/Bass/BassStemMixer.cs
```diff
@@ -127,30 +181,38 @@ protected override double GetVolume_Internal()
...
+            foreach (var channel in channels)
             {
-                // Pause when seeking to avoid desyncing individual stems
-                Pause_Internal();
+                if (!BassMix.MixerRemoveChannel(channel))
+                {
+                    YargLogger.LogDebug("Failed to remove channel from mixer");
+                };
             }
+            AddChannelsToMixer(_stemDatas.ToList());
```
wyrdough: I don't see any reason to allocate a new list here, just pass _stemDatas.

wyrdough (changes requested):
Minor change, otherwise it seems fine (and seems to work in my testing with your click tracks)

Assets/Script/Audio/Bass/BassStemMixer.cs
polson: Done

PR #1194, opened 2025-10-31
Add star power phrase completion whoosh effect

grishhung:
## Changes

Adds the "whoosh" effect when a star power phrase is completed.

https://github.com/user-attachments/assets/9b70bf09-d1ce-4f61-b007-25eb5909df13


Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -123,6 +127,9 @@ public virtual void Initialize(int index, YargPlayer player, SongChart chart, Tr
             var change = ZeroFadePosition - DEFAULT_ZERO_FADE_POS;
             _hudLocation.position = _hudLocation.position.AddZ(change);
 
+            // Must be done after the HUD location is set
+            StarPowerEffect.Initialize();
+
```
wyrdough, line 133: You should disable the GameObject until it's actually needed, otherwise we're running its Update() loop (and the shader) for no benefit whatsoever most of the time.

```suggestion
            StarPowerEffect.Initialize();
            StarPowerEffect.gameObject.SetActive(false);
 ```
```

Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -417,6 +424,11 @@ protected override void UpdateVisuals(double visualTime)
                 CameraPositioner.Scoop();
             }
 
+            if (SettingsManager.Settings.EnableTrackEffects.Value && currentStarPowerAmount > _previousStarPowerAmount)
+            {
+                StarPowerEffect.PlayAnimation();
```
wyrdough, line 431: Now that the effect is disabled by default, you'll need to enable it before calling its PlayAnimation.

```suggestion
                StarPowerEffect.gameObject.SetActive(true);
                StarPowerEffect.PlayAnimation();
```

(you might actually be able to have it activate itself in `PlayAnimation()` since you're calling a method directly rather than relying on Unity stuff, I'm not certain)

Assets/Script/Gameplay/Visuals/TrackElements/StarPowerEffectElement.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public void PlayAnimation()
+        {
+            _animTimestamp = 0f;
+        }
+
+        private void Update()
+        {
+            _animTimestamp += Time.deltaTime;
```
wyrdough, line 42: You should have this disable itself with `gameObject.SetActive(false)` once _animTimeStamp is large enough that the effect is no longer visible.

wyrdough (changes requested):

wyrdough:
Superseded by #1204

PR #1197, opened 2025-11-01
Implement an "Average Offset" text field in the score screen

satellitnorden:
Corresponding YARG.Core PR: https://github.com/YARC-Official/YARG.Core/pull/302

This implements an average offset text in the score screen, by averaging when a player hit a note versus the time of that note across one chart. This can be used together with the calibration tool to fine-tune the calibration in many ways. (:

First time contributor here, so very much possible I overlooked something. I've tried to understand the codebase and this seems like the way to add this "average offset" text, but there may be some context I'm missing here.

I did see that some other fields in BaseStats were serialized and saved in stuff like replays and such, but opted out of adding this new field there, as I failed to see the utility of storing the average offset in a replay file.

Average Offset

### DIFF
--- a/Assets/Prefabs/Menu/ScoreScreen/DrumsScoreCard.prefab
+++ b/Assets/Prefabs/Menu/ScoreScreen/DrumsScoreCard.prefab
@@ -70,12 +70,12 @@ PrefabInstance:
     - target: {fileID: 1007692541323341066, guid: 1b5d8949b9f2d7345a0fed53e687c843,
         type: 3}
       propertyPath: m_Size
-      value: 0.9047619
+      value: 0.8333333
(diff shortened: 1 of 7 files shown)


wyrdough:
Thanks for this, it seems like it will be super useful. I looked at the code and it all looked reasonable, I just need to do some testing before merging this. I'm hoping I'll be able to get to it sometime in the next couple of days

satellitnorden:
Awesome, thanks for taking a look! Lemme know if you have any questions or want me to make any modifications. (:

wyrdough:
All looks good to me. Given that #1222 is in the pipeline your code probably won't stick around for very long, but I think it's useful enough to merge anyway so that we have something in the meantime.

PR #1199, opened 2025-11-03
Converted playlists to be a filter on the main library.

AWhineOfKarens:
Demo: https://youtu.be/Ih-gIA0yaK8

+ User can select multiple playlists which joins the playlists as a single filter
  + Confirm selects/deselects checkboxes
  + Back button exits the filter selection menu
+ MusicLibrary sorts and other features now work on playlists
+ Playlist option on MusicLibrary is now a PlaylistManager that allows (create, rename, delete, sort).
  + Playlists can be sorted by A-Z,Z-A, Artist A-Z, Artist Z-A

First contribution to this project, so feedback as to your styles and conventions welcome.


AWhineOfKarens:
Note:  "regarding your pull request i think you are committing some unintended changes for the vocaltrack stuff reverting part of [#1188](<https://github.com/YARC-Official/YARG/pull/1188>) please review the changes in your pull request to make sure it's only stuff you actually want to change, cause that seems very unrelated to the actual request"

Resolution: This comment is correct and was my mistake trying to clean up my initial fork.

wyrdough:
I did a quick test just to see it in action and I think that this is generally a good idea. There are a couple of minor UX decisions I'm going to have to think on a bit because I'm not quite sure yet how I feel about them, but overall this is great work. 

I also took a cursory look at the code and nothing major stood out to me as a problem, so I don't expect any big changes will be required to make this mergeable.

Thanks for the PR, we can always use the help!

AWhineOfKarens:
Glad to hear @wyrdough.  Happy to take feedback on the UI.  I can see a few other options, from moving the playlist manager to a separate dedicated application for major edits, with minor edits in YARG - to making the manager part of the main menu rather quickplay - to just trying to keep it very simple.  The same with playlist selection.  100% worth thinking about and discussing; particularly to figure out how players will use the feature.

For me, I've already wrote some separate scripts to automatically create playlists based on score.  Things like, within 1% of scoring a Gold Star, or a greater than 99.9% towards a full combo.  But these are not meant to be integrated.  Just run as cron jobs to keep the playlists updated nightly.

LMK your thoughts when you're ready.

wyrdough:
Sorry for the delay in getting back to you. I haven't forgotten about this, it's just that things have been hectic the past couple of weeks. It seemed like people were generally OK with your demo video, so I'll try to make some time to do a full code review and test soon.

AWhineOfKarens:
Absolutely no worries on that; pretty sure that Unity upgrade was just a little higher priority ;).  I'll be hear to support anytime.  It's also possible to put it in when ready and let people use the feature for a little bit to solicit some feedback on the interface.  Sometimes just using it leads to good ideas.

Dansla116:
Not sure on the best way to get this past the finish line right now.  A few things have changed especially with the New UI.

- The buttons within the search bar are no more.  They've been mostly replaced by the filters menu.  On Discord, there have been some questions about the missing search buttons, but no one has really argued or pushed for them to come back as far as I know.
So if we merge this, that functionality will be gone.  The whole popup menu, multiple checkboxes for playlists, resulting library filtered down to those playlists.  **The good news is**, I think this actually follows a lot of the same logic in the new filters menu.  It's just, I didn't think of adding a filter for user-created playlists.

- Maybe I'm misunderstanding, but the "More Options > Sort By..." menu does not work in playlists for me.  Not when I pulled and built this commit.  And not when I merged into latest.  I'll look more closely into fixing that up.

- The Rename, Delete, Sort A-Z (Z-A), Sort Artist A-Z (Z-A) playlist menu options all survived the merge and still work fine.

There were some other issues with stale popup menus, but it's not broken in your commit (only after merge) so I'm thinking that is either a a New UI issue or a merge issue.

If you want, I could open a helper-PR now with what I currently have.  Or I could keep working on this to get **1)** a Playlists in filter menu and **2)** that "Sort By..." in Playlist Manager working.

polson:
Continued in https://github.com/YARC-Official/YARG/pull/1333

PR #1204, opened 2025-11-12
Add star power phrase completion whoosh effect (revision)

grishhung:
Revision of https://github.com/YARC-Official/YARG/pull/1194/files


wyrdough:
Great work, thanks!

I'm going to go ahead and merge this, but we probably do need to tweak the visuals a bit in the future so that it stands out more against the solo phrase background.

And as a note to myself, the unison SP awards need to be reworked. This PR makes it obvious that unisons are granting SP in two separate steps, so we probably need to delay the solo SP award until the unison phrase completes so that unison bonus and the normal SP award come as one unit.

PR #1213, opened 2025-11-14
Make band multiplier and doubled individual multiplier mutually exclusive

grishhung:
## Notes

> This change is primarily being done to prevent players from thinking that the band multiplier also applies on top of their own individual multiplier.

Changes score multiplier meters while star power is active to show doubled amounts (e.g., 8x instead of 4x or 12x instead of 6x) only if there are more than ony player. The actual internal multiplier (8x) remains the same. In multiplayer, the score shown is the result of the band multiplier being applied, while the per-player doubling applies only to the individual track.

### 8x case

- Tested with only one guitar player (8x)
- Tested with only one drums player (8x)
- Tested with only one pro keys player (8x)
- Tested with only one vocals player (8x)

### 4x case

- Tested with a guitar, drums, and bass player (4x)
- Tested with a guitar and vocals player (4x)
- Tested with two vocals players (4x)

## Gallery

Screenshot 2025-11-13 at 5 16 05 PM


Assets/Art/Fonts/RedHatDisplay/RedHatDisplay-SemiBold.asset
```diff
@@ -5620,7 +5620,7 @@ Material:
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
   m_Name: RedHatDisplay-SemiBold SDF Material
-  m_Shader: {fileID: 4800000, guid: dd89cf5b9246416f84610a006f916af7, type: 3}
+  m_Shader: {fileID: 4800000, guid: 2b50558a6f93646ee875054b4fa9eedd, type: 3}
```
wyrdough, line 5623: Did you intend to change this?

Assets/Script/Gameplay/Player/BasePlayer.cs
```diff
@@ -126,13 +129,14 @@ protected void Start()
             }
         }
 
-        protected void Initialize(int index, YargPlayer player, SongChart chart, int? lastHighScore)
+        protected void Initialize(int totalPlayers, int index, YargPlayer player, SongChart chart, int? lastHighScore)
```
wyrdough: There's not really any need to pass totalPlayers around and have a TotalPlayers field on the player class. GameManager has a list of all the players actually playing already, so you can just expose the count as a property of GameManager since the players already have a reference to GameManager.

Assets/Script/Gameplay/GameManager.Loading.cs
```diff
@@ -377,6 +378,9 @@ private void CreatePlayers()
             {
                 _players = new List<BasePlayer>();
 
+                // Get total players (needed for setting UI values)
+                int totalPlayers = YargPlayers.Count(player => !player.SittingOut);
```
wyrdough: You can avoid this by adding something like `public int TotalPlayers => _players.Count;` to GameManager.cs. The track players don't need that information until after everything is fully initialized anyway.

Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -386,7 +386,13 @@ protected override void UpdateVisuals(double visualTime)
             TrackMaterial.GrooveMode = groove;
             TrackMaterial.StarpowerMode = stats.IsStarPowerActive;
 
-            ComboMeter.SetCombo(stats.ScoreMultiplier, maxMultiplier, stats.Combo);
+            // In multiplayer, don't double the score multiplier in the strikeline element
+            // Otherwise, it looks like the band multiplier applies on top of the score multiplier
+            int displayMultiplier = TotalPlayers > 1 && stats.IsStarPowerActive
```
wyrdough: With the `TotalPlayers` property on `GameManager` instead, you can switch this to `int displayMultiplier = GameManager.TotalPlayers > 1 && stats.IsStarPowerActive`...

wyrdough (changes requested):

Assets/Art/Fonts/RedHatDisplay/RedHatDisplay-SemiBold.asset
grishhung: Yeah. At some point, the shader got bumped back to using the one that doesn't fade out properly on vocals.

Though I suppose this should probably be a whole separate font asset. I'll make that change in a follow-up commit before this gets merged in.

Assets/Script/Gameplay/Visuals/CameraPositioner.cs
```diff
@@ -246,7 +245,6 @@ private void RaiseHighway(bool isGameplayStart)
         // NOTE: Requires SONG_END_DELAY; will not animate until https://github.com/YARC-Official/YARG/pull/993 is in.
         private void LowerHighway(bool isGameplayEnd)
         {
-            _scoop?.Kill();
```
wyrdough, line 249: I think we do actually need this one if we are to avoid problems when the player fails a song just as the scoop animation is playing.

Assets/Script/Gameplay/GameManager.Loading.cs
wyrdough: This also needs to be removed.

wyrdough (changes requested):

wyrdough (approved):

PR #1214, opened 2025-11-14
Added playlist information in files exported for OuvertBot and plain …

bobschlomboggi:
I created this branch to add the playlist information for each song into the javascript file that OuvertBot needs.  Specifically, for streamers playing YARG that have multiple versions of the same song, the streamer will know which version the requester wants to see/hear without having to bring the stream to a halt to ask the requester


bobschlomboggi:
I've only just realized that I probably didn't go about submitting this change in the proper way.  If you need me to submit it differently, please let me know what to do

PR #1217, opened 2025-11-15
Add dynamics hit effects

grishhung:
## Notes

Added hit effects for correctly hit dynamics.

- Ghosts - Black out the inside of the fret, but keep lighting and fret color the same.
- Accents - Completely white out the inside and outside of the fret and multiply the light intensity by 5, but keep the light color the same.

Note that ghost note hit effects do not interrupt accent or normal notes, but normal notes do interrupt accents. This is because if the brightness decreases suddenly before the full 0.2 second animation plays out, it is a bit jarring, but the inverse is true during rolls with accents.

Also switched the solo box font from Barlow to Red Hat Display.

## Gallery

https://github.com/user-attachments/assets/60f7a6e2-cfdf-4275-a2e2-7b553813debe

## Follow-up

- Need to add an accent flame made by Venerabela to help make normal hits and accent hits a bit more distinct.
- Need to add effects for when a ghost note is hit too hard or an accent note is hit too soft (currently, if the velocity is wrong, the normal effect will play).

### DIFF
--- a/Assets/Prefabs/Gameplay/HUD/TrackView.prefab
+++ b/Assets/Prefabs/Gameplay/HUD/TrackView.prefab
@@ -560,8 +560,8 @@ MonoBehaviour:
       m_Calls: []
   m_text: 
   m_isRightToLeft: 0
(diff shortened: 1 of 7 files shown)


wyrdough (approved):
need to test, but code looks fine

PR #1218, opened 2025-11-16
Playlist information for OuvertBot

bobschlomboggi:
Added code to include playlist information for each song in files generated for OuvertBot and its plain text equivalent.

### DIFF
--- a/Assets/Script/Song/SongExport.cs
+++ b/Assets/Script/Song/SongExport.cs
@@ -19,6 +19,9 @@ namespace YARG.Song
             [JsonProperty("Album")]
             public string album;
 
+            [JsonProperty("Playlist")]
+            public string playlist;
+
             [JsonProperty("Genre")]
             public string genre;
 
@@ -46,7 +49,8 @@ namespace YARG.Song
                 {
                     string artist = RichTextUtils.StripRichTextTags(song.Artist);
                     string name = RichTextUtils.StripRichTextTags(song.Name);
-                    output.WriteLine($"{artist} - {name}");
+                    string playlist = RichTextUtils.StripRichTextTags(song.Playlist);
+                    output.WriteLine($"{artist} - {name} from {playlist}");
                 }
                 output.WriteLine("");
             }
@@ -65,6 +69,7 @@ namespace YARG.Song
                     songName = RichTextUtils.StripRichTextTags(song.Name),
                     artistName = RichTextUtils.StripRichTextTags(song.Artist),
                     album = RichTextUtils.StripRichTextTags(song.Album),
+                    playlist = RichTextUtils.StripRichTextTags(song.Playlist),
                     genre = RichTextUtils.StripRichTextTags(song.Genre),
                     charter = RichTextUtils.StripRichTextTags(song.Charter),
                     year = RichTextUtils.StripRichTextTags(song.UnmodifiedYear),
--- a/Assets/Script/Song/SongExport.cs
+++ b/Assets/Script/Song/SongExport.cs
@@ -109,3 +109,4 @@ namespace YARG.Song
             File.WriteAllText(path, json);
         }
     }
+}
--- a/Assets/Script/Song/SongExport.cs
+++ b/Assets/Script/Song/SongExport.cs
@@ -70,8 +70,23 @@ namespace YARG.Song
             var songs = new List<OuvertSongData>();
 
             // Convert SongInfo to OuvertSongData
-            foreach (var song in SongContainer.Songs)
+             foreach (var song in SongContainer.Songs)
             {
+                if(RichTextUtils.StripRichTextTags(song.Playlist) == "Unknown Playlist")
+                {
+                 songs.Add(new OuvertSongData
+                {
+                    songName = RichTextUtils.StripRichTextTags(song.Name),
+                    artistName = RichTextUtils.StripRichTextTags(song.Artist),
+                    album = RichTextUtils.StripRichTextTags(song.Album),
+                    genre = RichTextUtils.StripRichTextTags(song.Genre),
+                    charter = RichTextUtils.StripRichTextTags(song.Charter),
+                    year = RichTextUtils.StripRichTextTags(song.UnmodifiedYear),
+                    songLength = (ulong)song.SongLengthMilliseconds
+                });
+                }
+                else
+                {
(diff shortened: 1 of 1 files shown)


Assets/Script/Song/SongExport.cs
```diff
@@ -46,7 +49,8 @@ public static void ExportText(string path)
                 {
                     string artist = RichTextUtils.StripRichTextTags(song.Artist);
                     string name = RichTextUtils.StripRichTextTags(song.Name);
-                    output.WriteLine($"{artist} - {name}");
+                    string playlist = RichTextUtils.StripRichTextTags(song.Playlist);
+                    output.WriteLine($"{artist} - {name} from {playlist}");
```
wyrdough: Asking, not telling: Wouldn't it be better to check if the playlist value is empty and use `{artist} - {name}` in that case?

Assets/Script/Song/SongExport.cs
```diff
@@ -65,6 +69,7 @@ public static void ExportOuvert(string path)
                     songName = RichTextUtils.StripRichTextTags(song.Name),
                     artistName = RichTextUtils.StripRichTextTags(song.Artist),
                     album = RichTextUtils.StripRichTextTags(song.Album),
+                    playlist = RichTextUtils.StripRichTextTags(song.Playlist),
```
wyrdough: Similarly here, I believe if you set this even if it's an empty string the newtonsoft json writer will output it as an attribute with an empty string like `Playlist: ""` rather than just leaving it out entirely.

wyrdough (commented):
I don't have a problem with this (assuming it doesn't break Ouvert, which I know nothing about), just a question on a design decision...

Assets/Script/Song/SongExport.cs
bobschlomboggi: Im away from my computer at the moment, give me a couple hours and I'll test a couple things.

You raise an excellent point that I didn't consider until now, I'll keep you posted.

bobschlomboggi: As I understand it (and please correct me if I'm wrong), we can specify a folder as a 'root' for the entire song library, which contains playlist folders, which in turn contain song folders.  When a song following that folder structure is scanned, YARG will automatically place it in a playlist with the same name as the playlist folder (or in a playlist with a name specified by the song.ini).  

However, if a song folder is not placed into a playlist folder and is instead placed directly into the root folder, upon scanning YARG will put those songs in a playlist called 'Unknown Playlist', even if that song's song.ini file has the playlist line with no data assigned to it.

That being said, I'm going to try to figure out how to get it from outputting a playlist entitled 'Unknown Playlist'

Assets/Script/Song/SongExport.cs
bobschlomboggi: I think that's a good idea, I've adjusted code to reflect this.

bobschlomboggi (commented):
Plaintext list output has been updated to not provide playlist information for songs not in a playlist.

Json list outputs playlist information correctly for songs in playlists.  However, for songs not contained in a playlist, output lists playlist for those songs as either 'Unknown Playlist' (if line 80 of the current draft is included in the code) or 'null' (if line 80 is excluded).  That being said, it never outputs an empty field for the playlist entry, at least to my knowledge

wyrdough:
> Plaintext list output has been updated to not provide playlist information for songs not in a playlist.
> 
> Json list outputs playlist information correctly for songs in playlists. However, for songs not contained in a playlist, output lists playlist for those songs as either 'Unknown Playlist' (if line 80 of the current draft is included in the code) or 'null' (if line 80 is excluded). That being said, it never outputs an empty field for the playlist entry, at least to my knowledge

If Ouvert can handle `null` or you think it's better to have them as `Unknown Playlist` I don't have a strong opinion, but if you do want to have the serializer omit fields that are null, change the `SerializeObject` call to the following (untested, but should work):
```
            var json = JsonConvert.SerializeObject(songs, Formatting.Indented, new JsonSerializerSettings
            {
                NullValueHandling = NullValueHandling.Ignore
            });
```

bobschlomboggi:
Having a null field does not break Ouvert, so I added conditionals to prevent outputting a Playlist entry into the json for songs not in playlists.  Also, I deleted a curly brace by accident so I added it back in.

Assets/Script/Song/SongExport.cs
```diff
@@ -58,23 +70,43 @@ public static void ExportOuvert(string path)
...
                     genre = RichTextUtils.StripRichTextTags(song.Genre),
                     charter = RichTextUtils.StripRichTextTags(song.Charter),
                     year = RichTextUtils.StripRichTextTags(song.UnmodifiedYear),
                     songLength = (ulong)song.SongLengthMilliseconds
                 });
+                }
             }
```
wyrdough: Indentation got a bit messed up here. Once that's fixed we'll be good to go.


```suggestion
            foreach (var song in SongContainer.Songs)
            {
                if(RichTextUtils.StripRichTextTags(song.Playlist) == "Unknown Playlist")
                {
                    songs.Add(new OuvertSongData
                    {
                        songName = RichTextUtils.StripRichTextTags(song.Name),
                        artistName = RichTextUtils.StripRichTextTags(song.Artist),
                        album = RichTextUtils.StripRichTextTags(song.Album),
                        genre = RichTextUtils.StripRichTextTags(song.Genre),
                        charter = RichTextUtils.StripRichTextTags(song.Charter),
                        year = RichTextUtils.StripRichTextTags(song.UnmodifiedYear),
                        songLength = (ulong)song.SongLengthMilliseconds
                    });
                }
                else
                {
                    songs.Add(new OuvertSongData
                    {
                        songName = RichTextUtils.StripRichTextTags(song.Name),
                        artistName = RichTextUtils.StripRichTextTags(song.Artist),
                        album = RichTextUtils.StripRichTextTags(song.Album),
                        playlist = RichTextUtils.StripRichTextTags(song.Playlist),
                        genre = RichTextUtils.StripRichTextTags(song.Genre),
                        charter = RichTextUtils.StripRichTextTags(song.Charter),
                        year = RichTextUtils.StripRichTextTags(song.UnmodifiedYear),
                        songLength = (ulong)song.SongLengthMilliseconds
                    });
                }
            }
```

wyrdough (changes requested):

PR #1222, opened 2025-11-18
Add "Advanced Statistics" section to Scorecard 

ShawnHardwick:
Corresponding YARG.Core PR: https://github.com/YARC-Official/YARG.Core/pull/307

Props to https://github.com/YARC-Official/YARG/pull/1197 for some inspiration.

This PR implements a new section on the scorecard that details:
 - Accuracy (mean average of notes hit)
 - Precision (standard deviation of notes hit)
 - Score Judgement (Perfect, Great, Good, Poor)  
 
20251117_223704_YARG

This new Advanced Statistics section and the calculations handled for OnNoteHit are disabled by default and can be enabled with the "Enable Advanced Statistics on Scorecard" option under Experimental. Feel free to move this to whatever section you'd like.

Score judgement can also be customized as part of the Engine presets.

Limitations/caveats:
- When testing, I was receiving invalid game replay results.
- Vocal is not supported, did no testing for this instrument.
- Didn't test for dynamic windows
- I tried to implement some validation that the score judgement values were reasonable (Perfect < Great < Good < Poor), but there was no settings validation mechanism. In fact, the Confirm button the settings page doesn't do anything and all settings are saved immediately. The current behavior for validation in this PR is that an error dialog will appear if an engine preset is invalid.

image


ShawnHardwick:
Tested on Unity `6000.2.12f1`. Not sure the workflow with the YARG.Core submodule works. I'm sure I could fix it once the YARG.Core PR is merged.

wyrdough:
This looks like a great extension to the more basic functionality in #1197. I went ahead and merged that one since this is going to take some rework to deal with the issues with replay validation and sanity checking the judgement preset values in the settings screen. I don't think the conflict resolution should be anything difficult since you can replace everything from #1197 with yours since it's a superset of that functionality.

It may be a couple of days before I can really dig in and remind myself of how some of that will have to work, but if you're game I am as well. I'd love to have this in so we can build upon it later. People keep bringing up the idea of showing judgements during gameplay and your work will give us a great foundation to build upon to make that happen.

ShawnHardwick:
Sure thing. Just tell me what I can help with. I know nothing about the reply validation stuff. Long ago I requested for timing judgements to be added to the UI during playtime (like many other rhythm games), but that was out of my league. Also, I didn't really build this with that in mind though it is a step in that direction.

wyrdough:
Sorry for the unexpectedly long delay on this, work has been..intense..the past couple of weeks so I haven't really had any time for YARG stuff. I'm hoping things will calm down in the next few days so I can get to this next week.

Ape:
I just wanted to note that median offset might be better statistic that mean offset. Especially when trying to find the offset value for sync purposes.

ShawnHardwick:
> Sorry for the unexpectedly long delay on this, work has been..intense..the past couple of weeks so I haven't really had any time for YARG stuff. I'm hoping things will calm down in the next few days so I can get to this next week.

No rush. I appreciate the response.

ShawnHardwick:
> I just wanted to note that median offset might be better statistic that mean offset. Especially when trying to find the offset value for sync purposes.

Assuming a normal bell curve, I think that mean is best representing accuracy and therefore sync offset. I also have to factor in CPU and memory usage when calculating these things to not affect performance. I was using an absolute mean deviation instead of standard decision initially but was worried about performance and felt like it was still a decent representation of precision. It would've required storing every timing values in an array for each note and player and performing a calculation at end of song.

wyrdough:
> > I just wanted to note that median offset might be better statistic that mean offset. Especially when trying to find the offset value for sync purposes.
> 
> Assuming a normal bell curve, I think that mean is best representing accuracy and therefore sync offset. I also have to factor in CPU and memory usage when calculating these things to not affect performance. I was using an absolute mean deviation instead of standard decision initially but was worried about performance and felt like it was still a decent representation of precision. It would've required storing every timing values in an array for each note and player and performing a calculation at end of song.

Memory usage is pretty low to begin with, so as long as you preallocate the data structure (array does this implicitly, but list you need to give an explicit size to avoid extending the list and generating garbage) before gameplay starts and keep it alive until the song is over, I don't really have a problem with storing the timing offsets for each note and then calculating when showing the score screen if you think it would be a better way to go. The main thing is to avoid generating garbage since GC runs can cause frame pacing problems on low end devices.

I don't have a strong opinion on median vs. mean since the maximum offset should be limited by the size of the hit window so the impact of single outliers can't throw the calculation off *that* much except in pathological cases where the chart only has a few notes to begin with.

ShawnHardwick:
Had a hell of a time getting this to merge from `dev`. In doing so, it introduced some bugs, mostly around the prefabs. I gotta take some time to test it before submitting for review again.

wyrdough:
Coincidentally, I was just thinking of this PR last night. Sorry it's been so long. I do really want to get this in in some form. I think we've settled on having an advanced statistics page on the scorecard accessed by pressing some button or other where the stats here and the histogram from #1302 will go.

I really appreciate your patience. We should be able to get to this very soon.

ShawnHardwick:
Do you think I should invest anymore time into this MR then? Not sure if any of the code will be needed as an MR or if it's going to be refactored into a different feature branch.

wyrdough:
It looks to me like the this PR is pretty much what it needs to be in terms of code, we're just going to add a disabled by default section to the score card prefab that can be swapped for the present stats area when the player taps blue or orange or whatever we end up deciding on. (which I only just now realized will mean that mouse click will have to flip all the cards)

I do want to have a bit of a think on the validation issue, though, and I also need to decide what the best strategy for merging this really is. Could be merged as-is (well, I don't know how messed up the prefabs are since I haven't actually tried to run it recently) and then the "advanced flip" could be implemented later, I suppose. I'm a bit brain fried from the last couple of days, though, so give me a day or two to get my head fully back into the game, lol.

ShawnHardwick:
No rush man. I appreciate the work. Just tell me if there is anything I can do for this.

polson:
Closed in favor of https://github.com/YARC-Official/YARG/pull/1222#:~:text=Show%20overhit%20stats%20for%20each%20drum%20%231300

wyrdough:
> Closed in favor of #1222

Uh, this is #1222?

PR #1227, opened 2025-12-01
add toggle for venue post processing

4dispel:
toggle for venue post processing in graphic settings and quick pause settings.

some charts give me headaches from all the post processing effects so i would like to be able to just turn them off directly while playing the chart and just turn it back on the next chart.

toggling post processing while mid song will only affect the post  processing the next time a new effect would be processed, as that was a lot simpler to implement and i just wanted something that works.

image

image

### DIFF
--- a/Assets/Prefabs/Gameplay/HUD/Pause/QuickSettings.prefab
+++ b/Assets/Prefabs/Gameplay/HUD/Pause/QuickSettings.prefab
@@ -5,6 +5,7 @@ PrefabInstance:
   m_ObjectHideFlags: 0
   serializedVersion: 2
   m_Modification:
+    serializedVersion: 3
     m_TransformParent: {fileID: 4220475145157243466}
     m_Modifications:
     - target: {fileID: 375337769934742335, guid: 81863a63e0d96cf459c263ee0ea6f98e,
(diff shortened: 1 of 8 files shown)


wyrdough:
I like this idea, but I'm going to have to think on how best to make it take immediate effect when I get a minute because I suspect people will get confused if it doesn't take effect as soon as it is toggled from the quick settings menu.

..maybe have `CameraManager` save the value of the `VenuePostProcessing` setting and subscribe to `SettingsManager.SettingChanged` during initialization and have the callback function call `SetCameraPostProcessing` if the `VenuePostProcessing` value has changed? For going on to off it should just work™ without much trouble, though toggling on could present more of a challenge if Next/Previous/CurrentEffect aren't being updated while the setting is turned off.

(Note that this is just me thinking out loud, not me saying that this is definitely the best way to go about it, feel free to discuss here or on Discord)

4dispel:
i tried basically exactly your suggestion, I've never worked with events in C# before that was surprisingly simple.
Next/Previous/CurrentEffect are still being updated while the setting is turned off, so going from off to on seems to work.

this has some issues on the frame the game is paused going on -> off changing the post processing properties without any time passing.
and going off -> on only seems to show up as soon as the game starts running again.

i'm also not sure if starting the current effect with the tick it's supposed to have but default time is actually reasonable in CameraManager.SetPostProcessingEnabled due to my limited understanding of how the venue events actually work. i just changed the time because all the duration calculations use that.

4dispel:
i actually have no idea how the initialization code got lost, it wasn't commited for some reason and was gone when i restarted my pc today

wyrdough:
Thanks for the quick update! Sorry I haven't been able to review this yet, I should have some time to do that in the next day or two.

wyrdough (approved):

PR #1231, opened 2025-12-05
Fix background image/video rendering

theli-ua:
It was hidden under opaque venue output texture (that is unused when using song-specific background)

The only change in gameplay scene is a new field for venue  output on background  manager and disabling gameobject of venue output.

### DIFF
--- a/Assets/Scenes/Gameplay.unity
+++ b/Assets/Scenes/Gameplay.unity
@@ -1444,6 +1444,7 @@ MonoBehaviour:
   _videoPlayer: {fileID: 218824352}
   _backgroundImage: {fileID: 1773048783}
   _backgroundDimmer: {fileID: 1038164713}
+  _venueOutput: {fileID: 879254558}
 --- !u!114 &318454377
 MonoBehaviour:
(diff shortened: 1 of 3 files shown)


wyrdough (approved):

wyrdough:
FML, testing reveals that this breaks video and image backgrounds on windows (both vulkan and d3d)

theli-ua:
I've removed all the unnecessary line breaks/etc from the gameplay scene to make this tiny. I wish unity would have a consistent representation of those and stop making scene diffs be so bad

theli-ua:
Any update here? I'm not sure how could it possibly break video/image backgrounds on windows. Could you double check that you had them enabled?

wyrdough:
So you were right that it was a me problem, which makes sense given how strange it would be to have different behavior between Windows and Linux, but not quite what you thought. At some point background opacity in my dev settings got set to 0.

Which does remind me that it would be nice if venues were also affected by that setting, but that's a todo list item, nothing that needs to be fixed for this PR. Sorry for the delay, will merge shortly.

PR #1233, opened 2025-12-20
limit venue lighting to venue layer

4dispel:
this change iterates over all light object in the loaded venue to set the culling mask to only render the lighting on the actual venue.
this fixes the lighting for venues with extremely bright directional lights affecting the highway.
this is the same method of culling as currently used by gameplay lighting to exclude the venue layer just reversed.

an example of a brightened highway before the change
image

### DIFF
--- a/Assets/Script/Venue/BundleBackgroundManager.cs
+++ b/Assets/Script/Venue/BundleBackgroundManager.cs
@@ -20,7 +20,9 @@ namespace YARG.Venue
         public const string BACKGROUND_SHADER_BUNDLE_NAME = "_metal_shaders.bytes";
         public const string BACKGOUND_OSX_MATERIAL_PREFIX = "_metal_";
 
-        private const int VENUE_LAYER_NUMBER = 9;
+        private const string VENUE_LAYER_NAME = "Venue";
+
+        private int _venueLayerNumber = -1;
 
         // DO NOT CHANGE the name of this! I *know* it doesn't follow naming conventions, but it will also break existing
         // venues if we do change it.
@@ -36,6 +38,7 @@ namespace YARG.Venue
         {
             // Move object out of the way, so its effects don't collide with the tracks
             transform.position += Vector3.forward * 10_000f;
+            _venueLayerNumber = LayerMask.NameToLayer(VENUE_LAYER_NAME);
         }
 
         public void SetupVenueCamera(GameObject bgInstance)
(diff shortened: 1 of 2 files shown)


wyrdough:
Nice! All looks reasonable to me, I should hopefully have some time to test this weekend.

Assets/Script/Venue/BundleBackgroundManager.cs
```diff
@@ -20,6 +20,8 @@ public class BundleBackgroundManager : MonoBehaviour
         public const string BACKGROUND_SHADER_BUNDLE_NAME = "_metal_shaders.bytes";
         public const string BACKGOUND_OSX_MATERIAL_PREFIX = "_metal_";
 
+        private const int VENUE_LAYER_NUMBER = 9;
```
wyrdough: Given that Unity doesn't guarantee (afaik) that the layer number will never get silently changed, I think it would be better define the _name_ of the layer mask here and initialize the _number_ in `Awake()` with a `LayerMask.NameToLayer()` call.

If I'm wrong about that, lmk and I'll merge as-is.


```suggestion
        private       int    VENUE_LAYER_NUMBER = -1;
        private const string VENUE_LAYER_NAME = "Venue";
```

Assets/Script/Venue/BundleBackgroundManager.cs
wyrdough, line 42:
```suggestion
            
            VENUE_LAYER_NUMBER = LayerMask.NameToLayer(VENUE_LAYER_NAME);
        }
```

Assets/Script/Venue/BundleBackgroundManager.cs
```diff
@@ -54,6 +56,18 @@ public void SetupVenueCamera(GameObject bgInstance)
             }
         }
 
+        public void LimitVenueLights(GameObject bgInstance)
+        {
+            int venueLayer = 1 << VENUE_LAYER_NUMBER;
```
wyrdough:
```suggestion
            if (VENUE_LAYER_NUMBER == -1)
            {
                return;
            }
            
            int venueLayer = 1 << VENUE_LAYER_NUMBER;
```

wyrdough (changes requested):
It would be nice to make things more robust in case unity does something silly in the future and the layer number silently changes.

(Don't take the suggestions as gospel, I just banged them out in the github text box without testing anything)

4dispel:
i agree that using the name is better in case it breaks, and also makes the code a easier to understand just from reading it.
i did look for a bit but didn't find that method myself. was happy enough that i found out the bit shift stuff from the unity documentation.
only change i made to the suggestions is renaming the venue layer  number since it's no longer constant.
tested and still works the same as with just the index.

PR #1245, opened 2025-12-30
Fix some minor resource leaks (mostly around render textures)

theli-ua:
From documentation:
    As with other "native engine object" types, it is
    important to pay attention to the lifetime of any
    render textures and release them when you are finished
    using them, as they will not be garbage collected like normal managed types.


wyrdough (approved):
approved pending testing

PR #1247, opened 2025-12-30
Fix non-deterministic high-score selection in SQLite queries

Ape:
The previous queries relied on GROUP BY without proper aggregation, which lets SQLite return an arbitrary row for each song; that could be a lower score/stars/percent than the true best. This commit replaces those grouped queries with deterministic lookups using ORDER BY ... LIMIT 1.

### DIFF
--- a/Assets/Script/Scores/ScoreDatabase.cs
+++ b/Assets/Script/Scores/ScoreDatabase.cs
@@ -220,17 +220,26 @@ namespace YARG.Scores
         public List<GameRecord> QueryBandHighScores()
         {
             return Query<GameRecord>(
-                @"SELECT *, MAX(BandScore) FROM GameRecords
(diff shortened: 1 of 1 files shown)


wyrdough:
Thanks for the PR, this has been a thorn in our side for a while now.

It's been a long while since I last looked at the queries, but my recollection is that using a subselect in SQLite is rather slow. Have you done any benchmarking to quantify the difference in performance?

Ape:
The queries are filtering with song checksum, which has an index, so it would be slow only if there is a huge amount of scores for a given single song.

I wouldn't compare it to the previous version, since that one was just doing the wrong thing.

There's a potential optimization with a partitioning CTE, but that requires a later SQLite version.

I haven't measured the actual timing, but I don't believe it would be any issue in any realistic scenario.

PR #1251, opened 2026-01-04
Fix issues with hud position not saving for top elements

polson:
This fixes issues with the top elements not saving position when manually repositioned

This happened because those elements depended on the position of the track, which can change, so it was being reset each frame

Fix is to check if the user moved the hud element, and in that case stop updating the position each frame

### DIFF
--- a/Assets/Script/Gameplay/HUD/TrackView.cs
+++ b/Assets/Script/Gameplay/HUD/TrackView.cs
@@ -50,7 +50,7 @@ namespace YARG.Gameplay.HUD
             Vector2 position = _highwayRenderer.GetTrackPositionScreenSpace(highwayIndex, 0.5f, CENTER_ELEMENT_DEPTH);
             _centerElementContainer.transform.position = position;
 
-            if (_topDraggable?.HasCustomPosition == false)
+            if (_topDraggable != null && _topDraggable.HasCustomPosition == false)
             {
                 // Place top elements at 100% depth plus n screen independent units up to avoid highway overlap
                 var extraOffset = TOP_ELEMENT_EXTRA_OFFSET * Screen.height / 1000f;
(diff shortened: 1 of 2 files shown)


Assets/Script/Gameplay/HUD/TrackView.cs
```diff
@@ -46,10 +50,13 @@ public void UpdateHUDPosition(int highwayIndex, int highwayCount)
             Vector2 position = _highwayRenderer.GetTrackPositionScreenSpace(highwayIndex, 0.5f, CENTER_ELEMENT_DEPTH);
             _centerElementContainer.transform.position = position;
 
-            // Place top elements at 100% depth plus screen independent units up to avoid highway overlap
-            var extraOffset = 8 * Screen.height / 1000f;
-            Vector2 position2 = _highwayRenderer.GetTrackPositionScreenSpace(highwayIndex, 0.5f, 1.0f).AddY(extraOffset);
-            _topElementContainer.position = position2;
+            if (_topDraggable?.HasCustomPosition == false)
```
wyrdough: `DraggableHudElement` derives (circuitously) from `MonoBehavior`, so we shouldn't use the null conditional operator on it since it can cause issues with unity object lifecycle checks. Basically, the underlying unity C++ object may be dead/disposed/whatever you want to call it even though the C# object isn't null and thus throw on an access attempt. Unity has special handling for this, but it only works when you use `== null ` and `!= null`.

```suggestion
            if (_topDraggable != null && _topDraggable.HasCustomPosition == false)
```

wyrdough (changes requested):

Assets/Script/Gameplay/HUD/TrackView.cs
polson: Love that unity messes up language features.  Good catch :)

PR #1252, opened 2026-01-05
Fix issue where missing a star power note broke inputs

polson:
Discord issue: https://discord.com/channels/1086048856678084609/1440742564960403476/1440742564960403476

**To repro this issue:**
1. Create a guitar profile using circular theme
2. Select a song which starts with several star power notes immediately
3. Hit one of the star power notes
3. Miss the next star power note

Observe that inputs no longer work

**The cause:**

When using circular theme, the note model changes when missing a star power sequence.

We have an optimization in place which prevents us from changing note models on notes that have already been hit, since they are hidden.

So when we miss the first note, the models change except for the hidden note.

Immediately after this we try to set the color on the notes.  This causes an exception on the hidden notes because the new note model has not been initialized.

The fix is to check if the note was hit before attempting to change the color.

We also add this logic on all set color calls since it is an easy optimization.

### DIFF
--- a/Assets/Script/Gameplay/Visuals/TrackElements/Drums/FiveLaneDrumsNoteElement.cs
+++ b/Assets/Script/Gameplay/Visuals/TrackElements/Drums/FiveLaneDrumsNoteElement.cs
@@ -114,11 +114,14 @@ namespace YARG.Gameplay.Visuals
                 color = colors.GetNoteStarPowerColor(pad);
             }
 
-            // Set the note color
-            NoteGroup.SetColorWithEmission(color.ToUnityColor(), colorNoStarPower.ToUnityColor());
+            // Set the note color if not hidden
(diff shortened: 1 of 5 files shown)


polson:
Keeping this in draft until we can test it with a chart that starts out with a few measures of starpower notes

I am not 100% sure this will fix the issue so...

wyrdough:
Seems not to break anything and *appeared* to fix the problem in my limited testing, so I'm going to merge this

wyrdough (approved):

wyrdough:
It occurred to me after the fact that the underlying issue is probably something in `NoteGroup` not null checking properly. I was not able to get it to break outside of very fast sections, so this fix is fine for now. I'm just writing it down so I have a better than zero chance of remembering to look at it in the future.

PR #1254, opened 2026-01-06
add song backgrounds as venue textures

4dispel:
adds song background as render texture to the texture manager

looks for _Yarg_VideoTex in venue shaders

if found will try to load the song background, if it's an image or video
- video: uses the video player normally used to render background video to render to texture instead
- image: renders the background to the texture once on load 
- default: if nothing is found the texture is black with 0 alpha so that the shader author can set a default

demo:
https://youtu.be/eR9T7Ul8WyE

### DIFF
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -7,6 +7,7 @@ using UnityEngine.Rendering.RendererUtils;
 using UnityEngine.UI;
 using UnityEngine.Video;
 using YARG.Core.IO;
+using YARG.Core.Song;
 using YARG.Core.Venue;
 using YARG.Helpers.Extensions;
 using YARG.Settings;
@@ -121,11 +122,13 @@ namespace YARG.Gameplay
 #endif
                     // Hookup song-specific textures
                     var textureManager = GetComponent<TextureManager>();
+                    // Load SongBackground here to determine if textures need to be replaced
+                    var songBackground = GameManager.Song.LoadBackground();
                     foreach (var renderer in renderers)
                     {
                         foreach (var material in renderer.sharedMaterials)
                         {
-                            textureManager.ProcessMaterial(material);
+                            textureManager.ProcessMaterial(material, songBackground?.Type);
                         }
                     }
 
@@ -140,7 +143,7 @@ namespace YARG.Gameplay
 
                     if (textureManager.VideoTexFound())
                     {
-                        SetUpVideoTexture(bgInstance);
+                        SetUpVideoTexture(songBackground);
                     }
 
                     break;
@@ -180,30 +183,20 @@ namespace YARG.Gameplay
             }
         }
 
-        private void SetUpVideoTexture(GameObject bginstance)
+        private void SetUpVideoTexture(BackgroundResult songBackGround)
         {
             var textureManager = GetComponent<TextureManager>();
-            var result = GameManager.Song.LoadBackground();
-            if (result == null || result.Type == BackgroundType.Yarground)
+            if (songBackGround == null || songBackGround.Type == BackgroundType.Yarground)
             {
                 return;
             }
-            //assign background textures after confirming there is a background to render
-            var renderers = bginstance.GetComponentsInChildren<Renderer>(true);
-            foreach (var renderer in renderers)
-            {
-                foreach (var material in renderer.sharedMaterials)
-                {
-                    textureManager.AssignBackgroundTextures(material);
-                }
-            }
-            switch (result.Type)
+            switch (songBackGround.Type)
             {
                 case BackgroundType.Video:
                     //set venue source to song to enable video seeking/pausing features
                     _source = VenueSource.Song;
 
-                    switch (result.Stream)
+                    switch (songBackGround.Stream)
                     {
                         case FileStream fs:
                         {
@@ -218,7 +211,7 @@ namespace YARG.Gameplay
                             VIDEO_PATH = Path.Combine(Application.persistentDataPath, sngStream.Name);
                             using var tmp = File.OpenWrite(VIDEO_PATH);
                             File.SetAttributes(VIDEO_PATH, File.GetAttributes(VIDEO_PATH) | FileAttributes.Temporary | FileAttributes.Hidden);
-                            result.Stream.CopyTo(tmp);
+                            songBackGround.Stream.CopyTo(tmp);
                             _videoPlayer.url = VIDEO_PATH;
                             break;
                         }
@@ -234,7 +227,7 @@ namespace YARG.Gameplay
                     enabled = true;
                     break;
                 case BackgroundType.Image:
-                    var songTex = result.Image.LoadTexture(false);
+                    var songTex = songBackGround.Image.LoadTexture(false);
                     //render image background flipped to match video
                     Graphics.Blit(songTex, textureManager.GetVideoTexture(), new Vector2(1, -1), new Vector2(0, 1));
                     //clean up unused texture
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -51,7 +51,7 @@ namespace YARG.Gameplay
             // We don't need to update unless we're using a video
             enabled = false;
 
-            using var result = VenueLoader.GetVenue(GameManager.Song, out _source);
+            var result = VenueLoader.GetVenue(GameManager.Song, out _source);
             if (result == null)
             {
                 return;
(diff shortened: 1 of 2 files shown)


Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -23,25 +23,31 @@ public class TextureManager : GameplayBehaviour
...
+        private static int _videoTexId = Shader.PropertyToID("_Yarg_VideoTex");
 
         private const double MIN_DB = -100.0;
         private const double MAX_DB = -30.0;
         private const double DB_RANGE = MAX_DB - MIN_DB;
         private const int FFT_SIZE_LOG = 11 /* aka log2(2048) */;
         private const int FFT_SIZE = 1 << FFT_SIZE_LOG;
         private const int FFT_TEXTURE_WIDTH = 512;
+        private const int VIDEO_TEX_WIDTH = 1280;
+        private const int VIDEO_TEX_HEIGHT = 720;
```
theli-ua: Would it be better to allow venue specify desired size? fwiw it could be a tiny screen in the venue

Assets/Script/Gameplay/BackgroundManager.cs
```diff
@@ -175,6 +180,69 @@ private async UniTaskVoid Start()
...
+                case BackgroundType.Image:
+                    var songTex = result.Image.LoadTexture(false);
+                    //render image background flipped to match video
+                    Graphics.Blit(songTex, textureManager.GetVideoTexture(), new Vector2(1, -1), new Vector2(0, 1));
+                    //clean up unused texture
+                    Destroy(songTex);
+                    return;
+            }
+        }
+
```
theli-ua, line 189: Seems a little awkward and inconsistent how this sets up differently from the rest, iterates all materials multiple times, etc. Why not create and assign video texture within `ProcessMateral` like the rest are done. 
Then later in `BackGroundManager` check if `TextureManager` has non-null `VideoTexture` and it it is indeed non-null, then create video player and make it use that texture? that sounds a little more consistent with how the things are being set up right now.

4dispel: that is how it was implemented before [7b955a9](https://github.com/YARC-Official/YARG/pull/1254/commits/7b955a9f7cd9b9e9a933e7f786ffdc7b3ecea997) i changed it so that it doesn't replace the default texture of the material if it doesn't find a background

Assets/Script/Gameplay/TextureManager.cs
4dispel: would probably be good to let the venue decide, i just set some defaults cause i'm not sure what the best way would be to set these from the venue

theli-ua: You could use the resolution of the texture that was already set in the venue itself (as long as its not 1x1)

Assets/Script/Gameplay/BackgroundManager.cs
theli-ua: I see, how about then

* Setup videoplayer/bg, set `hasBg = true`, set `videoplayer.enabled = false`
* Add boolean parameter to `ProcessMaterial` that indicates whether bg/video texture is present for  song and therefore texture should be hooked up
* Also have a ref parameter "usesBgTexture" to `ProcessMaterial` that is only updated (set to true) if a bg texture was used
* if usesBgTexture and bg was video, `videoplayer.enabled = true`

Something like that. Seems like it would be a little more streamlined with the current textures, without multiple iterations of materials

Assets/Script/Gameplay/TextureManager.cs
4dispel: that assumes either just one texture in the venue using it, or that the first texture found is the highest resolution needed.
absolutely doable but will need to be clarified in the documentation that the textures need to be same resolution.

theli-ua: Yes,I was thinking either the first or the largest

Assets/Script/Gameplay/BackgroundManager.cs
4dispel: didn't change it exactly as specified here, but refactored the assigning of materials back into `ProcessMaterial` in [beba957](https://github.com/YARC-Official/YARG/pull/1254/commits/beba9573e50c7add5accc798d6f8179e7983fc9d)

Assets/Script/Gameplay/TextureManager.cs
4dispel: [9fe9a20](https://github.com/YARC-Official/YARG/pull/1254/commits/9fe9a20d6bbb35ad020b9a2101f398039e1bc042)
texture now starts at minimum of 256x144, to ignore textures that are 1x1 or similar.
width and height get updated to the respective maximum of the textures found in materials.

4dispel:
updated current behaviour:
now leaves default textures in material instead of setting to black with 0 alpha.
render texture now is scaled to either 256x144 or maximum resolution found among material textures.

updated to have 3 different shader parameters:
- `_Yarg_VideoTex` only gets replaced by video backgrounds
- `_Yarg_ImageTex` only gets replaced by image backgrounds
- `_Yarg_BackgroundTex` gets replaced by either image or video backgrounds

Assets/Script/Gameplay/BackgroundManager.cs
```diff
@@ -175,6 +183,59 @@ private async UniTaskVoid Start()
             }
         }
 
+        private void SetUpVideoTexture(BackgroundResult songBackGround)
+        {
+            var textureManager = GetComponent<TextureManager>();
+            if (songBackGround == null || songBackGround.Type == BackgroundType.Yarground)
```
theli-ua, line 165: It would be  kind cool to load a song specific yarground and render it :joy_cat:

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -86,7 +95,30 @@ protected Texture2D GetSoundTexture()
...
+                _videoTexture = new RenderTexture(_videoTexWidth, _videoTexHeight, 0);
+                _videoTexture.Create();
+            }
+            if (width > _videoTexture.width)
+            {
+                _videoTexture.width = width;
+            }
+            if (height > _videoTexture.height)
+            {
+                _videoTexture.width = height;
```
theli-ua: I think you meant to change height here

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -23,25 +24,33 @@ public class TextureManager : GameplayBehaviour
...
+        private static int _backgroundTexId = Shader.PropertyToID("_Yarg_BackgroundTex");
 
         private const double MIN_DB = -100.0;
         private const double MAX_DB = -30.0;
         private const double DB_RANGE = MAX_DB - MIN_DB;
         private const int FFT_SIZE_LOG = 11 /* aka log2(2048) */;
         private const int FFT_SIZE = 1 << FFT_SIZE_LOG;
         private const int FFT_TEXTURE_WIDTH = 512;
+        private int _videoTexWidth = 256;
+        private int _videoTexHeight = 144;
```
theli-ua: You never update those... so `const`  perhaps?

Assets/Script/Gameplay/TextureManager.cs
4dispel: yeah missed that oops. fixed in [96c5db9](https://github.com/YARC-Official/YARG/pull/1254/commits/96c5db996d4ddaad18d7413780a13fd98cd06888)

Assets/Script/Gameplay/TextureManager.cs
4dispel: i made it not const cause i thought i would change it but i just changed it in the render texture and forgot to make it const again. fixed in [96c5db9](https://github.com/YARC-Official/YARG/pull/1254/commits/96c5db996d4ddaad18d7413780a13fd98cd06888)

Assets/Script/Gameplay/BackgroundManager.cs
```diff
@@ -175,6 +183,59 @@ private async UniTaskVoid Start()
...
+            {
+                return;
+            }
+            switch (songBackGround.Type)
+            {
+                case BackgroundType.Video:
```
theli-ua, line 171: I'm sorry I didn't call this out earlier - but do you think it would make sense to refactor some common code for loading image/video backgrounds into a separate methods and call it here and inside `Start`?

4dispel: image backgrounds are loaded pretty differently in the render texture and the normal background but video loading is just copied code so making that into a method makes a lot of sense. refactored it in [cac5404](https://github.com/YARC-Official/YARG/pull/1254/commits/cac540455d62643207ee53cb499baaca7d60fc9a)

theli-ua (approved):

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -86,7 +95,30 @@ protected Texture2D GetSoundTexture()
...
+                _videoTexture.Create();
+            }
+            if (width > _videoTexture.width)
+            {
+                _videoTexture.width = width;
+            }
+            if (height > _videoTexture.height)
+            {
+                _videoTexture.height = height;
+            }
```
wyrdough, line 111: Would it not be better to get the final width/height before creating the rendertexture? 

Since this isn't happening every frame I don't think it really matters, but it seems like good practice to avoid possibly forcing Unity/the GPU to create the texture a second time and copy/scale contents.

4dispel: fixed in [e4e4de0](https://github.com/YARC-Official/YARG/pull/1254/commits/e4e4de06bc904ee0259805501959c7a7073a7447)

PR #1256, opened 2026-01-08
Add instrument difficulties to ouvert export

polson:
Adds a difficulty ulong to the output json for ouvert

This is similar to the way other games export for ouvert.

There is discussion in the ouvert discord here: https://discord.com/channels/693969788526919832/693970125283524608/1458494738285072414

### DIFF
--- a/Assets/Script/Song/OuvertExport.cs
+++ b/Assets/Script/Song/OuvertExport.cs
@@ -1,0 +1,126 @@
+using System.Collections.Generic;
+using System.IO;
+using Newtonsoft.Json;
+using YARG.Core;
+using YARG.Core.Song;
+using YARG.Core.Utility;
+
+namespace YARG.Song
+{
+    public static class OuvertExport
+    {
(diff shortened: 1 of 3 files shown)


Assets/Script/Song/OuvertExport.cs
```diff
@@ -0,0 +1,126 @@
...
+            if (id == -1)
+            {
+                return 0;
+            }
+
+            return (ulong) song[inst].Difficulties << (id * DIFFS_PER_INSTRUMENT);
+        }
+
+        // Re-map instrument ids, since a ulong can only fit 10 instruments (64 bits / 6 difficulties = 10 instruments)
+        private static int GetInstrumentId(Instrument instrument)
```
wyrdough, line 108: Annoying that we'll have to rework this for any new instruments, but I see from the Discord conversation that it's already done on the other end, so we'll just have to deal with it until change is forced upon us.

PR #1259, opened 2026-01-11
Show high score card also on the first clear

Ape:
Technically it's the high score even if there was no previous score.

Alternatively, we could show something like "Song Cleared!".


wyrdough:
My inner pedant does not think one can set a high score without there being a previous score, but I am not married to that line of thinking if you can get consensus from the community that we should call the first play a high score.

polson:
I would be okay with the "Song Cleared!" message, it feels weird to me get a high score when you hit 1% of the notes.  Just my opinion

wyrdough:
There's been no movement on this for a while, so I'm going to go ahead and close it. Feel free to reopen if mass opinion is against me...

PR #1261, opened 2026-01-13
Add Modifier to make Opens into Green notes

4dispel:
see https://github.com/YARC-Official/YARG.Core/pull/315

### DIFF
--- a/Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
+++ b/Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
@@ -439,6 +439,8 @@ namespace YARG.Menu.DifficultySelect
                 _menuState = State.Main;
                 UpdateForPlayer();
             });
+
+            _navGroup.SelectFirst();
         }
 
         private void CreateHarmonyMenu()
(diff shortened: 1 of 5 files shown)


4dispel:
grafik

in the difficulty select menu there is a scrollbar at the side which works and with mouse you can navigate fine but it doesn't scroll automatically when selecting with a controller

4dispel:
i should probably add the scrollbar as a serialized field, but i didn't want to deal with the git diff on the prefab

wyrdough:
> i should probably add the scrollbar as a serialized field, but i didn't want to deal with the git diff on the prefab

Eh, I think it's probably better this way

Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
```diff
@@ -149,6 +151,32 @@ private void OnEnable()
...
+
+
+            _scrollbar = GetComponentInChildren<Scrollbar>();
+            _navGroup.SelectionChanged += UpdateForSelectionChanged;
+        }
+
+        private void UpdateForSelectionChanged(NavigatableBehaviour navigatableBehaviour,
+            SelectionOrigin selectionOrigin)
+        {
+            int? index = _navGroup.SelectedIndex;
```
wyrdough, line 168: Just to be safe in the face of possible future changes, it would be good to check if `_scrollbar` is null and bail early if it is.

4dispel: done in [a8d221c](https://github.com/YARC-Official/YARG/pull/1261/commits/a8d221c75ddd9827fe51b6f29835a1c022c01b4b)

wyrdough:
Venerabela is going to make an icon for this and I'm going to do some functional testing to make sure it all works for me. Should be able to merge as soon as that's done.

wyrdough (approved):
Did some testing and it works a treat. Great job!

PR #1263, opened 2026-01-14
Chord names

WalterRalts:
Adds Pro Keys chord names and a toggle to disable them in settings.


Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -90,26 +84,18 @@ public void OpenVenueFolder()
             public ToggleSetting ShowActivePlayers { get; } = new(false, ShowActivePlayersCallback);
             public ToggleSetting ShowActiveBots { get; } = new(false, ShowActiveBotsCallback);
 
-            public ToggleSetting ReconnectProfiles  { get; } = new(true);
-            public ToggleSetting AutoCreateProfiles { get; } = new(true);
+            public ToggleSetting ReconnectProfiles { get; } = new(true);
 
             public ToggleSetting ReduceNoteSpeedByDifficulty { get; } = new(true);
 
+            public ToggleSetting LearningGuides { get; } = new(true);
```
wyrdough: Should this really be enabled by default?

Assets/Script/Gameplay/Visuals/TrackElements/ProKeys/ProKeysChordBarElement.cs
```diff
@@ -77,6 +108,519 @@ public void CheckForChordHit()
...
+                chord_name = GetRoot(0);
+                if (chord_qnty == 3)
+                {
+                    switch (chord_notes[1] - chord_notes[0])
+                    {
+                        case 2: //major 2nd
+                            switch (chord_notes[2] - chord_notes[1])
+                            {
+                                case 5: //diminished
+                                    chord_name += "<sup>sus2</sup>";
```
wyrdough: Strings in C# are immutable, so concatenating strings generates a bunch of garbage, which can end up causing frame drops in gameplay if the garbage collector has too much to clean up.

Ideally, you'd precalculate all the TextMeshPro objects and swap them as necessary since TMP itself is also garbage prone, but if that's not practical (keeping in mind that you can have more than one text field and populate them individually, so you don't necessarily have to store every possible permutation separately) you should at the very least use StringBuilder to add up the various components of the strings you're setting as TMP text.

wyrdough (changes requested):
Thanks for taking this on! Needs some work, but it's definitely a good start.

Assets/Script/Gameplay/Visuals/TrackElements/ProKeys/ProKeysChordBarElement.cs
WalterRalts: There probably is a way to optimize this that I'm not aware of, but I think they've all been StringBuildified.

TheNathannator: Also be informed that the `SetText` method exists for setting text *without* creating a string, and ZString provides `SetText` overloads that allow you to build text using a `Utf16ValueStringBuilder`.

```cs
using var builder = new Utf16ValueStringBuilder(true); // omit `true` or set to `false` when using multiple builders in the same scope
builder.Append("...");
_textField.SetText(builder);
```

(last i recall, anyways... I may be thinking of other similar overloads that ZString provides, such as `SetTextFormat`.)

PR #1265, opened 2026-01-15
Show percentage scores with one decimal

Ape:
I just want to see more precision, especially scores like `99.7%` instead of just `99%`.

20260115_19h57m54s_grim
20260115_19h52m09s_grim

### DIFF
--- a/Assets/Prefabs/Menu/MusicLibrary/Instrument Difficulty.prefab
+++ b/Assets/Prefabs/Menu/MusicLibrary/Instrument Difficulty.prefab
@@ -297,8 +297,8 @@ RectTransform:
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0.5, y: 0.5}
   m_AnchorMax: {x: 0.5, y: 0.5}
-  m_AnchoredPosition: {x: 16, y: 0}
-  m_SizeDelta: {x: 56, y: 28}
+  m_AnchoredPosition: {x: 19, y: 0}
+  m_SizeDelta: {x: 70, y: 28}
   m_Pivot: {x: 0.5, y: 0.5}
 --- !u!222 &7345415684953432056
 CanvasRenderer:
@@ -362,7 +362,7 @@ MonoBehaviour:
   m_fontSizeMin: 18
   m_fontSizeMax: 72
   m_fontStyle: 0
-  m_HorizontalAlignment: 2
+  m_HorizontalAlignment: 4
   m_VerticalAlignment: 1024
   m_textAlignment: 65535
   m_characterSpacing: 0
@@ -372,7 +372,7 @@ MonoBehaviour:
   m_lineSpacingMax: 0
   m_paragraphSpacing: 0
   m_charWidthMaxAdj: 0
-  m_TextWrappingMode: 1
+  m_TextWrappingMode: 0
   m_wordWrappingRatios: 0.4
   m_overflowMode: 0
   m_linkedTextComponent: {fileID: 0}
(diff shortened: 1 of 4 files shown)


4dispel:
if this is added i think it should be a setting

Dansla116:
Confirmed working as intended with no merge conflicts.  Output matches existing screenshots.

I see this mainly benefiting players who care about the precision gap between high 99% scores and 100%.  Since the UI already has space for the extra decimal and it doesn't impact readability, I'm not sure it needs to be configurable.  But I'm not opposed if others feel strongly.

wyrdough:
I would prefer not to make this a setting, at least until we have a better settings presentation for nitpicky tweaks where a setting for this could be buried, but I would like to know if there is consensus among the folks on Discord before merging this. 

(I am not personally a huge fan, but it may be more of a resistance to change than anything else, so don't take my opinion too seriously. I don't, lol)

Ape:
> I would like to know if there is consensus among the folks on Discord before merging this.

I'm not saying this is any kind of consensus, but a quick poll in `#general` shows 15 votes in favor and 6 votes against (counting "only behind a setting" as against). [1]

[1] https://discord.com/channels/1086048856678084609/1086048857714069519/1467408801895415899

polson:
How does it appear with `100%`?  Does it show `100.0%`?  `99.0%`?

Personally I like the look of a clean `100%` if I FC a track (rare) but that's me, so I'd prefer it off.

I totally understand wanting the extra decimal however.  

My thoughts:
* Given the split opinions on this I think it warrants a setting.  Maybe `General -> Other`.
* Unless we're going to get started on a settings overhaul soon, I'd be OK with merging this as a yet another setting for now
* I like settings

Dansla116:
> How does it appear with `100%`? Does it show `100.0%`? `99.0%`?
@polson

It does in fact say `100.0%`.  I could go either way on that - remove the decimal for `100%` or not.  As is, makes perfects pop out further right slightly which is nice.  Getting rid of the `.0` will make them slightly more right (but still kinda stand out).

Image one is with PR1265 as is.  Image two and three are photoshop mockups I made really quick.  Image two specifically shows `100.0%` as `100%` moving it right while keeping `99.0%` as is to stay lined up with other `99.X%`.  Image three show all `.0`'s cut off showing how `99%` gets pushed way right.
image
image
image

> I'm not saying this is any kind of consensus, but a quick poll in #general shows 15 votes in favor and 6 votes against
@Ape

> Given the split opinions on this I think it warrants a setting. Maybe General -> Other.
@polson

I understand, that in the year 2026, the people playing 5-lane rhythm games **and** voting on development decisions for 5-lane rhythm games in Discord are probably in the hard-core player-base area.  And I understand we want YARG to appeal to a more casual player-base.  With that said, I'll concede, it can be an option.  My actual only concern is scope creep.  I actually think a really good place for it is in "Songs > Music Library" right below "High Score Display".

image

Ape:
Rebased to fix tiny merge conflict.

Ape:
> How does it appear with `100%`? Does it show `100.0%`? `99.0%`?

I think `100.0%` looks a bit cleaner since it aligns the digits with all the other scores. However, I wouldn't really mind going for `100%` either.

Ape:
This needs to be properly designed on top of the new music library UI.

wyrdough:
TBH, I don't think you're going to find the extra room now, but perhaps I will be surprised...

Ape:
This is pretty important feature for me to see the more precise number. I'll see if I can figure something out.

Ape:
I pushed a new version:
- Rebased
- The score screen commit is same as before on this PR
- Reimplemented the music library commit for the new music library UI

v2

Dansla116:
Closing this because it was integrated into #1374.

PR #1268, opened 2026-01-18
Add option to save scores when bots are present

polson:
Adds an experimental setting to "Save Scores with Bots".  Only human player scores are recorded.  
 
* Bot scores are excluded from band scores and stars
 
* Saved score on music picker and saved scores in History represent the score with no bots.
 
* The score screen however, shows the score including band bonuses coming from bots.
 
* We use the replay analyzer to calculate human-only scores since it already strips bots

* Score display in the music library is now: 
```
1 Human player, 0 bots - Show solo scores
1 Human player, 1+ Bots - Show solo scores
2+ Human players, 0+ bots - Show band scores
```

**Setting:**
Screenshot 2026-01-18 at 5 26 00 PM

**Score screen with bot:**
CleanShot 2026-02-04 at 11 20 04

**Saved score on music picker:**
CleanShot 2026-02-04 at 11 20 15

**Saved score in history**
CleanShot 2026-02-04 at 11 20 22

**Things to test:**
  - Play a song with bots and setting off - scores should not save
  - Play a song with bots and setting on - scores should save
  - Verify saved band score matches human players only
  - Verify saved stars are averaged from human players only
  - Verify score in history view reflects human players only
  - Confirm replays still work correctly with bots


4dispel:
judging by the Band score just being the sum of Player scores, the calculated band score will exclude the band bonus from activating at the same time, so that playing with 2 players and a bot will not result in the same score as playing with 2 players without a bot

wyrdough:
Hm, it would probably be good to add a column to the GameRecord table that indicates whether there were bot players so that we keep our options open for excluding them should we need to do that in the future. (Plus we can add some logic to the history view to keep it from trying to find replays that we know won't exist)

polson:
> judging by the Band score just being the sum of Player scores, the calculated band score will exclude the band bonus from activating at the same time, so that playing with 2 players and a bot will not result in the same score as playing with 2 players without a bot

Yikes.  Ok maybe this is more complicated than I thought.  I'll look into it

polson:
Updated this to use replay analyzer to calculate scores after the fact, rather than keeping two scores on the fly.  Much simpler this way.

PR #1273, opened 2026-01-21
Migrate to Unity 6.3

theli-ua:
6.2 is not supported. 6.3 is an LTS version and migration is very straightforward with no code changes.o

Motivation is to use newer urp version while working on rendergraph migration. This will need to happen at some point anyway (as 6.2 is not supported version)

Builders' vars need to be updated to 6.3 as well


wyrdough (approved):
In my limited testing I didn't find anything wrong, so I'm happy with this. I would like to get some more eyes on it before merging if at all possible, though.

wyrdough:
Oh, there is one thing that needs to be added to this PR: The README needs to be updated to reflect the new Unity version...

theli-ua:
Done

README.md
wyrdough: Version also needs to be changed here

wyrdough: ..and here

wyrdough (changes requested):
Just to keep you in the loop on my plans for this, I'm going to wait on merging this until @sjdaws gets their PR whipped into shape just to make reviews/testing easier (switching back and forth between versions is annoying) and then merge this since there doesn't seem to be anything else super active at the moment. 

It looks like that should happen within the next day or two based on what I saw from @polson's review.

wyrdough (approved):
Thanks for dealing with the nitpicks, I'll merge this as soon as #1261 goes. (which should be shortly, just need to make sure it doesn't like, crash the game)

PR #1250, opened 2026-01-03
Output device and channel selection

sjdaws:
**Experimental**

* Allows output device to be selected independently of the operating system default.
* When using a device which supports multiple channels, audio can be routed independently
 
Channel routing is especially useful for hardware mixers, DAWs, and other cases where you just want main track audio + VST sounds to be output to the main/recording channel, but still want atmosphere like crowd noise or [metronome clicks](https://github.com/YARC-Official/YARG/pull/1242) to go into your IEMs.

Device selection:
Screenshot 2026-01-04 at 12 07 32 am

Channel selection based on support from device:
Screenshot 2026-01-04 at 12 07 43 am
Screenshot 2026-01-04 at 12 07 57 am

Handles are updated in real time when selection is made, previous devices are disposed of correctly to ensure memory isn't leaked.

Requires https://github.com/YARC-Official/YARG.Core/pull/314

Resolves [YARG-102](https://yarg.youtrack.cloud/issue/YARG-102/Audio-output-device-setting)

PR #1242, opened 2025-12-26
Add basic metronome

sjdaws:
Adds a basic quarter note metronome to tracks. Click track uses `Measure` and `QuarterNote` subscriptions to stay in time with track, correctly shifts with bpm, tempo, and time changes.

- 3/4 example with speed changes: https://github.com/user-attachments/assets/b2e28bf5-5fe9-4c2c-8852-13f8df9fd26a
- 4/4 example: https://github.com/user-attachments/assets/a3f99917-7135-4fd5-bbf0-d4718015572a

Sound is user selectable:
Screenshot 2026-01-07 at 12 00 05 pm

Requires https://github.com/YARC-Official/YARG.Core/pull/310

PR #1276, opened 2026-01-22
Combined metronome and output selection

sjdaws:
This PR combines #1242 and #1250 and adds an additional commit to allow output channel selection for metronome.

I've been using this as my daily driver for several weeks now without issue.

Requires https://github.com/YARC-Official/YARG.Core/pull/318


Assets/Script/Audio/Bass/BassAudioManager.cs
```diff
@@ -167,9 +157,36 @@ public BassAudioManager()
...
+
             YargLogger.LogFormatInfo("Current Device: {0}", Bass.GetDeviceInfo(Bass.CurrentDevice).Name);
         }
 
+        protected override void SetOutputDevice(string name)
+        {
+            int currentDevice = Bass.CurrentDevice;
+
+            OutputDevice? device = GetOutputDevice(name);
+            if (device == null || device is not BassOutputDevice bassDevice || bassDevice.DeviceId == currentDevice)
```
polson: I think the null check is unecessary, since `device is not BassOutputDevice bassDevice` will return `true` when device is null

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -517,8 +535,36 @@ public void OpenExecutablePath()
...
+            public OutputChannelSetting OutputChannelDrumSfx { get; } = new(-1, OutputChannelDrumSfxCallback);
+            public OutputChannelSetting OutputChannelSfx { get; } = new(-1, OutputChannelSfxCallback);
+            public OutputChannelSetting OutputChannelVox { get; } = new(-1, OutputChannelVoxCallback);
+            public OutputChannelSetting OutputChannelMetronome { get; } = new(-1, OutputChannelMetronomeCallback);
+
             public ToggleSetting EnableNormalization { get; } = new(false);
+            #endregion
+
+            #region Helpers
+            private static void resetChannelSetting(OutputChannelSetting channelSetting, SongStem stem, int defaultValue = -1)
```
polson: This should be capitalized: `ResetChannelSetting`

Assets/Script/Settings/SettingsManager.Settings.cs
```diff
@@ -517,8 +535,36 @@ public void OpenExecutablePath()
...
+                if (currentValue > GlobalAudioHandler.GetOutputChannelCount())
+                {
+                    channelSetting.Value = defaultValue;
+                    setOutputChannel(channelSetting, stem);
+                }
+
+                SettingsMenu.Instance.RefreshAndKeepPosition();
+            }
 
+            private static void setOutputChannel(OutputChannelSetting channelSetting, SongStem stem)
```
polson: Capitalize this

polson:
When we change outputs, then adjust metronome volume, I see errors getting thrown in the console:

```
[<a href="Assets/Script/Audio/Bass/BassMetronomeSampleChannel.cs" line="87">BassMetronomeSampleChannel.cs:SetVolume_Internal:87</a>] Failed to set Sine hi volume: Handle!

[<a href="Assets/Script/Audio/Bass/BassMetronomeSampleChannel.cs" line="92">BassMetronomeSampleChannel.cs:SetVolume_Internal:92</a>] Failed to set Sine lo volume: Handle!

[<a href="Assets/Script/Audio/Bass/BassMetronomeSampleChannel.cs" line="87">BassMetronomeSampleChannel.cs:SetVolume_Internal:87</a>] Failed to set Square hi volume: Handle!

[<a href="Assets/Script/Audio/Bass/BassMetronomeSampleChannel.cs" line="92">BassMetronomeSampleChannel.cs:SetVolume_Internal:92</a>] Failed to set Square lo volume: Handle!

[<a href="Assets/Script/Audio/Bass/BassMetronomeSampleChannel.cs" line="87">BassMetronomeSampleChannel.cs:SetVolume_Internal:87</a>] Failed to set Trashcan hi volume: Handle!

[<a href="Assets/Script/Audio/Bass/BassMetronomeSampleChannel.cs" line="92">BassMetronomeSampleChannel.cs:SetVolume_Internal:92</a>] Failed to set Trashcan lo volume: Handle!
```

sjdaws:
Thanks @polson, great pickup. I had wrongly assumed [disposing of the device](https://github.com/YARC-Official/YARG/pull/1276/changes#diff-219c732ecba2a1db6da1ff610dd1fb1d1d2e3b08444e61771dc75986242a5db9R178) would dispose of all loaded samples, but this isn't the case. They're now [disposed on load](https://github.com/YARC-Official/YARG/pull/1276/changes/92c3dd0eda9c9e251f037d8931c8509ecf40fe74) if they're loaded, which is invoked each time the device is switched.

Appreciate the review!

polson:
Thanks for fixing this so quickly @sjdaws 

Given that most people will not be using a metronome, I would like to propose a couple of hopefully small changes to the UX:

1. Add a default Metronome type `None` (or `Disabled`, or similar) which will not attempt to play any samples
2. Set default volume to 50% instead of 0%

This makes it more explicit about how to turn off the metronome if it's not wanted, rather than having to turn volume down to 0.  This is more consistent with other settings like `Use Crowd Sfx` for example.  

IMO It's fine that the volume slider then will have no effect as there's a lot of precedent for this already.

Lastly, feel free to join the discord and say hi in `#yarg-dev` if you would like :)

sjdaws:
Thanks again for the feedback @polson, I appreciate it. I was trying to figure out how to best do this but didn't want to add another option of disable metronome (giving us three separate metronome settings), so settled on 0 volume. Your solution is much better.

I've updated the PR with a 'None' metronome by default and adjusted default volume to 50%.

I am already on discord, just lurking! I'm going to address the [audio offset issue](https://discord.com/channels/1086048856678084609/1086052453801283624/1467586625772060910) then I think we're good to go.

sjdaws:
Audio offset issue should be resolved. 

May need to add `offset:` to the subscription, but unsure as clap example is -20ms on a 536ms sample, each metronome tick is 170ms.

sjdaws:
[Audio offset issue resolved](https://discord.com/channels/1086048856678084609/1086052453801283624/1467936520664121465)

Assets/Script/Audio/Bass/BassAudioManager.cs
```diff
@@ -242,18 +260,85 @@ public BassAudioManager()
...
+        }
+
+#nullable enable
+        protected override OutputDevice? GetOutputDevice(string name)
+#nullable disable
+        {
+            for (int deviceIndex = 0; Bass.GetDeviceInfo(deviceIndex, out var info); deviceIndex++)
+            {
+                // Ignore disabled devices
+                if (!info.IsEnabled) continue;
```
polson: Add newline braces around these ifs

Assets/Script/Audio/Bass/BassAudioManager.cs
```diff
@@ -242,18 +260,78 @@ public BassAudioManager()
...
+            return BassOutputDevice.Create(deviceId, name);
+        }
+
+        protected override List<(int id, string name)> GetAllOutputDevices()
+        {
+            var devices = new List<(int id, string name)>();
+
+            for (int deviceIndex = 1; Bass.GetDeviceInfo(deviceIndex, out var info); deviceIndex++)
+            {
+                // Ignore disabled devices
```
polson, line 310: Also newline braces around these ifs

polson:
@sjdaws looks like square and trashcan are missing in the ui:

CleanShot 2026-02-02 at 14 09 44

CleanShot 2026-02-02 at 14 09 48

Assets/Script/Audio/Bass/BassHelpers.cs
```diff
@@ -148,5 +149,28 @@ public static unsafe bool ApplyGain(float gain, IntPtr buffer, int length)
...
+            Bass.GetInfo(out BassInfo info);
+
+            return info.SpeakerCount;
+        }
+
+#nullable enable
+        public static void UpdateOutputChannels(int stream, OutputChannel? channel)
+#nullable disable
+        {
+            if (channel == null || channel is not BassOutputChannel bassChannel)
```
polson: Don't need the null check

polson (changes requested):
Looks good, just needs some small code cleanup things

wyrdough:
> Audio offset issue should be resolved.
> 
> May need to add `offset:` to the subscription, but unsure as clap example is -20ms on a 536ms sample, each metronome tick is 170ms.

I haven't looked at the sample, but in general the point of the offset is to line up the sample with the time that the player perceives it hitting, so if the sample has a couple milliseconds of silence or attack that isn't yet perceived as being the main "hit" of the sample you add the offset to adjust that.

polson:
Added in a fix to reset volumes of samples on recreation (this was an issue with all samples)

PR #1277, opened 2026-01-23
Scale highway pop in animation delay by song speed

polson:
This fixes an issue where the delay for the highway pop in was too long when playing songs on higher speeds, which means the song would start playing before we even see the strikeline.

To fix this we scale the delay by the song speed, only for when the song speed is > 100%

Video of the bug:

https://github.com/user-attachments/assets/d526aef5-1be3-47ab-a959-6fa35819acaa

Fixed:

https://github.com/user-attachments/assets/ba01b385-073a-4fbe-9f0c-e26c7b9ae152


wyrdough (approved):
Good catch, thanks for the PR!

PR #1278, opened 2026-01-23
Chord names

WalterRalts:
The font was being cut off by the shaders that I didn't account for.


wyrdough:
Hm, I think you need to pull `dev` and make a new branch from there, because whatever you did to make this PR didn't quite work right.

PR #1280, opened 2026-01-24
Remove FSR3 in favor of Unity's own TAA

theli-ua:
Since noone cares about it its not worh fixing it.


Packages/manifest.json
```diff
@@ -45,11 +45,10 @@
     "com.unity.probuilder": "6.0.7",
     "com.unity.render-pipelines.universal": "17.2.0",
     "com.unity.timeline": "1.8.9",
-    "com.unity.toolchain.linux-x86_64": "2.0.10",
+    "com.unity.toolchain.linux-x86_64": "2.0.11",
```
wyrdough, line 48: where'd this come from?

theli-ua: From Unity? 
I don't always check those in but this pr had meaningful package changes.

theli-ua: I'm not sure what exactly are you asking. If why are you not seeing that package change? I assume because you're on windows. Similar to https://github.com/YARC-Official/YARG/pull/1010

wyrdough: I was making sure it was an intentional change.

I believe Windows and Mac also had updates, so pls add those to this PR so they stay in sync.

theli-ua: Those are native tool chains I believe, as in it will only update Linux on Linux , etc. I do not have osx/win unities

Assets/Script/Gameplay/VenueCameraRenderer.cs
```diff
@@ -87,8 +87,8 @@ private void Awake()
                     _renderCamera.allowMSAA = true;
                     cameraData.antialiasing = AntialiasingMode.SubpixelMorphologicalAntiAliasing;
                     break;
-                case VenueAntiAliasingMethod.FSR3:
-                    _renderCamera.gameObject.AddComponent<FSRCameraManager>();
+                case VenueAntiAliasingMethod.TAA:
+                    cameraData.antialiasing = AntialiasingMode.TemporalAntiAliasing;
```
wyrdough, line 91: I'm going to merge this anyway since it should be fine for visualizer and other single camera venues, but it occurs to me that I am not actually sure that `_renderCamera` gets updated when switching cameras in multi-camera venues.

PR #1281, opened 2026-01-27
Fix issue where solo box was overlapping track

polson:
Discord issue: https://discord.com/channels/1086048856678084609/1459274675304534037/1459274675304534037

This PR fixes three things:

1. There was a highway camera renderer bug which was causing lanes to be taller than they should be.  This caused the solo box to overlap the highway.  The fix is to wait for `OnCameraPreRender` to reset the cameras when a new track is added

2. Solo box was overlapping the practice mode UI even with the above issue fixed.  So we moved the "restart section" to the left

3. With vocals in practice mode, the ui overlapped the vocals highway.  Fix is to push everything down enough

Fixed screenshots (with a silly tall preset):

CleanShot 2026-01-26 at 22 13 29@2x

CleanShot 2026-01-26 at 22 14 11@2x

CleanShot 2026-01-26 at 22 23 41@2x

### DIFF
--- a/Assets/Scenes/Gameplay.unity
+++ b/Assets/Scenes/Gameplay.unity
@@ -93,10 +93,8 @@ LightmapSettings:
     m_ExportTrainingData: 0
     m_TrainingDataDestination: TrainingData
     m_LightProbeSampleCountMultiplier: 4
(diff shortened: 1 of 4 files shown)


wyrdough (commented):
If you're happy with it, merge it. If it breaks something I'm sure one of us can fix it tomorrow.

PR #1283, opened 2026-01-28
Use pro drums icon when part is available

trainlos:
This will show the pro drums icon in place of the standard drums icon if pro drums is available. This does unfortunately make pro drums the only pro icon to not be in the second row, but I feel like that's preferable to not showing it at all.


wyrdough:
I don't have a strong feeling either way on this, but I do think it should be discussed on Discord so everyone has a chance to share their opinion on the change before this gets merged.

PR #1285, opened 2026-01-29
Include more data in CSV export

trainlos:
Column headers mostly follow the order of the sort options in game. Instrument difficulties needed a HasInstrument check because some songs include difficulty values for parts they don't have. 

Also removed playlist info from the plain text export to make it a simpler alternative to the csv and json options.

### DIFF
--- a/Assets/Script/Song/SongExport.cs
+++ b/Assets/Script/Song/SongExport.cs
@@ -64,7 +64,7 @@ namespace YARG.Song
         private static void ExportCsv(string path)
         {
             using var output = new StreamWriter(path);
-            
+
             output.WriteLine(
                 "Name,Artist,Album,Genre,Year,Length,Charter,Playlist,Source," +
                 "Master,Age Rating,Vocal Parts," +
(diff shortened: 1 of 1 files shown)


Assets/Script/Song/SongExport.cs
```diff
@@ -73,20 +64,87 @@ private static void ExportJson(string path)
...
-            output.WriteLine("Name,Artist,Album,Genre,Year,Length");
+
+            output.WriteLine(
+                "Name,Artist,Album,Genre,Year,Length,Charter,Playlist,Source," +
+                "Master,Age Rating,Vocal Parts," +
+                "Guitar (5-Fret),Bass (5-Fret),Rhythm (5-Fret),Co-op (5-Fret),Keys," +
+                "Guitar (6-Fret),Bass (6-Fret),Rhythm (6-Fret),Co-op (6-Fret)," +
+                "Drums (4-Lane),Pro Drums,Drums (5-Lane),Elite Drums," +
+                "Pro Guitar (17-Fret),Pro Guitar (22-Fret),Pro Bass (17-Fret),Pro Bass (22-Fret),Pro Keys," +
+                "Vocals,Harmony,Band,Format,Hash"
```
wyrdough: Do you think it would be worth changing the header to put "Difficulty" or "Diff" in the instrument column headers or something to make it more clear what those columns are for?

polson: +1

trainlos: Sure, I can add that in

PR #1286, opened 2026-01-29
Updated strikeline hit/sustain animation

LlysiX:
From my suggestion in the discord, this animation is much more responsive and more closely matches the one used in Rock Band 4. My idea was to make the animations fully customizable so you can adjust the bounce height, shakiness, bounce repeat rate, etc. however you want, but I'm not knowledgeable enough on how the preset system in YARG or how unity animations work. This is a basic example of what could be feasible with a setting like that.
Here's a video showing the difference between this and what is currently available in YARG, my change is on the bottom.

https://github.com/user-attachments/assets/76e59c94-ebfe-454f-b9c0-bd37da89c5b2

Higher quality comparison video here: https://youtu.be/PS03TlBMcz4


wyrdough:
Interesting idea, but it was decided a while back we weren't going to do this. Customizable animation parameters would be pretty cool if you're interested in working on that, though. (I doubt it given how long this PR has been left to langish, sorry about that)

PR #321, opened 2023-05-13
Small Graphics Improvements

berthrage:
+Blue Fret color changed slightly (but kept saturation);
+Fret point lights hit vfx light bleed fixed and ranges reduced as requested;
+Normal Highway track has a slight change on the lighting in its patterns;
+Starpower bar fill shadergraph improved with an improved animation, color and settings;

PR #1287, opened 2026-01-31
Trill/tremolo lane support

wyrdough:
What it says..

This is an update of #909, building on @Purplo-cf's original work and efforts to bring it up to date. Definitely not perfect yet, but I think it's good enough for play testing/soliciting feedback. I'm putting this up for comment before merging.

Requires [YARG.Core PR #321](https://github.com/YARC-Official/YARG.Core/pull/321)


theli-ua:
Do you wanna try fixing engine inconsistencies first before merging? Seems like it triggers it pretty consistently?

wyrdough:
Yeah, probably better to nip it in the bud now. Annoyingly, different scores are being generated at different fps, so maybe another rounding issue biting us in the ass?

Assets/Script/Gameplay/Visuals/TrackElements/TrackElement.cs
```diff
@@ -29,6 +29,41 @@ protected override void GameplayAwake()
             base.GameplayAwake();
         }
 
+        // private void Start()
```
polson: Uncomment or remove?

wyrdough: Oh, that does need to be removed (old fading code that came back from the dead during the merge)

wyrdough:
I made lane enablement a profile setting/engine parameter (enabled by default to get eyes on it as much as possible), so I'm going to merge this since people can just disable it if they get annoyed.

PR #1293, opened 2026-02-03
Queue instrument notes to prevent missed ticks in UDP data stream.

TheFatBastid:
Add per-instrument queues and synchronization to buffer note values so notes aren't missed between timer ticks.


Assets/Script/Integration/Data Stream/DataStreamController.cs
```diff
@@ -12,6 +13,12 @@ namespace YARG.Integration
 {
     public class DataStreamController :  MonoSingleton<DataStreamController>
     {
+        // Queues for instrument notes to prevent missed notes between timer ticks
+        private static readonly Queue<int> _drumQueue = new();
+        private static readonly Queue<int> _guitarQueue = new();
+        private static readonly Queue<int> _bassQueue = new();
+        private static readonly Queue<int> _keysQueue = new();
```
wyrdough: Just a question/comment/opening for discussion, not really an immediate request for changes: It might be good to specify an initial size for the queue so that if it doesn't have to allocate a new backing array and copy everything over during gameplay unless things end up getting super out of control. Not sure what a reasonable preallocation would be, though.

TheFatBastid: Right now the udp packets are sent 88 times a second. Arbitrary really, it is the dmx frame rate x2, also good for time travel. For visual stuff like this I like to use a blink of eye for max time (1/20 second).  If things lag more than that it is noticeable and means something is deeply wrong and needs to be solved elsewhere. So, if 88*(1/20)=4.4 so like 6 seems to be okay. So I changed it to ceiling(TARGET_FPS*(1/20))+1.

PR #1294, opened 2026-02-03
Implement auto calibration for audio

polson:
Adds an auto calibration toggle in the calibration quick settings menu

* When turned on, the auto calibrator will periodically adjust the audio calibration to line up with the user's inputs. 

* Auto calibration is turned off at the beginning of the song 

* We use the median accuracy to calculate.  The first adjustment is made after 10 notes are hit, then every 20 notes to fine tune

* Calibration is considered stable when the accuracy of the last 20 notes was +/- 10 ms

* We show a toast notification on each calibration value update

* Scores are not saved if auto calibration was enabled

* Auto calibration is only allowed when there is a single player.  We cannot calibrate multiple players at once

Video:

https://github.com/user-attachments/assets/7be6aec4-4895-49ec-a066-6e7803aa654e


Assets/StreamingAssets/lang/es-ES.json
```diff
@@ -777,7 +777,7 @@
             },
             "AudioCalibration": {
                 "Name": "Calibración de Audio (ms)",
-                "PauseName": "Calibración de Audio (ms)",
```
polson, line 780: I think these were merge conflicts, I'll undo

Assets/Script/Gameplay/HUD/Pause/QuickSettings.cs
```diff
@@ -85,10 +91,15 @@ public void OpenAudioSettings()
 
         public void OpenCalibrationSettings()
         {
-            OpenSubSettings(_calibrationSettings);
+            var settings = new List<string>(_calibrationSettings);
+            var allowAutoCalibration = PlayerContainer.Players.Count == 1 && !PlayerContainer.OnlyHasBotsActive();
+            if (!allowAutoCalibration)
+            {
+                settings.Remove(nameof(SettingsManager.Settings.AutoCalibration));
```
polson, line 99: should probably find a better way to get the string

Assets/StreamingAssets/lang/es-ES.json
polson: Actually nevermind, this is correct

Assets/Script/Helpers/AutoCalibrator.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        private void NotifyCalibrationStable()
+        {
+            ToastManager.ToastSuccess($"Auto calibration stable ({_calibration} ms)");
+        }
+
+        // Removes hits near the edges of the hit window.  We find the "middle 50%" of the data (Q1 to Q3),
+        // measure how spread out that last range is (IQR), then toss anything more than 1.5x of the range.
+        private static List<double> RemoveOutliers(List<double> values)
```
polson, line 78: Shout out claude on this one

PR #1297, opened 2026-02-04
fix venue load failing without default for video texture

4dispel:


wyrdough (approved):
YOLO, not even testing this lol

PR #1299, opened 2026-02-06
Fix compile warnings

Ape:


Assets/Script/Audio/Bass/BassAudioManager.cs
```diff
@@ -182,7 +182,7 @@ protected override bool SetOutputDevice(string name)
         {
             int currentDevice = Bass.CurrentDevice;
 
-            OutputDevice? device = GetOutputDevice(name);
```
wyrdough, line 185: I'm not looking at the code right this minute, but my recollection is that this was nullable for a reason.

Ape: This doesn't change the fact that it's nullable. It's just type inference. But it did get rid of the compiler warning.

I guess it might be more correct to add `#nullable enable` for SetOutputDevice.

PR #1300, opened 2026-02-06
Show overhit stats for each drum

Ape:
Needs https://github.com/YARC-Official/YARG.Core/pull/323

20260206_18h50m46s_grim

### DIFF
--- a/Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
+++ b/Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
@@ -1,6 +1,9 @@
 ﻿using TMPro;
 using UnityEngine;
+using YARG.Core;
 using YARG.Core.Engine.Drums;
+using YARG.Input;
+using YARG.Localization;
 
 namespace YARG.Menu.ScoreScreen
 {
@@ -9,12 +12,39 @@ namespace YARG.Menu.ScoreScreen
         [Space]
         [SerializeField]
         private TextMeshProUGUI _overhits;
+        [SerializeField]
+        private StatInfo _statInfoPrefab;
 
         public override void SetCardContents()
         {
             base.SetCardContents();
 
             _overhits.text = WrapWithColor(Stats.Overhits);
+
+            var overhitsRow = _overhits.transform.parent;
+            int siblingIndex = overhitsRow.GetSiblingIndex() + 1;
+            var bindings = BindingCollection.CreateGameplayBindings(Player.Profile.CurrentInstrument.ToNativeGameMode());
+
+            foreach (var binding in bindings)
+            {
+                if (!Stats.OverhitsByAction.TryGetValue(binding.Action, out int count))
+                {
+                    continue;
+                }
+
+                CreateOverhitRow(binding, count, siblingIndex, overhitsRow.parent);
+                siblingIndex++;
+            }
         }
+
+        private void CreateOverhitRow(ControlBinding binding, int count, int siblingIndex, Transform parent)
+        {
+            var info = Instantiate(_statInfoPrefab, parent);
+            info.transform.SetSiblingIndex(siblingIndex);
+            string key = Player.Profile.LeftyFlip ? binding.NameLefty : binding.Name;
+            info.Label.text = "<space=20px>" + Localize.Key("Bindings", key);
+            info.Value.text = WrapWithColor(count);
+        }
+
     }
 }
--- a/Assets/Prefabs/Menu/ScoreScreen/DrumsScoreCard.prefab
+++ b/Assets/Prefabs/Menu/ScoreScreen/DrumsScoreCard.prefab
@@ -1093,6 +1093,7 @@ MonoBehaviour:
   _advancedStatsRect: {fileID: 8900735449577185266}
   _basicStatsRect: {fileID: 2034896688656487534}
   _overhits: {fileID: 2974647992237571564}
+  _statInfoPrefab: {fileID: 1126423693335696835, guid: 7ceadb1ec3f67f64e8ca8ca3a866cd96, type: 3}
 --- !u!114 &7674939555971226258 stripped
 MonoBehaviour:
   m_CorrespondingSourceObject: {fileID: 2885240813424434787, guid: 1b5d8949b9f2d7345a0fed53e687c843, type: 3}
--- a/Assets/Prefabs/Menu/ScoreScreen/StatInfo.prefab
+++ b/Assets/Prefabs/Menu/ScoreScreen/StatInfo.prefab
@@ -10,6 +10,7 @@ GameObject:
   m_Component:
   - component: {fileID: 6356829577256650213}
   - component: {fileID: 5955370018740084784}
+  - component: {fileID: 1126423693335696835}
   m_Layer: 5
   m_Name: StatInfo
   m_TagString: Untagged
@@ -59,6 +60,20 @@ MonoBehaviour:
   m_FlexibleWidth: -1
   m_FlexibleHeight: -1
   m_LayoutPriority: 1
+--- !u!114 &1126423693335696835
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 4741857710018727948}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: d9b24c95829b4677b4f395b31f63cde6, type: 3}
+  m_Name:
+  m_EditorClassIdentifier:
+  Label: {fileID: 4947336598239539785}
+  Value: {fileID: 7098550245714764604}
 --- !u!1 &4782872457405020691
 GameObject:
   m_ObjectHideFlags: 0
--- a/Assets/Script/Menu/ScoreScreen/ScoreCards/StatInfo.cs
+++ b/Assets/Script/Menu/ScoreScreen/ScoreCards/StatInfo.cs
@@ -1,0 +1,11 @@
+using TMPro;
+using UnityEngine;
+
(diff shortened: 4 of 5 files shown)


polson:
Relates to https://github.com/YARC-Official/YARG/pull/1222

polson (changes requested):
Instantiate the `StatInfo.prefab` instead of cloning the existing ones

Ape:
Rebased and changed StatInfo prefab instantiation.

polson:
I think this can be simplified still.   Instead of iterating the children for the text objects, it might be clearer to make a little MonoBehavior that just exposes the text properties

Attach a StatInfo behavior to the prefab: 

```
using TMPro;
using UnityEngine;
namespace YARG.Menu
{
    public class StatInfo : MonoBehaviour
    {
        public TextMeshProUGUI Label;
        public TextMeshProUGUI Value;
    }
}

```
    
then you can reference these like
    
```
    var info = rowObject.GetComponent<StatInfo>();
if (info != null) {
    info.Label.text = "Label Text";
    info.Value.text = "Value Text";
}
```

Ape:
> I think this can be simplified still.

Do you mean like this version?

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
         public override void SetCardContents()
         {
             base.SetCardContents();
 
+            EnsureOverhitsReferences();
+            ClearSpawnedOverhitRows();
+
             _overhits.text = WrapWithColor(Stats.Overhits);
+
+            if (_overhitsInfo != null && _overhitsInfo.Label != null)
```
polson: none of this will be null

polson: I don't think we need to null check this.  It's expected to be there.

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
+            {
+                _overhitsInfo.Label.text = _baseOverhitsLabelText;
+            }
+
+            if (_statInfoPrefab == null || _overhitsRowTransform == null || _statsContainerTransform == null)
```
polson: Likewise here, all of this stuff should be there right? It's not optional, so no need to null check

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
+
+            foreach (var binding in bindings)
+            {
+                CreateOverhitRow(binding, siblingIndex + _spawnedOverhitRows.Count);
+            }
+        }
+
+        private void CreateOverhitRow(ControlBinding binding, int siblingIndex)
+        {
+            if (!Stats.OverhitsByAction.TryGetValue(binding.Action, out int count) || count <= 0)
```
polson: Can this ever actually happen?

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
+            {
+                return;
+            }
+
+            var rowObject = Instantiate(_statInfoPrefab, _statsContainerTransform);
+            rowObject.transform.SetSiblingIndex(siblingIndex);
+
+            var info = rowObject.GetComponent<StatInfo>();
+            if (info == null || info.Label == null || info.Value == null)
```
polson: I don't think we need to null check these either

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
+            string key = Player.Profile.LeftyFlip ? binding.NameLefty : binding.Name;
+            info.Label.text = "<space=20px>" + Localize.Key("Bindings", key);
+            info.Value.text = WrapWithColor(count);
+
+            _spawnedOverhitRows.Add(rowObject);
+        }
+
+        private void EnsureOverhitsReferences()
+        {
+            if (_overhitsRowTransform == null)
```
polson: no need to null check, it's there, if its not blow up

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
+            {
+                _overhitsRowTransform = _overhits.transform.parent;
+            }
+
+            if (_statsContainerTransform == null && _overhitsRowTransform != null)
```
polson: no need to null check, it's there, if its not blow up

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
+            {
+                _statsContainerTransform = _overhitsRowTransform.parent;
+            }
+
+            if (_overhitsInfo == null && _overhitsRowTransform != null)
```
polson: no need to null check, it's there, if its not blow up

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
+            {
+                _overhitsInfo = _overhitsRowTransform.GetComponent<StatInfo>();
+                if (_overhitsInfo != null && _overhitsInfo.Label != null)
```
polson: no need to null check, it's there, if its not blow up

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
...
+                {
+                    _baseOverhitsLabelText = _overhitsInfo.Label.text;
+                }
+            }
+        }
+
+        private void ClearSpawnedOverhitRows()
```
polson: I feel like this is unnecessary.  Does `SetCardContents` get called more than once?

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
         [Space]
         [SerializeField]
         private TextMeshProUGUI _overhits;
+        [SerializeField]
+        private GameObject _statInfoPrefab;
+
+        private StatInfo _overhitsInfo;
+        private string _baseOverhitsLabelText = string.Empty;
+        private Transform _overhitsRowTransform;
+        private Transform _statsContainerTransform;
```
polson: This should be serialized fields so we don't need to find it at runtime

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
         [Space]
         [SerializeField]
         private TextMeshProUGUI _overhits;
+        [SerializeField]
+        private GameObject _statInfoPrefab;
+
+        private StatInfo _overhitsInfo;
+        private string _baseOverhitsLabelText = string.Empty;
+        private Transform _overhitsRowTransform;
```
polson: This should be serialized fields so we don't need to find it at runtime

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
```diff
@@ -9,12 +13,103 @@ public class DrumsScoreCard : ScoreCard<DrumsStats>
         [Space]
         [SerializeField]
         private TextMeshProUGUI _overhits;
+        [SerializeField]
+        private GameObject _statInfoPrefab;
+
+        private StatInfo _overhitsInfo;
+        private string _baseOverhitsLabelText = string.Empty;
```
polson: I think we can get rid of this.  The label will already be assigned the default value from the prefab when its instantiated

polson (changes requested):

Assets/Script/Menu/ScoreScreen/ScoreCards/DrumsScoreCard.cs
Ape: I removed the unnecessary `count <= 0` check. The remaining check is needed.

Ape:
@polson, thanks for your review. I updated the code based on your comments and also did some further refactoring.

polson:
> @polson, thanks for your review. I updated the code based on your comments and also did some further refactoring.

Looking nice!

polson:
Not sure how, but I got a "Yellow Cymbal" overhit on 4-lane drums.  Seems like a bug perhaps?

Oh I'm wondering if it's counting overhits on 4-lane drums based on binds, so if you're playing 4-lane on a pro drum kit, you hit a cymbal, you get a cymbal overhit.  I guess that's... accurate? 

CleanShot 2026-03-10 at 11 28 41

PR #1305, opened 2026-02-07
Added Camera Cuts to UDP stream

TheFatBastid:


wyrdough (approved):
It compiles, so my work here is done. (I can take a look at the threading code later if you want me to, but it looked like you were doing what's needed to kill it, so I didn't feel the need to go any deeper)

PR #1306, opened 2026-02-07
Use base background color's alpha as highway output alpha instead of hardcoded 1.0

theli-ua:
And have highway render as transparent

### DIFF
--- a/Assets/Script/Settings/Metadata/Tabs/Tab.cs
+++ b/Assets/Script/Settings/Metadata/Tabs/Tab.cs
@@ -48,12 +48,15 @@ namespace YARG.Settings.Metadata
         {
             if (PreviewBuilder is not null)
             {
+                uiContainer.gameObject.SetActive(true);
                 var image = uiContainer.GetComponent<Image>();
                 if (image != null)
                 {
                     image.enabled = true;
                 }
                 return PreviewBuilder.BuildPreviewUI(uiContainer);
+            } else {
+                uiContainer.gameObject.SetActive(false);
             }
 
             return UniTask.CompletedTask;
--- a/Assets/Script/Settings/Metadata/Tabs/Tab.cs
+++ b/Assets/Script/Settings/Metadata/Tabs/Tab.cs
@@ -55,7 +55,9 @@ namespace YARG.Settings.Metadata
                     image.enabled = true;
                 }
                 return PreviewBuilder.BuildPreviewUI(uiContainer);
-            } else {
+            } 
+            else 
+            {
                 uiContainer.gameObject.SetActive(false);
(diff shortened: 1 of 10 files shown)


theli-ua:
I think this is the most straightforward way to allow users to have their highways be at opacity they want them to be

theli-ua:
Note that track trim will still be fully opaque, yarg does not provide any customization for track trim at the moment

wyrdough (changes requested):
Thanks for taking this on! Works better than I expected it to without sampling from the textures, actually.

One minor UX issue, though: The range of alpha values entered on the preset page does not correspond well to the effect actually seen in gameplay. I have a texture with 100% opacity and when the BG color is set to 50% opacity it is still barely visible. I think some sort of scaling of the value to better reflect the visual result needs to happen so people don't get confused.

theli-ua:
I think doing math on user specified opacity would actually complicate things. It makes it so much harder to answer the question of "so if I want highway at 50% opacity what so i set it to".

If a user sets it to 50 and it's "barely visible" to them they would go and adjust it until it is the way they like either way, scaled or not. 

But what I think would help with that workflow tremendously is changing the preview background from white to something a little more dynamic. At least a checkerboard maybe.

wyrdough:
Yes, I think using some kind of background for the track preview would render my point moot

screen cap of the default background might be relatively easy to implement?

theli-ua:
Playing with it I do think throwing smoothstep on it gives better control.

Note that in settings preview base color isn't actually used and groove base color 1 is used for opacity because fake highway in preview is always in groove mode and SP

wyrdough:
Thanks, that's much more useful now. I think we need to get some input before merging since you have the background showing up on all the settings pages, but I'm personally fine with this as is.

Assets/Script/Settings/Metadata/Tabs/Tab.cs
```diff
@@ -48,12 +48,15 @@ public virtual UniTask BuildPreviewUI(Transform uiContainer)
...
+                uiContainer.gameObject.SetActive(true);
                 var image = uiContainer.GetComponent<Image>();
                 if (image != null)
                 {
                     image.enabled = true;
                 }
                 return PreviewBuilder.BuildPreviewUI(uiContainer);
+            } else {
+                uiContainer.gameObject.SetActive(false);
             }
```
wyrdough: I hate having to be like this, but house style dictates the `else` go on a line by itself.

```suggestion
            } 
            else 
            {
                uiContainer.gameObject.SetActive(false);
            }

wyrdough (changes requested):

PR #1123, opened 2025-09-10
Updated Music Library UI

poliwhirlge:
This PR implements the music library as designed by Kadu on the Figma under Interfaces.

Changes/new features:
- Re-styled music library menu and sidebar
- New clickable play and favorite button in sidebar, back button on the header
- Removed green and red buttons from HelpBar by adding a hide option in NavigationScheme.Entry
- New header displaying the current sorting method, # of songs, and # of stars earned
- Re-styled HelpBarButton
- Changed displayed instrument rings due to addition of band difficulty bar (split out the vocals difficulty into vocals and harmony)
- Song content rating is displayed on sidebar (placeholder text, will be replaced with icons once I get high-res versions)
- Added Select as a keybind to cycle through sorting methods
- Changed behavior of clicking/right-clicking the instrument difficulty rings, now filters regardless of activeness
- Should scale a bit better under non 16:9 layouts compared to the current version

Some screenshots:
image
image

Future things that need to be taken care of:
- Update icons for song content ratings
- Re-organize the NavigationSchemes for all the different kinds of menus we have now
- Not a high priority, but I'd like to eventually place the SongViews from the top of the container downwards instead of having them centered (and thus removing the space that's at both ends), then we can have subheaders that stick to the top of the screen.

I tried my best to merge everything correctly and everything seems to be covered so far but I'll be here to help fix issues that crop up.

PR #1311, opened 2026-02-09
Updated Music Library UI, Conflict Resolution Helper

Dansla116:
Resolves merge conflicts from #1123 (Aug–Sep 2025).
- Conflicts resolved against `dev` as of Feb 8, 2026
- No functional or UI changes beyond conflict resolution
- Unity scene/prefab conflicts handled intentionally

Tested against:
- base of #1123
- current `dev`
- merged result

No negative impacts or unintended revisions observed.


polson:
For convenience, the previous PR description in lazy format

CleanShot 2026-02-09 at 11 45 31

Dansla116:
Addressed an issue mentioned in Discord where albums with long name would overflow in the sidebar.  If an album name exceeds 50 characters, the text now wraps at the last space before the 50th character and scales down to 80% of its original size.  Both the character limit (50) and scale factor (80%) are defined as constants so they can be easily adjusted if needed.

Before:
image

After:
image

wyrdough:
I am inclined to go ahead and merge this soon and just live with the stylistic difference with everything else for now. Otherwise it'll just succumb to bitrot again and I think that we can do *enough* restyling on the other menus to not make it jarringly different even if it's not fully realizing the original vision for the completed new UI

wyrdough:
Looks like something is missing from the addressables database (specifically "MusicLibraryUpIcon"). The other thing that strikes me at first glance is that it would probably be better to have the navigation bar text a little bit larger and perhaps (temporarily) a bit brighter so it meshes better with the menus that haven't yet been updated.

PR #1312, opened 2026-02-10
Fix stuck held inputs when resuming from rewind

fabricio-lp:
## Bug
Pausing during a sustain/long note and resuming (with rewind/seek) could leave inputs stuck in a held state. After the sustain ends, no further notes can be played until restarting the song.

## Root cause
`GameManager.ResumeCore()` called `player.SendInputsOnResume()` while `Rewinding == true`, so release events were blocked and never reached the engine.

## Fix
Set `Rewinding = false` before calling `SendInputsOnResume()` so queued resume inputs (including releases) are processed correctly.

## Manual test
- Start a song with sustains
- Hold a sustain, pause while holding
- Resume (rewind occurs)
- Release when the sustain ends and continue playing notes normally
- Verified with both all taps and strum

### DIFF
--- a/Assets/Script/Gameplay/GameManager.cs
+++ b/Assets/Script/Gameplay/GameManager.cs
@@ -497,12 +497,13 @@ namespace YARG.Gameplay
 
             _isReplaySaved = false;
 
+            Rewinding = false;
+
             foreach (var player in _players)
             {
                 player.SendInputsOnResume();
             }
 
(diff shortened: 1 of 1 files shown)


wyrdough:
Thanks for this. TBH, I probably just need to start again on the whole thing (I ended up overcomplicating the whole thing), but this will serve as a good interim fix!

PR #1314, opened 2026-02-11
Add Filters Menu to Music Library

Dansla116:
Music Library Changes:
- Changed blue from "Play a Show" to "Filters"
- Changed yellow from "Add to Setlist" to "Play a Show"
- Yellow is still "Start The Set" when setlist has content
- Added complex punnett square green behavior
- - If setlist is empty: tap green to "Play Song", hold green to "Add to Setlist"
- - If setlist has content: tap green to "Add to Setlist", hold green to "Start the Set"
- Changed song count to show "Filter Result" / "Total Songs" when filters are active
- Remove screen elements and pause preview when navigating to Filters Menu:

Music library. no filters, setlist is empty:
image

Music library, no filters, setlist has content:
image

Music library, with filters (only Source is "Rock Band Network 1.0" or "Rock Band Network 2.0" and Difficulty is "Solid" (2):
image

Filters Menu Features:
- Added screen copying many components/layout from Settings Menu
- Added "Sort By" option that is also still available from Music Library "More Options"
- Added "Show Recommendation" toggle
- Added "Reset All Filters" button
- Added filters for Genres, Decades, Vocal Parts, Sources, Charters, Difficulties, and Lengths
- Added "Select All" and "Deselect All" buttons to each category
- Left comments for different Length buckets and toggle
- Left comments for Vocal Gender and Has Solos filters that are harder to implement

Filters menu, initial landing:
image

Filters menu, "Genres" category has focus:
image

Filters menu, inside "Song Sources", deselected all, selected two options, "Rock Band Network 2.0" has focus:
image


Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -753,12 +868,51 @@ private bool IsButtonHeldByPlayer(YargPlayer player, MenuAction button)
             return _heldInputs.Any(i => i.Context.Player == player && i.Context.Action == button);
         }
 
-        private void OnButtonHit(NavigationContext ctx)
+        private const float GREEN_HOLD_SECONDS = 1f;
+        private void OnGreenHit(NavigationContext ctx)
```
wyrdough: Don't take me too seriously because I'm still dead-ass tired, but is this really the best place for this? Wouldn't it be better to fix `StartNavigationHold` and `EndNavigationHold` in Navigator so they can do the long press action correctly on their own? (perhaps optionally, but I'm not sure that it's worth the trouble since I can't think of anywhere we actually want the action to fire on release)

wyrdough (commented):
Looks pretty reasonable, though I haven't done an in depth review. For now just a couple of questions about implementation choices. One in another comment and another here since it doesn't apply to a specific bit of code: Is it not possible to reuse existing settings prefabs for the most part? It makes things a lot easier if we ever need to reskin in the future.

Nice work, by the way. Functionally I'm super happy with this, from what I've seen so far it does the business quite nicely.

Dansla116:
Address initial feedback - asset maintainability and green tap/hold logic
No UI/UX changes here.  As said in commit message:
- Removed duplicate settings button in FiltersMenu
- Tried making FiltersMenu and FilterEntryRow variants from SettingsMenu but couldn't get them working right.
- Moved logic for tap/hold green from MusicLibraryMenu to Navigator and NavigationScheme

FilterEntryRow especially, I could not get to work right, even though it seemed possible.  The problem is, it is a toggle row that goes [ label | toggle | label ] and they are otherwise, just [label | toggle ].  No matter what I tried, things would not line up right.  Maybe it's something in Unity editor I just couldn't figure out.

Assets/Script/Menu/Filters/FiltersMenu.cs
```diff
@@ -0,0 +1,1384 @@
...
+            // Match SettingsMenu behavior: initialized at startup, then hidden.
+            gameObject.SetActive(false);
+            _ready = true;
+        }
+
+        private void OnEnable()
+        {
+            if (!_ready) return;
+
+            _leftNavGroup.SelectionChanged += OnSelectionChanged;
```
polson, line 164: _leftNavGroup is treated as required but _rightNavGroup isn't

My opinion is just assume we have a _rightNavGroup and don't bother null checking

Assets/Script/Menu/Filters/FiltersMenu.cs
```diff
@@ -0,0 +1,1384 @@
...
+            AddHeader(container, Localize.Key("Menu.Filters.FiltersHeader"));
+            AddButton(container, navGroup, Localize.Key("Menu.Filters.ResetAllFilters"), ResetAllFilters);
+            AddGroup(container, navGroup, FilterGroup.Genre,  Localize.Key("Menu.Filters.Genres"));
+            AddGroup(container, navGroup, FilterGroup.Decade, Localize.Key("Menu.Filters.Decades"));
+            AddGroup(container, navGroup, FilterGroup.VocalParts, Localize.Key("Menu.Filters.VocalParts.Name"));
+            AddGroup(container, navGroup, FilterGroup.Source, Localize.Key("Menu.Filters.Sources"));
+            AddGroup(container, navGroup, FilterGroup.Charter, Localize.Key("Menu.Filters.Charters"));
+            AddGroup(container, navGroup, FilterGroup.Difficulty, Localize.Key("Menu.Filters.Difficulties.Name"));
+            AddGroup(container, navGroup, FilterGroup.Length, Localize.Key("Menu.Filters.Length.Name"));
+            // AddGroup(container, navGroup, FilterGroup.VocalGender, "Vocalist Gender");
```
polson: Remove or uncomment

Assets/Script/Menu/Filters/FilterCategoryRow.cs
```diff
@@ -0,0 +1,46 @@
...
+            if (_secondaryLabel == null) return;
+
+            _secondaryLabel.text = text ?? string.Empty;
+            _secondaryLabel.gameObject.SetActive(!string.IsNullOrEmpty(text));
+        }
+    }
+
+    public enum FilterGroup
+    {
+        SortedBy,
```
polson: Unused

Assets/Script/Menu/Filters/FilterCategoryRow.cs
```diff
@@ -0,0 +1,46 @@
...
+        ShowRecommended,
+
+        Genre,
+        Decade,
+        VocalParts,
+        Source,
+        Charter,
+        Difficulty,
+        Length
+        // VocalGender,
```
polson: Remove or uncomment

Assets/Script/Menu/Filters/FiltersMenu.cs
```diff
@@ -0,0 +1,1384 @@
...
+            var counts = def.GetCounts();
+            BuildOptions(
+                values,
+                def.Enabled,
+                counts,
+                () => UpdateSummary(def.Group, def.Enabled, def.GetValues()),
+                def.LabelTransform);
+        }
+
+        private static bool TryGetSelectedSet(
```
polson, line 641: Am I correct in thinking that the list of keys for `enabled` are identical to `all` here?  

Perhaps I'm missing something, but otherwise if that's the case we could simplify things here I think

Assets/Script/Menu/Filters/FiltersMenu.cs
```diff
@@ -0,0 +1,1384 @@
...
+            if (TryGetSelectedSet(_difficultyEnabled, GetAllDifficultiesCached(difficultyInstrument), NormalizeFilterKey, out var difficulties))
+                predicates.Add(entry =>
+                {
+                    var label = GetDifficultyLabel(entry, difficultyInstrument);
+                    return label != null && difficulties.Contains(NormalizeFilterKey(label));
+                });
+
+            if (predicates.Count == 0) return null;
+
+            return entry =>
```
polson: I think you could just do like `return entry => predicates.All(predicate => predicate(entry));`

wyrdough:
If they need to be separate prefabs, that's fine, I just want to be sure we actually need them to be different since it will add overhead when we eventually get around to fully reskinning the rest of the UI to better match what we're doing with the music library.

For visual consistency and making (at least) my eyes glaze over a bit less with the long lists it would be nice if this adopted the "odd/even" background logic from the settings menu. Otherwise I'm fine with this as an initial implementation from the UI standpoint aside from considering whether it might be better to make the reset filters option a navigation button thing. (my opinion on that is not strong, just a possible consideration)

Dansla116:
I'll be honest, I didn't even realize we alternated colors.  But yeah, the proof is there.  I got that added to both filter categories (left) and filter entries (right).  I also realized I didn't have dividers between filter entries.  So those should look more separate now.
image

Also, per the feedback on Discord, I've removed the "Reset All Filters" button and added "(Hold) Reset All Filters" to the Navigator.  It uses existing hold logic so will trigger after one second held.

Here is the new implementation:
image

wyrdough:
Thanks for the changes, that's much more readable to me now!

Sorry to keep throwing stuff at you, but doing some more testing, I found a couple of other things that need to be addressed:

* FilterMenu should be in the MusicLibraryMenu prefab, not in PersistentScene. (Eventually, I plan to remove all the settings from PersistentScene, so I'd prefer not to add anything new, plus the more stuff we have stuffed directly in scenes the more likely we are to end up with merge conflicts going forward). Since FilterMenu is a singleton that disables itself on awake, it seems like you should just be able to add the prefab below everything else in the MusicLibraryMenu prefab and it will cover the library menu just as it does now.
* I think it would be better if recommended was implicitly disabled when filters are set. It's a bit confusing to set a source filter, for example, and end up having the first thing selected be something that isn't in the filter.
* It is presently not possible to start a setlist with the mouse or with ekits that don't have any inputs that hold. The easy (and probably best, tbh) option is to make sure whatever hold actions we have available are also in the context menu. This doesn't need to be some automatic thing, just tell context menu to enable/disable the menu item where ever you're changing the navigation context or have it check a flag on the music library or whatever. (I forget how we're doing it for everything else rn, just do that lol)
  * Longer term, we should probably make it possible for mouse clicks to have a hold action, but that is not something you need to deal with, I'm just noting it here for posterity.

polson:
> I think it would be better if recommended was implicitly disabled when filters are set. It's a bit confusing to set a source filter, for example, and end up having the first thing selected be something that isn't in the filter.

What about having recommended respect the selected filters? 

I would say out of scope for this PR, just implicitly disable like you suggested, but I could see that being useful

Dansla116:
Addressed more feedback:
- Moved `FiltersMenu` from `PersistentScene.unity` to `MuiscLibraryMenu.prefab`. *(improve code maintainability, not a UI/UX change)*
- **Explicitly** turn "Show Recommendations" off automatically if you apply any filters.  In testing, I found **implicitly** turning this off caused confusion.  You can turn it back on, but recommendations still won't care about your filters.
- Added green hold actions to `PopupMenu` - either "Add to Setlist" or "Start the Set", never both.  Gives alternate way to trigger action when hold isn't available (e.g. edrums).
- "Play a Show" will only pull options that match selected filters. @agentevil

polson (changes requested):

Dansla116:
Addressed code review feedback and cleaned up other bits I noticed.  As mentioned in commit message:
- Removed comments for non-implemented filters by Vocal Gender and Has Solo
- Removed null check on right menu
- Removed unnecessary `using` and fully qualified types
- Removed unused `AddButton()`
- Ensure defaults before building selected set.  Within the current menu flow `enabled` and `all` should stay in sync, but I kept `EnsureDefaults()` here to make that invariant explicit and safe if that ever changes (i.e. rescan from library).
- Use LINQ `All()` for filter predicate

polson (approved):
Looks good

PR #1315, opened 2026-02-11
Fix open sustains rendering

theli-ua:
By setting track shader alpha mode to additive


wyrdough (approved):
Works a treat, thanks! (And yes, transparency still works)

PR #1316, opened 2026-02-12
Fix dpi spam in player.log

polson:
Well, hopefully


wyrdough:
Perhaps @theli-ua can test this at some point...

theli-ua:
> Perhaps @theli-ua can test this at some point...

I don't think I ever had it

polson:
This runs every frame so I'm pretty sure it was the issue.  Only question is if the new call still asks for DPI but I would guess not

PR #1317, opened 2026-02-12
Rewrite lyrics bar to support transition animations

Supersonicboss1:
Unfortunately, due to the way it was previously written (two text objects which just get set at the start of each phrase) there was no way to cleanly do transitions without a complete rewrite.

A high level overview is that 3 lyric text objects are created from a prefab, which then each get assigned 1/3rd of the lyric phrases in a round robin fashion. All lyric phrase transition timings get precalculated in `OnChartLoaded`, and then just use this data rather than doing any calculations on the fly.

All transition animations are handled by directly setting the values every frame (where applicable), using `DOVirtual.EasedValue` for easing to avoid allocations from using `DOTween` directly.

Fading the lyric text in and out is kept from the previous version, but because of the different design, now uses a `CanvasGroup` for fading in and out all of the lyric objects at once.

This doesn't include the ability to set the scale of the lyric bar, I figured I would look at that later so this doesn't get too big.

This is the largest PR I have attempted for YARG, so apologies in advance if this code sucks!

Demo 1 (normal song):

https://github.com/user-attachments/assets/ca540d1a-634e-4254-a824-6c5f0aad67d7

Demo 2 (chart with the gimmicky lyric stuff, same one I posted on Discord):

https://github.com/user-attachments/assets/a8fca9d0-68cd-4dd8-8594-46c52879c847


polson (approved):
Looks good to me

wyrdough:
Just want to make sure @Frickitickitavi doesn't have any thoughts since he has worked on lyric-related things relatively recently.

Frickitickitavi:
> Just want to make sure @Frickitickitavi doesn't have any thoughts since he has worked on lyric-related things relatively recently.

No flags from me - the fact that this works with Mindgerb is all I need to see lol

PR #1328, opened 2026-02-16
Fix discord sdk console error on macos

polson:
We have both a .bundle and a .dylib which conflict for the discord sdk

Both discord_game_sdk.bundle and discord_game_sdk.dylib had the same internal plugin name, and Unity can only load one in the Editor.

Fix is to deactivate the dylib and keep the bundle


theli-ua:
Nice. I was wondering what's up with that couple times I had to debug osx problems

PR #1329, opened 2026-02-17
Add "dedicated open lane" setting for Keys, generalize highway ordering logic

Frickitickitavi:
[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/329)

The main user-facing change in this PR is a new `Dedicated Open Note Lane` profile setting for Keys profiles which controls how open notes are displayed (presumably when playing a guitar-native chart, although opens can unconventionally be placed in PART KEYS as well). It allows the player to opt into displaying opens in a sixth lane to the left of green, which is much more readable for players who bind open to B♮ on their keyboard controller. The setting defaults to `Never`, which preserves preexisting behavior, and also has `Always` and `If chart has open notes` settings.

Behind the scenes, this also involves a huge overhaul of how `Player` classes and `FretArray` interact. Previously, the logic for placing frets, gems, lane markers, and range shift markers was heavily duplicated across the codebase and full of edge cases to account for lefty flip and the various drum highway swap/split options. These changes establish each `Player` class as the source of truth for all of that information; each player class instantiates a small `Dictionary` that maps each note type to its position (and, in the case of drums, color, since lefty flip on drums affects position and color independently).

Drumbs had suggested using a `List` instead of a `Dictionary` for performance reasons - I'm certain he's right and that this could be optimized, but I found it much easier to wrap my head around a `Dictionary` of explicit note-to-position bindings. Feel free to suggest any places where what I've done can be made more efficient. I'm particularly conscious of the changes I made to the range shift fret dimming logic, which went from a lean `bool[]` to a `List<int>` which gets cleared and refilled; I'm sure it can be cleaned up.

Along the way, I caught and fixed various minor bugs and bits of technical debt, such as:
* Range shift indicators did not properly handle lefty flip
* Split-lane Pro Drums bots would "press" the Gcym fret whenever they hit a Gtom gem
* While Keys profiles correctly hide the lefty flip option, it was being incorrectly applied to 5L Keys charts if set via another profile type
* All `shiftLeft` terminology in the range shift code was phrased backwards, and really meant `shiftRight` (not user visible)


wyrdough:
Aside from a couple of places where you you misplaced some braces nothing jumped out at me as being horribly wrong here. I probably will have some thoughts when I have the chance to do a full review (should be able to get to it in a few days when I have replaceable characters done), but overall this seems to have turned out better than I thought it might.

Assets/Script/Gameplay/Visuals/Fret/FretArray.cs
```diff
@@ -234,46 +212,46 @@ public void UpdateAccentColorState(int fretIndex, bool shouldWhiten)
         public void SetFretColorPulse(int fretIndex, bool pulse, float duration)
         {
             _pulseDuration = duration;
-            _pulsingFrets[fretIndex] = pulse;
+
+            if (pulse)
+            {
+                _pulsingFrets.Add(fretIndex);
+            } else
```
wyrdough: Insert whiny "else goes on its own line" comment here

Assets/Script/Gameplay/Visuals/Fret/FretArray.cs
```diff
@@ -93,59 +118,12 @@ public void Initialize(ThemePreset themePreset, VisualStyle style,
                 _kickFrets.Add(rightKick.GetComponent<KickFret>());
             }
 
-            InitializeColor(fretColorProvider, leftyFlip, splitProTomsAndCymbals);
-
-            _activeFrets = new bool[FretCount];
-            _pulsingFrets = new bool[FretCount];
+            _activeFrets = new();
```
wyrdough: Unless we're reinitializing during the lifetime of the FretArray, you can do the initialization and declaration in the same place.

The only real advantage to doing it here is if you check how many frets exist and tell `List` to preallocate the correct size of backing array to begin with. (Like if there's 5 frets, you can tell List that in the constructor and it will allocate a 5 element backing array..doesn't really matter for such a small list, but it can avoid new allocations if the list has to grow larger than the default size)

Assets/Script/Gameplay/Visuals/TrackElements/FiveLaneKeys/FiveLaneKeysRangeIndicatorElement.cs
```diff
@@ -23,11 +22,29 @@ public class FiveLaneKeysRangeIndicatorElement : TrackElement<FiveLaneKeysPlayer
...
+            var rangeIncludesOpen = (Player.UsingOpenLane && RangeShift.Position is (int) FiveFretGuitarFret.Green);
+            var rangeShiftSize = RangeShift.Size + (rangeIncludesOpen ? 1 : 0);
+
+            var newXScale = (rangeShiftSize / scaleDenominator) * 2;
+
+            int positionOffset;
+            if (Player.UsingOpenLane)
+            {
+                positionOffset = rangeIncludesOpen ? -1 : 0;
+            } else
```
wyrdough: else goes on its own line

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -729,19 +670,16 @@ private Fret.AnimType GetAnimType(int fret)
...
-            // Refers to the lane where 0 is red
-            int fret = GetFret(action);
-
             if (_fiveLaneMode)
             {
                 // Only use cymbal animation if the cymbal gems are being used
                 if (Player.Profile.UseCymbalModels && action is DrumsAction.YellowCymbal or DrumsAction.OrangeCymbal)
                 {
-                    _fretArray.PlayCymbalHitAnimation(fret);
+                    _fretArray.PlayCymbalHitAnimation(DrumsActionToHighwayIndex(action));
```
wyrdough: I think this would be more readable if you did something like `var index = DrumsActionToHighwayIndex(action)` up above and used index as the method call parameter for all of these.

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -795,137 +730,105 @@ is FourLaneDrumPad.YellowCymbal
...
+                LaneCount = 5;
+                _highwayOrdering = new()
+                {
+                    { (int)FiveLaneDrumPad.Red,    new(ApplyHandednessToPosition(Player.Profile.SwapSnareAndHiHat ? 1 : 0), ApplyHandednessToFiveLaneColor(FiveLaneDrumsFret.Red) ) },
+                    { (int)FiveLaneDrumPad.Yellow, new(ApplyHandednessToPosition(Player.Profile.SwapSnareAndHiHat ? 0 : 1), ApplyHandednessToFiveLaneColor(FiveLaneDrumsFret.Yellow) ) },
+                    { (int)FiveLaneDrumPad.Blue,   new(ApplyHandednessToPosition(2),                                        (int)FiveLaneDrumsFret.Blue) }, // No need to waste a function call on this
+                    { (int)FiveLaneDrumPad.Orange, new(ApplyHandednessToPosition(3),                                        ApplyHandednessToFiveLaneColor(FiveLaneDrumsFret.Orange) ) },
+                    { (int)FiveLaneDrumPad.Green,  new(ApplyHandednessToPosition(4),                                        ApplyHandednessToFiveLaneColor(FiveLaneDrumsFret.Green) ) }
+                };
+            }
```
wyrdough: lol, I guess this makes up for one of the others that was missing a line break

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -795,137 +730,105 @@ is FourLaneDrumPad.YellowCymbal
...
+                    { (int)FourLaneDrumPad.YellowDrum,    new(ApplyHandednessToPosition(2),                                          ApplyHandednessToFourLaneColor(FourLaneDrumsFret.YellowDrum)) },
+                    { (int)FourLaneDrumPad.BlueCymbal,    new(ApplyHandednessToPosition(Player.Profile.SwapCrashAndRide ? 5 : 3),    ApplyHandednessToFourLaneColor(FourLaneDrumsFret.BlueCymbal)) },
+                    { (int)FourLaneDrumPad.BlueDrum,      new(ApplyHandednessToPosition(4),                                          ApplyHandednessToFourLaneColor(FourLaneDrumsFret.BlueDrum)) },
+                    { (int)FourLaneDrumPad.GreenCymbal,   new(ApplyHandednessToPosition(Player.Profile.SwapCrashAndRide ? 3 : 5),    ApplyHandednessToFourLaneColor(FourLaneDrumsFret.GreenCymbal)) },
+                    { (int)FourLaneDrumPad.GreenDrum,     new(ApplyHandednessToPosition(6),                                          ApplyHandednessToFourLaneColor(FourLaneDrumsFret.GreenDrum)) },
+                };
+            }
```
wyrdough: Well damn, on average it was right...

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -16,13 +17,84 @@
 using YARG.Player;
 using YARG.Settings;
 using YARG.Themes;
+using static YARG.Core.Game.ColorProfile;
 
 namespace YARG.Gameplay.Player
 {
     public class DrumsPlayer : TrackPlayer<DrumsEngine, DrumNote>
     {
+
```
wyrdough: Don't think we need the extra space here. (not that a couple of bytes costs anything)

And just as a note, I didn't check the lane ordering switches for correctness in terms of what goes where. I assume you tested that sufficiently.

Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -1,13 +1,14 @@
-﻿using System;
+﻿using NUnit.Framework;
```
wyrdough: ???

Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
```diff
@@ -27,6 +29,49 @@ public sealed class FiveFretGuitarPlayer : TrackPlayer<GuitarEngine, GuitarNote>
...
+        {
+            if (_lanePositions.ContainsKey(fret))
+            {
+                return _lanePositions[fret];
+            }
+
+            return (LANE_COUNT - 1) / 2;
+        }
+
+        private FiveFretGuitarFret _getFretIndex(GuitarAction action)
```
wyrdough: Why is this method prefixed with an underscore? (We don't typically do that for private methods, just fields)

Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
```diff
@@ -148,22 +193,25 @@ protected override void FinishInitialization()
...
-                false  // Not applicable to five fret
-                );
+                Player.ThemePreset,
+                VisualStyle.FiveFretGuitar
+            );
 
             if (Player.Profile.RangeEnabled)
             {
-                _activeFrets = new bool[_fretArray.FretCount];
+                _activeFrets = new();
```
wyrdough: I know we're not recreating a whole player, so instantiate with the declaration.

Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
```diff
@@ -679,8 +712,8 @@ private void InitializeRangeShift(double time = 0)
                     _shiftIndicators.Enqueue(new RangeShiftIndicator
                     {
                         Time = beatlines[realIndex].Time,
-                        LeftSide = shiftLeft,
-                        Offset = shiftLeft ? ((shift.Position + shift.Size) - 6) * -1 : shift.Position - 1,
+                        RightSide = shiftRight,
+                        Offset = shift.Position < lastShiftRange.Position ? shift.Position - 1 : 5 - shift.Position - shift.Size + 1,
```
wyrdough, line 715: I have zero clue how I came up with the offset calculation in the first place, so I'm going to assume this is correct lol

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
```diff
@@ -142,19 +189,21 @@ protected override void FinishInitialization()
...
-                false  // Not applicable to five fret
-                );
+                Player.ThemePreset,
+                VisualStyle.FiveFretGuitar
+            );
 
             if (Player.Profile.RangeEnabled)
             {
-                _activeFrets = new bool[_fretArray.FretCount];
+                _activeFrets = new();
```
wyrdough: Initialization with declaration, please (or make use of the fact that you know whether you need 5 or 6 list elements by the time we get here)

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
```diff
@@ -380,12 +424,12 @@ protected override void OnNoteHit(int index, GuitarNote note)
 
             (NotePool.GetByKey(note) as FiveLaneKeysNoteElement)?.HitNote();
 
-            if (note.FiveLaneKeysAction is FiveLaneKeysAction.OpenNote)
+            if (note.FiveLaneKeysAction is FiveLaneKeysAction.OpenNote && !UsingOpenLane)
             {
                 _fretArray.PlayOpenHitAnimation();
             } else
```
wyrdough: ...

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
```diff
@@ -28,7 +29,53 @@ public sealed class FiveLaneKeysPlayer : TrackPlayer<FiveLaneKeysEngine, GuitarN
...
+                return _lanePositions[fret];
+            }
+
+            return (LaneCount - 1) / 2;
+        }
+
+        public int LaneCount { get; private set; }
+        public bool UsingOpenLane { get; private set; }
+
+        private FiveFretGuitarFret _getFretIndex(FiveLaneKeysAction action)
```
wyrdough: Same question as in guitar..why an underscore?

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
```diff
@@ -586,11 +626,26 @@ private void InitializeRangeShift(double time = 0)
...
+                    int offset;
+                    if (shiftRight)
+                    {
+                        offset = LaneCount - shift.Position - shift.Size + (UsingOpenLane ? 0 : 1);
+                    } else
+                    {
+                        if (UsingOpenLane && shift.Position is (int)FiveFretGuitarFret.Green)
+                        {
+                            offset = 0; // When shifting down to GRY[B] in open lane mode, treat P as part of the range
+                        } else {
```
wyrdough: ...

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
```diff
@@ -622,13 +687,57 @@ private void SetActiveFretsForShiftEvent(FiveFretRangeShift range)
...
+                        {
+                            if (child.Fret is (int)FiveFretGuitarFret.Open)
+                            {
+                                return true;
+                            }
+                        }
+                    }
+                    return false;
+                default:
+                    throw new Exception("Unreachable");
```
wyrdough: I don't really care if you change this since it's a personal preference, but I like to use more specific exception types when it makes sense. Like an `ArgumentOutOfRangeException` here...

wyrdough (changes requested):
Some minor changes and questions, but overall a very good showing. Thanks!

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
polson: curiously Rider doesn't complain about this for me.  Idk if it's a common C# convention for private methods

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
polson: I think you mean to say bracket on newline

polson: btw not to start a discussion here but if we had a lint runner on PRs that literally checked only for newline brackets I think that would be cool

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
```diff
@@ -586,11 +626,26 @@ private void InitializeRangeShift(double time = 0)
 
                     firstBeatTime = beatlines[realIndex].Time < firstBeatTime ? beatlines[realIndex].Time : firstBeatTime;
 
+                    int offset;
+                    if (shiftRight)
+                    {
+                        offset = LaneCount - shift.Position - shift.Size + (UsingOpenLane ? 0 : 1);
```
polson: You can DRY this up since its used twice: (UsingOpenLane ? 0 : 1)

polson (approved):
Straight up no idea what most of this is doing but code looks fine to me

Assets/Script/Gameplay/Player/DrumsPlayer.cs
Frickitickitavi: Not sure what you're referring to - the blank line between the `}` and the `else if`? Should I remove that?

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
Frickitickitavi: I don't see the guitar comment you're referring to, but either way - I always get mixed up with the function name styling standards. I was thinking this was the standard for a private method - what should it actually be? Do we have all those standards documented somewhere?

Frickitickitavi: Oh, the guitar comment was hidden. Will fix

Assets/Script/Gameplay/Player/DrumsPlayer.cs
Frickitickitavi: Beats me. Will remove

Assets/Script/Gameplay/Player/DrumsPlayer.cs
wyrdough: Yeah, sorry, thought I was being funny

wyrdough (approved):
Approved pending testing revealing something horrible

Frickitickitavi:
FYI @wyrdough - dawned on me that I completely forgot to include activation highlights in this PR. Just did an implementation, and it seems to be working from a quick smoke test

polson:
functionality looks good to me

PR #1334, opened 2026-02-18
Allow adjusting the start and end practice positions using the arrow keys and mouse scroll wheel

neilweber:
The start and end practice positions can now be adjusted 1 second by navigating to the appropriate menu item and pressing buttons bound to Menu.Decrease and Menu.Increase.  Menu.Decrease and Menu.Increase were previously ignored in this context.

Likewise, the positions can now be adjusted 5 seconds by navigating to the appropriate menu item and scrolling forwards or backwards with the mouse scroll wheel.  The code in PracticeSectionMenu.cs was used as a model for these changes.

This code change also fixes a minor bug.  Previously setting the start or end position to the current strike line did not reset the HUD statistics.  Now it does.

### DIFF
--- a/Assets/Script/Gameplay/HUD/Pause/PracticePause.cs
+++ b/Assets/Script/Gameplay/HUD/Pause/PracticePause.cs
@@ -1,6 +1,8 @@
 using System;
 using TMPro;
 using UnityEngine;
+using YARG.Core.Input;
+using YARG.Menu.Navigation;
 
 namespace YARG.Gameplay.HUD
 {
@@ -12,11 +14,46 @@ namespace YARG.Gameplay.HUD
         [SerializeField]
         private TextMeshProUGUI _bPositionText;
 
+        private NavigatableBehaviour _aPositionNav;
+        private NavigatableBehaviour _bPositionNav;
+
         protected override void OnEnable()
         {
-            base.OnEnable();
+            // Cache navigatable references for the A/B position buttons
+            _aPositionNav ??= _aPositionText.GetComponentInParent<NavigatableBehaviour>();
+            _bPositionNav ??= _bPositionText.GetComponentInParent<NavigatableBehaviour>();
+
+            // Push scheme with Left/Right for position adjustment (replaces base.OnEnable scheme)
+            Navigator.Instance.PushScheme(new NavigationScheme(new()
+            {
+                NavigationScheme.Entry.NavigateSelect,
+                new NavigationScheme.Entry(MenuAction.Red, "Menu.Common.Back", Back),
+                NavigationScheme.Entry.NavigateUp,
+                NavigationScheme.Entry.NavigateDown,
+                new NavigationScheme.Entry(MenuAction.Left, "Menu.Common.Decrease", () => AdjustSelectedPosition(-1.0)),
+                new NavigationScheme.Entry(MenuAction.Right, "Menu.Common.Increase", () => AdjustSelectedPosition(1.0)),
+            }, false));
 
             UpdatePositionText();
+        }
+
+        private void AdjustSelectedPosition(double deltaSeconds)
+        {
+            var selected = NavigationGroup.CurrentNavigationGroup?.SelectedBehaviour;
+            if (selected == null) return;
+
+            if (selected == _aPositionNav)
+            {
+                double newTime = Math.Max(0, GameManager.PracticeManager.TimeStart + deltaSeconds);
+                GameManager.PracticeManager.SetAPosition(newTime);
+                UpdatePositionText();
+            }
+            else if (selected == _bPositionNav)
+            {
+                double newTime = Math.Max(0, GameManager.PracticeManager.TimeEnd + deltaSeconds);
+                GameManager.PracticeManager.SetBPosition(newTime);
+                UpdatePositionText();
+            }
         }
 
         public void SetAPosition()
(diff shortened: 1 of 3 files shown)


Assets/Script/Gameplay/HUD/Pause/PracticePause.cs
```diff
@@ -12,9 +15,74 @@ public class PracticePause : GenericPause
...
+        private NavigatableBehaviour _aPositionNav;
+        private NavigatableBehaviour _bPositionNav;
+        private float _scrollTimer;
+
         protected override void OnEnable()
         {
-            base.OnEnable();
+            // Cache navigatable references for the A/B position buttons
+            _aPositionNav ??= _aPositionText.GetComponentInParent<NavigatableBehaviour>();
+            _bPositionNav ??= _bPositionText.GetComponentInParent<NavigatableBehaviour>();
```
wyrdough: Unity object lifetime is weird, so you can't use null conditionals on components that derive from MonoBehavior (or any other Unity classes). The state of the underlying C++ object does not always match its C# representation, so you have to do an explicit `== null` or `!= null` since Unity overrides those operators so that both are checked.

Assets/Script/Gameplay/HUD/Pause/PracticePause.cs
```diff
@@ -12,9 +15,74 @@ public class PracticePause : GenericPause
...
+
+            // Push scheme with Left/Right for position adjustment (replaces base.OnEnable scheme)
+            Navigator.Instance.PushScheme(new NavigationScheme(new()
```
wyrdough: I haven't actually tested anything, but I suspect you're leaving a scheme on the stack, which may cause issues down the line. You should probably pop the one set in base before pushing this one.

Assets/Script/Gameplay/HUD/Pause/PracticePause.cs
```diff
@@ -12,9 +15,74 @@ public class PracticePause : GenericPause
...
+            else if (delta < 0f)
+            {
+                AdjustSelectedPosition(-SCROLL_ADJUST_SECONDS);
+                _scrollTimer = SCROLL_TIME;
+            }
+        }
+
+        private void AdjustSelectedPosition(double deltaSeconds)
+        {
+            var selected = NavigationGroup.CurrentNavigationGroup?.SelectedBehaviour;
```
wyrdough: See null conditional comment above...

wyrdough (changes requested):
Thanks for the PR! This seems like a nice quality of life improvement for those who use A/B positioning. I haven't tested anything yet, but I did note a couple of minor issues with the code.

Is there any UI to indicate that the position has changed? (I don't use practice mode much, so I'm not sure if there might be some timestamps that get updated when the positions are changed or what)

neilweber:
> Is there any UI to indicate that the position has changed? (I don't use practice mode much, so I'm not sure if there might be some timestamps that get updated when the positions are changed or what)

Yes there is.  One menu item shows the timestamp of the start position and another menu item shows the timestamp of the end position.  The timestamp updates immediately as it is adjusted.

Thanks for looking at my PR.  I will make the appropriate code changes.

Assets/Script/Gameplay/HUD/Pause/PracticePause.cs
neilweber: This navigation scheme is replacing the one the base class GenericPause was pushing.  GenericPause is/was already popping the pushed navigation scheme in OnDisable() so this sub class doesn't need to.

neilweber:
I converted the null check operators with if-statements.  Thanks, again!

One other thing I've been thinking about regarding this change.  Currently the practice HUD does not display the current timestamp.  Instead it displays the current section.  The section name is of limited value.  With this change, the current timestamp becomes more useful and really should be displayed. Should the current timestamp replace the section name or should it be added?

polson:
1. Clicking on the A position and B position does weird things.  Not sure if that's a regression or not though


https://github.com/user-attachments/assets/e5930348-9839-4acc-8297-350774aa1b57



2. Menu only shows "Decrease" when I can both increase and decrease
3. I can increase past the song's length (this song is like 3 minutes long)

CleanShot 2026-02-19 at 11 08 48

4. It would be nicer if we didn't show the `Increase/Decrease` button if you aren't hovering over a Set A Position or Set B position row.  But if it's not easy to do I could consider it a bug in the menu handling instead and out of scope

neilweber:
Clicking on the Set A/B position w/ the mouse is the original awkward behavior.  On mouse click, YARG sets the position to the strikeline.  And that was the only way to adjust the A/B position without going back and picking start and stop sections.  Very non-granular and the reason for my code changes.

That's a good catch about going past the end of the song.  I'll fix that.

I hadn't noticed that bottom menu was affected.  I'll look into it.

Thanks for looking!

polson:
Ok we can ignore the click behavior if that's how it was before (not a regression)

neilweber:
The code now clamps the end position B to the song length.  It also clamps the start position A to the song length though this is not really necessary because the existing code ensured that the start position was less than the end position.

Regarding the "Decrease" button label: HelpBar.cs skims through the scheme actions looking for left/right actions to coalesce into a single button.  It chooses the first action (Decrease in this case) to set the label.  The code in MusicLibraryMenu.cs uses a special localization key for the button label.  I'd have to do the same to change "Decrease" to something else.

To only have the "Decrease/Increase" button display while either "Set (A/B) Position" menu item is active is a little more involved.  Could push/pop different schemes depending on which menu item is active.

Instead, I went with the pattern used in PracticeManager.  Instead of specifying the Increase/Decrease in the navigation scheme, listen for the Left/Right entries directly.  This completely removes the "Decrease" button.  This approach is a little cleaner in the code but at the expense of discoverability.

polson (approved):

PR #1335, opened 2026-02-18
Small UI Tweaks, Round 1

Dansla116:
- Show scrollbar thumb on first load and proper default "empty" `Sidebar` content
- When typing in search, force `MusicLibraryMenu` to first song position when adding text, or last song position when removing text.  **In other words: 1)** When adding more text to the search bar, make sure we stay at the top (first result), which is the most likely match.  **2)** When removing text (backspace/delete) from the search bar, stay on the last song position we were on.  This will be useful when sorting by artist, searching for a song, remove the search bar text, and having all that artists songs in the area.
- Update `Sidebar` and preview when doing actions that move scroll position automatically (e.g. Searching, Re-scanning Songs)
- Shrink/Wrap Charter and Source with same logic for Album.  Considered Unity RichText so <color="#ff0000"> and what not get discarded when deciding when to shrink/wrap, but will still show up in the display.  Also, made length criteria slightly less to accommodate text with lots of capital "W" and not many lowercase "l".
- Fixed a race condition in album art loading and reuse the same cancellation token for both sidebar album covers.  This fixes an occasional issue when coming from `FilterMenu` to `MusicLibraryMenu` where `Sidebar` would only have the smaller foreground album art or the larger background album art, but not both.

### DIFF
--- a/Assets/Script/Helpers/Extensions/ImageExtensions.cs
+++ b/Assets/Script/Helpers/Extensions/ImageExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using Cysharp.Threading.Tasks;
@@ -62,14 +63,34 @@ namespace YARG.Helpers.Extensions
             return texture;
         }
 
-        private static SongEntry _current = null;
+        private sealed class AlbumCoverRequestState
+        {
+            public int RequestId;
+            public SongEntry SongEntry;
+        }
+
+        private static readonly Dictionary<int, AlbumCoverRequestState> _currentByImage = new();
         public static async void LoadAlbumCover(this RawImage rawImage, SongEntry songEntry, CancellationToken cancellationToken, float alpha = 1)
         {
-            _current = songEntry;
+            if (rawImage == null)
+            {
+                return;
+            }
+
+            var instanceId = rawImage.GetInstanceID();
+            if (!_currentByImage.TryGetValue(instanceId, out var state))
+            {
+                state = new AlbumCoverRequestState();
+                _currentByImage[instanceId] = state;
+            }
+
+            state.RequestId++;
+            state.SongEntry = songEntry;
+            var requestId = state.RequestId;
             using var image = await UniTask.RunOnThreadPool(songEntry.LoadAlbumData);
             // Everything that happens after this conditional should only happen *once*
             // There's no reason to create and destroy texture objects if they're just gonna be overriden
-            if (_current != songEntry)
+            if (!_currentByImage.TryGetValue(instanceId, out state) || state.RequestId != requestId || state.SongEntry != songEntry)
             {
                 return;
             }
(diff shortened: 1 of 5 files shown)


Assets/Script/Menu/MusicLibrary/Sidebar.cs
```diff
@@ -138,15 +141,15 @@ public void RefreshFavoriteState()
...
 
-        public void UpdateSidebar()
+        public void UpdateSidebar(bool force = false)
         {
             if (_musicLibraryMenu.ViewList.Count <= 0)
             {
                 return;
             }
 
             var viewType = _musicLibraryMenu.CurrentSelection;
```
wyrdough: It would be more obvious what you're doing here if you called this variable `selected` or something like that.

Assets/Script/Menu/MusicLibrary/Sidebar.cs
```diff
@@ -284,31 +286,45 @@ private void ShowSongInfo(SongViewType songViewType)
...
-            const int maxCharsBeforeWrap = 50;   // number of characters before we wrap the text
+            const int maxCharsBeforeShrink = 36; // number of characters before we shrink the text
+            const int maxCharsBeforeWrap = 45;   // number of characters before we wrap the text
             const float shrinkFactor = 0.8f;     // percentage to shrink the font by after shrink threshold
 
-            if (_albumBaseFontSize <= 0f)
-                _albumBaseFontSize = _album.fontSize > 0f ? _album.fontSize : 20f;
+            if (string.IsNullOrEmpty(text))
+            {
+                container.SetActive(false);
```
wyrdough: Do we actually want to remove the container from view if the album/artist/whatever happens to be empty? I don't think having everything jump around as you scroll is necessarily a better option than just having blank space...

Assets/Script/Helpers/Extensions/ImageExtensions.cs
```diff
@@ -62,14 +63,34 @@ public static Texture2D LoadTexture(this YARGImage image, bool mips)
...
+            }
+
+            state.RequestId++;
+            state.SongEntry = songEntry;
+            var requestId = state.RequestId;
             using var image = await UniTask.RunOnThreadPool(songEntry.LoadAlbumData);
             // Everything that happens after this conditional should only happen *once*
             // There's no reason to create and destroy texture objects if they're just gonna be overriden
-            if (_current != songEntry)
+            if (!_currentByImage.TryGetValue(instanceId, out state) || state.RequestId != requestId || state.SongEntry != songEntry)
```
wyrdough, line 120: So just to be sure I'm understanding here, we saved the fact that we made a request and a reference to the request state above. And here we're checking if the request was logged, that the request ID is what we expect, and that the request is for the expected `SongEntry` and if not, we are abandoning the request.

What I don't see is where states saved in `_currentByImage` ever get cleared after being completed (whether handled or cancelled/failed)

wyrdough (changes requested):
Some thoughts without the benefit of having actually tested anything..

Dansla116:
Copy/paste from git commit message.  Addressed code review feedback.
- changed variable `viewType` to `selected` for updating sidebar
- keep sidebar text containers active to prevent layout shift
- clean up album cover request state
- made goto top on recommended re-roll more explicit because it did not seem guaranteed
- very minor code cleanup

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -694,6 +707,39 @@ private void UpdateSearch(bool force)
             {
                 EnsureValidSelectionAfterFilter();
             }
+
+            if (!PlaylistMode && searchChanged)
```
polson, line 712: Could you add a comment to this block helping describe what it does?  It's not obvious what it's doing otherwise

polson (commented):
If you don't mind adding the comment it will probably prevent a headache later, then we can merge

PR #1336, opened 2026-02-19
Add subgenre to CSV export

trainlos:
Also removes redundant conditional that was added with the new music library UI

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/Sidebar.cs
+++ b/Assets/Script/Menu/MusicLibrary/Sidebar.cs
@@ -353,14 +353,7 @@ namespace YARG.Menu.MusicLibrary
             }
             else
             {
-                if (entry.HasInstrument(Instrument.ProDrums))
-                {
-                    _difficultyRings[2].SetInfo("realDrums", Instrument.ProDrums, entry[Instrument.ProDrums]);
-                }
-                else
-                {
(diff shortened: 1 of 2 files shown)


Assets/Script/Menu/MusicLibrary/Sidebar.cs
```diff
@@ -353,14 +353,7 @@ private void UpdateDifficulties(SongEntry entry)
...
             else
             {
-                if (entry.HasInstrument(Instrument.ProDrums))
-                {
-                    _difficultyRings[2].SetInfo("realDrums", Instrument.ProDrums, entry[Instrument.ProDrums]);
-                }
-                else
-                {
-                    _difficultyRings[2].SetInfo("drums", Instrument.FourLaneDrums, entry[Instrument.FourLaneDrums]);
-                }
```
wyrdough, line 363: Do we not still require this since pro and 4L are still using the same diff ring?

trainlos: When the new music library got merged it duplicated this bit so this is just removing one of those

wyrdough: Ah, gotcha, diff view is hard lol

wyrdough (approved):

PR #1338, opened 2026-02-19
Implement advanced settings toggle

polson:
Adds a blue button for show/hide advanced settings

Advanced settings have a small label indicating they are advanced

Whenever we toggle, we scroll to the top of the screen to avoid a jarring transition in which settings are displayed

**I am open to feedback on which settings should be basic or advanced**

Video:

https://github.com/user-attachments/assets/a399b550-444b-40df-a104-bc1a50a4ee9a


wyrdough:
I had hoped to get things so that there was an arrow on the sections that had advanced settings and collapse/expand the advanced settings per section, but it was a right pain in the ass so I never finished. Given that, I'm fine with the button press/navbar click to swap between showing and hiding advanced settings.

That said, I think I preferred the vertical blue stripe on the left to denote advanced settings over the tag you have now. Maybe we could get some more feedback from the artisty types...

polson:
> I had hoped to get things so that there was an arrow on the sections that had advanced settings and collapse/expand the advanced settings per section, but it was a right pain in the ass so I never finished. Given that, I'm fine with the button press/navbar click to swap between showing and hiding advanced settings.
> 
> That said, I think I preferred the vertical blue stripe on the left to denote advanced settings over the tag you have now. Maybe we could get some more feedback from the artisty types...

Yeah I don't love how this looks either.  I don't even love how it works.  

We could try to do advanced per section, and I think that would help people only show advanced settings they care about.  

The only thing about that is we would probably still want a "show all advanced" because some people like me just want to see every possible setting without having to click a dozen times.  So maybe we could start here and do both?

polson:
Updated screenshot

CleanShot 2026-02-27 at 22 47 38@2x

polson:
CleanShot 2026-03-03 at 15 59 52@2x

Updated UI

polson:
CleanShot 2026-03-08 at 14 59 55@2x

PR #1339, opened 2026-02-20
Implement random sort

polson:
As it says, shows a long list of randomly sorted songs

Video:

https://github.com/user-attachments/assets/86fda18b-9c15-43d1-b70c-fe081490b623

### DIFF
--- a/Assets/Script/Song/SongContainer.cs
+++ b/Assets/Script/Song/SongContainer.cs
@@ -1,4 +1,4 @@
-﻿using Cysharp.Threading.Tasks;
+using Cysharp.Threading.Tasks;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


wyrdough:
Probably would be good to add a header below playlists indicating that we're using random sort.

Other than that, code looks deliciously simple. Great work! 👍

polson:
It does have the standard "Sorted by Random" at the top, do you think we should add another header?

wyrdough:
The underlying problem here is how button headers look like...headers lol

polson:
It appears we are already hiding the headers if there's only one header?  

Given that's the case, "no header" on random sort is consistent with that behavior.

Here's an example of sorting by source, when all songs are the same source.  You can see the header is hidden.

If we want to change this behavior, that's cool, but maybe in a separate PR

CleanShot 2026-02-24 at 15 43 42@2x

PR #1342, opened 2026-02-21
Small UI Tweaks, Round 2

Dansla116:
- Various fixes to scroll position and song preview
  - When changing sort (from `FiltersMenu` or `PopupMenu`), move scroll position to match last hovered song when hovered over a song, or fallback on first song below a header when hovered over a header
  - When filtering out songs, stay on song if it is still visible, or go to next header up if song has been filtered out
  - When hovering over a recommendation, going to `FiltersMenu` and hiding recommendations moves you to first header
  - When re-scanning songs (and therefore changing recommendations), fixed song-preview/`Sidebar` still showing old recommendation
  - When re-rolling recommended songs, keep scroll position on "Recommended Songs" header allowing multiple re-rolls
  - When navigating to `Playlist` summary, stop song-preview and remove info from `Sidebar`
  - When navigating to specific `Playlist` start song-preview
  - When navigating from specific `Playlist` back to `Playlist` summary, move scroll position to most recent playlist
  - When using a dynamic sort (Stars/Play Count) and playing a song, don't move scroll position to match song and instead keep scroll position in old header
- Removed duplicate `Playlist` header and back button
- Fixed `Playlist` summary count and individual `Playlist` star count
- Fixed `ToastManager` z-index and navigation to `SettingsMenu` "Songs" tab
- Fixed weird layout issues when pressing buttons while loading/caching


Dansla116:
In testing, I found several more scroll position / preview / `Sidebar` issues:
- When playing a song within a playlist, and going back to playlist summary, you would snap to top of the list and the preview would play for the most recent song you were on.
- When **recommendations were hidden and...** hovering over a header and changing sort via `PopupMenu`, you would go to the first song under that header.  When hovering over a header and changing sort via `FiltersMenu`, you would go to the header for the first song under the old header.  Made the latter match the former.
- When **recommendations were shown and...** hovering over a header and changing sort via `PopupMenu`, you would go to first recommended song.  When hovering over a header and changing sort via `FiltersMenu`, you would go to the first recommended song in it's natural `MusicLibraryMenu` location.  Made these work the same as when recommendations are hidden.
- When hovered over a header, going to `FiltersMenu`, and not changing anything that affects your header, you would still get moved to the first song below that header.

Other issues actually not related to scroll position:
- Fixed `MusicLibraryMenu` banner song and star count updating when searching to reflect song count and star count within search criteria just like filters.
- Fixed `MusicLibraryMenu` banner showing filtered down totals when applying filter/search and re-scanning library.
- Removed duplicate "Search Results" and summary header in favor of static banner.  Removed "Sorted by {sort}" in banner because the sort is always most likely match first.
- Allow navigating to "Random Song" and "Playlists" with green button while setlist has content.

Also did minor code cleanup and merge.

Dansla116:
- Fix "Go To Currently Playing" song
- When using a dynamic sort (Stars/Play Count) and playing a song, don't move scroll position to match song and instead keep scroll position in old header... again
- Fix lag when sorted by stars and searching, search overrides sort to match anyway

wyrdough:
Quick testing seems to indicate you did indeed fix the things you said you did. Couple of things to think about for the future, though:

* We're still not following the active song when using the search bar (previously, the selected song would get pinned to the top of search until it was filtered out)
* Clicking on recommended still doesn't just keep the current selectedindex. I think the play here is just to leave SelectedIndex alone entirely when the recommended re-roll is activated.

Give me a few to look at the code a bit more.

Assets/Script/Menu/Filters/FiltersMenu.cs
```diff
@@ -1431,17 +1434,55 @@ private void OnDisable()
         {
             if (!_ready) return;
 
+            bool filtersChanged = HaveFiltersChanged();
```
wyrdough: Not really seeing the point of declaring filtersChanged here when you could just do `if (HaveFiltersChanged() || showRecommendationsChanged)`. If the name of the method were particularly long, you were passing a bunch of parameters, or were using it repeatedly, sure.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -198,6 +212,19 @@ protected override void OnEnable()
...
 
+            if (MenuState == MenuState.Library && _hasSavedSelectionSnapshot)
+            {
+                RestoreSelectionSnapshot(_savedSelectionSnapshot);
+                _hasSavedSelectionSnapshot = false;
+            }
+
+            if (_forceGoToCurrentlyPlaying && MenuState == MenuState.Library && !PlaylistMode)
+            {
+                bool selected = TrySelectCurrentSongPreferNaturalLocation(_forceGoToSong ?? _currentSong);
```
wyrdough: Seems like `selected` isn't actually getting used and `TrySelectCurrentSongPreferNaturalLocation` is only used by this, so unless there's some other reason to keep them, get rid of them.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -515,7 +540,9 @@ private List<ViewType> CreateNormalViewList()
                 }
 
                 SortHeaderViewType sortHeader = null;
-                if (showSortHeaders)
+                bool hideSearchResultsHeader = _searchField.IsSearching &&
```
wyrdough, line 546: Why would we ever hide the search results header if we're searching?

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -656,17 +719,40 @@ private void UpdateSearch(bool force)
...
             _currentSearch = currentSearch;
+            _searchField.UpdateSearchText();
 
-            if (_reloadState != MusicLibraryReloadState.Partial && !searchChanged)
+            var predicate = YARG.Menu.Filters.FiltersMenu.ActiveFilterPredicate;
+            bool inLibrary = !PlaylistMode && MenuState == MenuState.Library;
+            bool shouldApplyFilters = inLibrary && predicate != null;
+            bool shouldShowFilteredCounts = inLibrary && (_searchField.IsSearching || predicate != null);
+            if (shouldApplyFilters)
+                _sortedSongs = ApplyFilterPredicate(_sortedSongs, predicate);
```
wyrdough: Missing braces around the body of the if statement. Also add a blank line between your conditions and the if to break it up a little.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -656,17 +719,40 @@ private void UpdateSearch(bool force)
...
-                    newPositionStartIndex = _primaryHeaderIndex;
-                }
+                var baseList = SongContainer.GetSortedCategory(SettingsManager.Settings.LibrarySort);
+                _totalSongCountUnfiltered = CountSongs(baseList);
+            }
+            else
+            {
+                _totalSongCountUnfiltered = 0;
+            }
+            RequestViewListUpdate();
```
wyrdough, line 746: Also stick some blank lines around `RequestViewListUpdate`

Assets/Script/Menu/Navigation/Navigator.cs
```diff
@@ -3,6 +3,7 @@
 using System.Linq;
 using Cysharp.Threading.Tasks;
 using UnityEngine;
+using YARG;
```
wyrdough: Not sure how this got there...

Assets/Script/Menu/Navigation/Navigator.cs
```diff
@@ -103,7 +104,31 @@ public HoldContext(NavigationContext context)
...
+                    _repeatInputs.Clear();
+
+                    for (int i = _holdInputs.Count - 1; i >= 0; i--)
+                    {
+                        _holdInputs[i].Tracker.StopHolding();
+                        _holdInputs[i].Tracker.ClearEvents();
+                    }
+
+                    _holdInputs.Clear();
+                }
```
wyrdough: I know we use inline setters elsewhere, but I don't care for them unless they're just a couple of lines. Extract the body into a `SetDisableMenuInputs(bool value)` method and do the work there. Then this can be just `set => SetDisableMenuInputs(value);`

Assets/Settings/AddressableAssetsData/AssetGroups/Default Local Group.asset
```diff
@@ -175,11 +175,6 @@ MonoBehaviour:
     m_ReadOnly: 0
     m_SerializedLabels: []
     FlaggedDuringContentUpdateRestriction: 0
-  - m_GUID: 9fe3f03b859cb4846bfe733521bdbe70
```
wyrdough, line 178: It appears something went wrong when merging/rebasing. This shouldn't be getting disappeared.

Assets/Settings/AddressableAssetsData/AssetGroups/Default Local Group.asset
```diff
@@ -245,11 +240,6 @@ MonoBehaviour:
     m_ReadOnly: 0
     m_SerializedLabels: []
     FlaggedDuringContentUpdateRestriction: 0
-  - m_GUID: edae7e06bc0cb23478dcea1b38796c5d
```
wyrdough, line 248: Also this...

wyrdough (changes requested):
I'm going to have to leave wrapping my head around the snapshot thing in the music library until later. It seems a bit overcomplicated, but that may just be because I'm tired, not sure.

In the meantime, there are some minor cleanups to be made and an inadvertent removal of some Addressables that need to be dealt with.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
Dansla116: With the **New UI**, we actually already made the "Search Results" header invisible.  I'm guessing with `ViewObject.Hide()` somewhere.  But it still takes up physical space on the screen.  So here, I am actually removing it so it doesn't take up space.
Also from the **New UI**, we've added a static header (aka a banner) that normally says "Sorted by {sort}".  But when searching, that's not really accurate because we're sorted by relevance.  So I've changed that to say "Search Results", so it better replaces the old hidden header.

I'll add a comment in code to clarify.

A before/after picture probably explains it better.  Notice, when searching, the banner has changed from "Sorted by Artist" to "Search Results", and you can see one more result in the screenshot.

image
image

Assets/Settings/AddressableAssetsData/AssetGroups/Default Local Group.asset
Dansla116: Never mind my reply before edit.  I totally did not fully read what you highlighted.  I'll get my deletion reverted.

Dansla116:
I made changes on everything you pointed out.  Or left a reply on the 2-3 questionable things.  I still haven't gone through the whole `SelectionSnapshot` logic and added comments like I said in Discord.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -997,18 +1091,397 @@ public void RefreshAndReselect(bool selectTopOfList = false)
...
+            if (MenuState != MenuState.Library || PlaylistMode)
+                return null;
+
+            if (previousSong != null || CurrentSelection is SongViewType)
+                return null;
+
+            return GetFirstSongAfterIndex(selectedIndex);
+        }
+
+        private void RestoreSelectionSnapshot(SelectionSnapshot snapshot)
```
polson, line 1196: Without looking at this too deeply, I do think this could be greatly simplified.  A simpler solution would probably be to assign stable ids to each row.  

We have 4 different kinds of `ViewType`s to deal with.  I think we should add into a the base `ViewType` class an abstract StableID string property that the subclasses will implement.  This ID should uniquely identify the view

For SongViewtype for example, you would use the Hash and the file path (because of duplicate songs)

Something like this:

```
public override string StableId => $"Song_{SongEntry.Hash}_{SongEntry.ActualLocation}";
```

Then your Snapshot only has to remember IDs, not every bit of information about the row, for every kind of row all smooshed together.

Then as long as we have a way to select an item in the list by ID, I think we could reduce the complexity here pretty massively.  But feel free to let me know your thoughts

polson (changes requested):

polson:
I see this error in the logs sometimes when exiting the filters menu.  Not sure if its related to these changes or not

```
GameObject is already being activated or deactivated.
UnityEngine.GameObject:SetActive (bool)
YARG.Menu.MenuManager:ReactivateCurrentMenu () (at Assets/Script/Menu/MenuManager.cs:149)
YARG.Menu.Filters.FiltersMenu:OnDisable () (at Assets/Script/Menu/Filters/FiltersMenu.cs:1460)
UnityEngine.GameObject:SetActive (bool)
YARG.Menu.Filters.FiltersMenu:HandleBack () (at Assets/Script/Menu/Filters/FiltersMenu.cs:222)
UnityEngine.EventSystems.EventSystem:Update ()
```

Dansla116:
Addressed more code-review feedback

@polson
- Refactored the music library selection to use stable view IDs instead of ad‑hoc comparisons
- Added `StableId` across view types (song, header, category, button, playlist), plus a song content ID to match entries even when context changes
- Selection refresh now preserves index when updating recommended songs and keeps the highlight stable across filter/recommendation changes

@wyrdough
> We're still not following the active song when using the search bar (previously, the selected song would get pinned to the top of search until it was filtered out)

Thinking about it, do we actually still want to do this?  Like, if you're searching for something, you probably didn't care to play what you were already on, right?  And doing that also adds some complexity like the frustration that's been reported.
For example, how it worked before my last **Round 1** change, if I type "Rag", the first result is "Rag Doll" by Aerosmith.  I add an "e" to spell "Rage", the first result is a song by Rage Against the Machine.  But because "Rag Doll" is still a fuzzy match further down, I jump to that.
With my change, every addition snaps you back to the top, and every deletion will keep you on the current song.

> Clicking on recommended still doesn't just keep the current selectedindex. I think the play here is just to leave SelectedIndex alone entirely when the recommended re-roll is activated.

I've made a change so re-rolling recommended won't change index at all.

Dansla116:
Also, just fixed those `GameObject is already being activated or deactivated.` errors.  It was caused by a last minute change to the original FiltersMenu implementation where I moved them from PersistentScene into MusicLibraryMenu and made ReactivateCurrentMenu() run during the same enable/disable chain.

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
wyrdough: ah, gotcha, makes sense

Assets/Script/Menu/Filters/FiltersMenu.cs
Dansla116: I just got out of this rabbit hole where I tried inlining `HaveFiltersChanged()`, but that introduced an issue where filter changes without sort changes wouldn't trigger a refresh.  I was thinking it was something to do with a `ReactiveCurrentMenu()` race that I fixed around the same time, but it was not.
We need to evaluate `HaveFiltersChanged()` **and save that state** before SaveFilters() runs.  If we inline it inside the conditional, it executes after the save, which overwrites the previous state and prevents the change detection from firing.
So the local isn't just for readability.  It preserves the pre-save comparison.
I've added a comment to clarify this since it's definitely not obvious.

Assets/Script/Menu/Filters/FiltersMenu.cs
```diff
@@ -1431,17 +1434,57 @@ private void OnDisable()
...
 
+            bool showRecommendationsChanged = _showRecommendationsOnOpen !=
+                SettingsManager.Settings.ShowRecommendedSongs.Value;
+            // Must check before SaveFilters() so we don't overwrite the previous state
+            // otherwise filter changes won't trigger a library refresh
+            bool filtersChanged = HaveFiltersChanged();
             SaveFilters();
             ActiveFilterPredicate = BuildFilterPredicate();
             var library = FindFirstObjectByType<MusicLibrary.MusicLibraryMenu>();
             library?.SetSidebarDifficultiesVisible(true);
```
polson: you can move this down into the if

Assets/Script/Menu/MusicLibrary/ViewTypes/SongViewType.cs
```diff
@@ -24,18 +24,24 @@ public class SongViewType : ViewType
...
         private PlayerScoreRecord _playerScoreRecord;
         private PlayerScoreRecord _playerPercentRecord;
         private GameRecord _bandScoreRecord;
 
-        public SongViewType(MusicLibraryMenu musicLibrary, SongEntry songEntry)
+        public SongViewType(MusicLibraryMenu musicLibrary, SongEntry songEntry, string context = "library")
         {
             _musicLibrary = musicLibrary;
             SongEntry = songEntry;
+            _contentStableId = $"Song:{SongEntry.Hash}_{SongEntry.ActualLocation}";
```
polson: I feel like we don't need two IDs here, but I could be wrong.

Could we uniquely find the exact row we want by using a combination of the closest header stable Id + the _contentStableId here, then we dont need the other _stableId?

Seems to me like it could work.  Then restoring would be:

1. Find that header, then find matching content under that header 
2. If header missing, find the matching song anywhere
3. If song missing, then existing fallbacks

What do you think?

polson:
This is a rather large and complex PR so bear with me as this might take some iterations to get it into good shape.  

I am still thinking we could simplify things further.  We should think about the minimum information needed to handle finding the rows and the fallbacks.  Right now I feel like we have more mutable state than we need to do that.

I am wondering if we can uniquely identify a row by two things:

1. the row's StableID
2. the StableID of the closest header (or category) above it (or the same if the row is already a header)

If that's all we need I think we could simplify this more.  What do you think?

Dansla116:
I tried, but that didn't work out.  I think I ended up adding more lines. 😅
I **did** find some redundancy in having a selected song or header and where we navigate the user when all the content under the header (and therefore, the header itself) gets hidden.  So I cleaned that up.

Then I added comments for every return point to explain what happened and why we're taking the user where we're taking them.  So hopefully, future people (myself included), will have an easier time following along.

Otherwise, I couldn't find a way to get rid of any of the snapshot data.

`SelectedIndex` keeps selection near the same index when the list reorders or shrinks
`SelectedStableId` restores the exact selected row when it still exists
`SelectedSongContentStableId` restores the same song even if its row/context changes (recommended vs natural position in library)
`HeaderStableId` identifies the header above the selection to keep you in the same section when some songs are filtered out
`HeaderFirstSongContentStableId` is for when a header disappears after a sort change, we land on the first song that was under that old header
`HeaderPreviousSongContentStableId` is for when all the songs under a header (and therefore, also the header) gets filtered out, we land on the last song above it to avoid jumping to the top
`PreserveIndexOnDynamicSort` is for dynamic sorts like Playcount/Stars, when the list reorders after a song is played - we intentionally keep the same index instead of the same song

polson (approved):
The Music LIbrary flickers when we return from the filters menu after making changes

But other than that, the functionality looks good.  If we want try to fix that in a separate PR that's fine.

`MusicLibraryMenu.cs` is in desperate need of cleanup generally.  I think it's doing too much and is very difficult to read.  For now I'm ok with just getting this in and dealing with the tech debt later (hopefully)

wyrdough (approved):
I am not happy with how crazy `MusicLibraryMenu` is getting, but it needs a refactor and probably another partial or two and that isn't your problem. Good enough for now.

PR #1346, opened 2026-02-22
More descriptive exception when attempting to build Venue / Character without MacOS build support

gingerphoenix10:
If the `_metal_shaders.bytes` file is still missing after running BuildAssetBundles, create a log and popup guiding to the Venue Creation wiki page


wyrdough (approved):
lgtm

wyrdough:
Just as a note to self, we also should clean up the mess that gets created in the `Assets` folder when the metal shader build fails. (Or maybe use some other folder that we unconditionally remove after the build as temporary storage)

PR #1347, opened 2026-02-22
Auto-set practice positions from current song position

neilweber:
Finding a troublesome sequence in Practice Mode is awkward.  This PR attempts to make it easier to transition from playing a song to practicing those sequences.  The idea is that after messing up a sequence, the player can pause gameplay and then enter practice mode already initialized with the current song position.  This applies to both starting a practice from the pause menu and starting a practice from the fail menu.  It does not apply when starting practice after using the Practice main menu item.

When practice mode is entered, the current song position is used for the end/B position.  The start/A position is ten seconds earlier.  The player can adjust either before starting practice.

### DIFF
--- a/Assets/Script/Gameplay/PracticeManager.cs
+++ b/Assets/Script/Gameplay/PracticeManager.cs
@@ -12,6 +12,7 @@ namespace YARG.Gameplay
     public class PracticeManager : GameplayBehaviour
     {
         private const double SECTION_RESTART_DELAY = 2;
+        private const double PRACTICE_ENTRY_LOOKBACK = 10; // seconds
 
         [Header("References")]
         [SerializeField]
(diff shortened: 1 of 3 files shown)


wyrdough:
Thanks for the PR! Practice doesn't get a lot of love, so it's especially good to see this. One question from someone who doesn't often use practice mode:

Would it not be better to set the bounds to entirety of the current section? (or maybe previous and current if we're just barely into the new section)

neilweber:
I find that sections are too large to be used as "units of practice" if that makes any sense.  At least for me, even ten seconds is often too long to memorize and develop muscle memory.  The primary purpose of this change is make it easy to locate a troublesome one bar fill in a 4 minute song.  Once the fill is pinpointed, I'll adjust the start and stop so I can loop that fill over and over and over.

Assets/Script/Gameplay/PracticeManager.cs
```diff
@@ -106,6 +107,32 @@ private void OnNavigationEvent(NavigationContext ctx)
...
+                    var startSection = sections[0].Time <= startTime
+                        ? sections.Last(s => s.Time <= startTime)
+                        : sections[0];
+                    var endSection = sections[0].Time <= endTime
+                        ? sections.Last(s => s.Time <= endTime)
+                        : sections[0];
+                    SetPracticeSection(startSection, endSection);
+                    SetAPosition(startTime);
+                    SetBPosition(endTime);
+                    GameManager.SetSongTime(TimeStart, SettingsManager.Settings.PracticeRestartDelay.Value);
```
polson: Also move this into its own var: SettingsManager.Settings.PracticeRestartDelay.Value

Assets/Script/Gameplay/PracticeManager.cs
```diff
@@ -106,6 +107,32 @@ private void OnNavigationEvent(NavigationContext ctx)
...
+            {
+                double endTime = GlobalVariables.State.SavedInputTime.Value;
+                double startTime = Math.Max(0.0, endTime - 10.0);
+
+                var sections = _chart.Sections;
+                if (startTime < endTime && sections.Count > 0)
+                {
+                    // Fall back to sections[0] when startTime or endTime precedes the first
+                    // section — some charts have an unsectioned intro before the first marker.
+                    var startSection = sections[0].Time <= startTime
```
polson: This section is hard to read, I think it would be nice if we extracted it to a private method like

```
SectionInfo FindSectionAtTime(double time)
{
return _chart.Sections[0].Time <= time
    ? _chart.Sections.Last(s => s.Time <= time)
    : _chart.Sections[0];
}
```

and then you could do 
```
        var startSection = FindSectionAtTime(startTime);
        var endSection = FindSectionAtTime(endTime);
```

Assets/Script/Gameplay/PracticeManager.cs
```diff
@@ -106,6 +107,32 @@ private void OnNavigationEvent(NavigationContext ctx)
         public void DisplayPracticeMenu()
         {
             GameManager.Pause(showMenu: false);
+
+            if (GlobalVariables.State.SavedInputTime.HasValue)
```
polson: move this into its own var for readability: GlobalVariables.State.SavedInputTime.HasValue

polson (changes requested):

Assets/Script/Gameplay/PracticeManager.cs
```diff
@@ -106,6 +107,32 @@ private void OnNavigationEvent(NavigationContext ctx)
...
+            {
+                double endTime = GlobalVariables.State.SavedInputTime.Value;
+                double startTime = Math.Max(0.0, endTime - 10.0);
```
polson: move 10.0 into a const

wyrdough:
> I find that sections are too large to be used as "units of practice" if that makes any sense. At least for me, even ten seconds is often too long to memorize and develop muscle memory. The primary purpose of this change is make it easy to locate a troublesome one bar fill in a 4 minute song. Once the fill is pinpointed, I'll adjust the start and stop so I can loop that fill over and over and over.

Fair enough... 

Even if it turns out not to be the right general default, it's closer to what the behavior should be than the "lol you figure it out" thing we have now so I'm fine with that.

polson (approved):

JustinAiken:
>  The start/A position is ten seconds earlier

What if it was... a small amount of preroll before the first WRONG note played after breaking a combo?

neilweber:
That could be a future enhancement.  For now I just want something in place.

PR #1349, opened 2026-02-24
Use different checkerboard background pattern (with no abberation)

theli-ua:


polson:
Video:


https://github.com/user-attachments/assets/aec82331-d86f-4d9f-a2e3-24d3fb6df229

polson (approved):

PR #1350, opened 2026-02-24
Prevent sustains from affecting fadeout

theli-ua:
by excluding them from fade pass.
This moves sustains from "Notes" layer into "ExcludeFade" layer. As far as I can tell we don't really use layers anywhere anyway apart from excluding Venue layer from highway rendering.


wyrdough:
So at the risk of asking a stupid question, does this not make it so they don't fade at all?

wyrdough:
Does not seem to cause any regression on Windows/DX11, so just waiting on @polson to check whether it actually fixes the problem on Mac/metal.

polson:
Looks good to me

Before:
image


After:
CleanShot 2026-02-24 at 15 58 33@2x

polson (approved):

theli-ua:
> So at the risk of asking a stupid question, does this not make it so they don't fade at all?

It would make it so that unless there is something else going on in the same pixel it would be fully transparent.
However, since they are above track - track object will contribute correct fade opacity

PR #1351, opened 2026-02-24
Drum highway configuration menus

Frickitickitavi:
[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/333)

Adds new submenus to the profile settings for drums profile types, where the user can arbitrarily rearrange their highway lanes. This replaces the old set of individual split and swap settings, which have been grandfathered in here via legacy parsing that translates them into modern highway orderings.

The UI is fully functional and I'm decently happy with the general workflow/clickpath, but visually it's completely unpolished. It doesn't respond correctly to different aspect ratios, the alignment is a mess, and I'm cobbling together random unrelated assets to make something that looks plausible. You have carte blanche to rip out my programmer art and make a better UI than I could.


Assets/Script/Menu/HighwayConfiguration/DrumsHighwayConfigurationMenu.cs
```diff
@@ -0,0 +1,311 @@
...
+
+        public void Initialize(
+            Dictionary<DrumsHighwayItem, HighwayOrderingItemSpec> specs,
+            IFretColorProvider colorProvider,
+            List<DrumsHighwayItem> defaultList,
+            string header,
+            SetOrdering setOrdering,
+            Instrument instrument,
+            YargProfile profile
+        ) {
```
wyrdough: Would be a little clearer with the `)` immediately after `profile` and the `{` on its own line

Assets/Settings/AddressableAssetsData/AssetGroups/Default Local Group.asset
```diff
@@ -50,11 +50,6 @@ MonoBehaviour:
     m_ReadOnly: 0
     m_SerializedLabels: []
     FlaggedDuringContentUpdateRestriction: 0
-  - m_GUID: 274c3f10426fd914fae0818bfe56db5d
-    m_Address: Setting/FileInfo
-    m_ReadOnly: 0
-    m_SerializedLabels: []
-    FlaggedDuringContentUpdateRestriction: 0
```
wyrdough, line 57: I do not think you intended to mess with the Addressables

Assets/Settings/AddressableAssetsData/AssetGroups/Default Local Group.asset
```diff
@@ -80,11 +75,6 @@ MonoBehaviour:
     m_ReadOnly: 0
     m_SerializedLabels: []
     FlaggedDuringContentUpdateRestriction: 0
-  - m_GUID: 47970bda16e44c543b7f1b72e04335f7
-    m_Address: MusicLibraryUpIcon
-    m_ReadOnly: 0
-    m_SerializedLabels: []
-    FlaggedDuringContentUpdateRestriction: 0
```
wyrdough, line 87: ...

wyrdough (changes requested):
I'm gonna have to do the rest when my brain is actually engaged...

Assets/Settings/AddressableAssetsData/AssetGroups/Default Local Group.asset
Frickitickitavi: what on earth lol

nickjvandyke:
Thanks for attempting this @Frickitickitavi 🙏 I recently centered my snare between my rack toms and love it. But the 4-lane arrangement now doesn't match, so it's hard to hit fast sections with that mental pause to map them in my head. Newbies especially find it unintuitive in jam sessions.

Frickitickitavi:
> Thanks for attempting this @Frickitickitavi 🙏 I recently centered my snare between my rack toms and love it. But the 4-lane arrangement now doesn't match, so it's hard to hit fast sections with that mental pause to map them in my head. Newbies especially find it unintuitive in jam sessions.

@nickjvandyke You should be able to also customize your non-Pro 4L highway to move the snare around. Both MIDI Drumkit and 4L Drumkit profiles should have separate 4L and Pro config menus that are independent from each other

nickjvandyke:
@Frickitickitavi that's with this PR right? Not in nightly?

Frickitickitavi:
> @Frickitickitavi that's with this PR right? Not in nightly?

@nickjvandyke Correct - this PR will replace the current Nightly/Stable functionality where you have a limited set of specific split/swap options (split all tom/cymbal lanes, swap R/Ycym, and swap Bcym/Gcym). Those options will disappear and be replaced with new submenus where you can arbitrarily configure your highway layout for 4L, Pro, and 5L individually. I've designed this PR to grandfather in the old settings so your Pro layout should still look the same once the functionality reaches Nightly, and then you can go and configure 4L how you want it

polson:
One thing to think about and maybe fix here.  If say we released this in a nightly, then a user went back to a previous nightly before this change, profiles will be broken I think?

Frickitickitavi:
> One thing to think about and maybe fix here. If say we released this in a nightly, then a user went back to a previous nightly before this change, profiles will be broken I think?

@polson I don't think they'd be _broken_ per se, but there would be some loss of information. Say we release this in vX. If a player has a vX-1 profile with drum configuration settings and they carry that forward to vX and then revert to vX-1, they'll find that their settings were wiped along the way

And honestly, even that doesn't have to be the case - the wiping of the old settings was just something I added to reinforce the fact that they're fully obsolete and should never be touched outside of grandfathering code. But if we'd rather keep them around so they can act as a time capsule if a player reverts to an old version, that's fine too

Let me know if I'm missing some reason that this is all worse than I realized and I actually _am_ breaking profiles

**EDIT:** I was wrong and it was breaking profiles, but that's now been fixed.

polson (approved):

polson:
Merged manually via squash commit 4ff2aa7ed046bbd1ec676f67278c1b8d99463a3d

nickjvandyke:
Thanks so much for this y'all, I'm super excited for it!! 😀 🥁

PR #1352, opened 2026-02-25
Fix Reggae/Ska special case in CONs

Frickitickitavi:
Genrelizer's Reggae/Ska special case only accounts for how the genre is rendered in MID/.chart conversions (`reggae/ska`), not in raw CONs (`reggaeska`). This fixes that by checking both renderings.

### DIFF
--- a/Assets/Script/Song/Genrelizer.Lists.cs
+++ b/Assets/Script/Song/Genrelizer.Lists.cs
@@ -87,6 +87,7 @@ namespace YARG.Song
         private const string POP_DANCE_ELECTRONIC = "pop/dance/electronic";
         private const string PROG = "prog";
         private const string REGGAE_SKA = "reggae/ska";
+        private const string REGGAE_SKA_RAW = "reggaeska"; // As rendered in raw CONs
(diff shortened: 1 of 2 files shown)


Assets/Script/Song/Genrelizer.Lists.cs
```diff
@@ -87,6 +87,7 @@ public static partial class Genrelizer
         private const string POP_DANCE_ELECTRONIC = "pop/dance/electronic";
         private const string PROG = "prog";
         private const string REGGAE_SKA = "reggae/ska";
+        private const string REGGAE_SKA_RAW = "reggaeska"; // As rendered in raw CONs
```
wyrdough, line 90: Didn't you say it was `reggae_ska` on discord? (I don't actually know, so...)

Frickitickitavi: Oh yeah I did - I was mistaken and `reggaeska` is correct. Underscores replace spaces (`new_wave`) and slashes just disappear (`popdanceelectronic`)

PR #1354, opened 2026-02-26
make low fps camera script

4dispel:
script to make a camera render at a reduced frame rate, motivated by axel's venue with render texture where the rendertexture doesn't respect venue fps.
not sure if setting fps manually or just matching it to venue fps would be better.


wyrdough:
Hm, if it were me, I'd probably make it render at the same rate as the venue on the assumption that if players are messing with the venue fps cap they're doing it for a reason.

wyrdough (approved):

PR #1355, opened 2026-02-28
Reduce vertical padding of music library song row

Ape:
This change is very opionated so I wouldn't mind if you just close the PR as rejected. However, I think it looks a bit better this way.

Before:
before

After:
after


Dansla116:
Oh good, this is exactly one of the things on my list I was planning to do after I got filters and playlists up to the point I want them.

So yeah, looks good to me.  For those like me on 16:9, it looks like this:
image

wyrdough:
Ah, good @Dansla116 answered the one question I had..I think this looks good. Not too cramped but not too spread out. The only thing I think might be worth changing is to make the headers slightly larger, as it seems like the star header might need a tad bit more vertical room.

wyrdough:
One thing I think might be nice is to have the currently selected song be slightly larger than the rest, but I'm certainly not going to make you do that...

PR #1356, opened 2026-02-28
Enable mipmapping for song row icons to reduce texture aliasing

Ape:
Before:
before

After:
image


wyrdough:
Good call.. I'll merge this as soon as I figure out if there are any other outstanding PRs this will create conflicts with...

PR #1358, opened 2026-02-28
Double the resolution of selected music library icons

Ape:
Some of the icons in the new music library UI were noticeably low resolution at least on my 4K screen. I reproduced the ones I noticed with double resolution.

Before:
before1
before2
before3 (the bg box, not the source icon itself)

After:
after1
after2
after3

SVGs in case somebody wants to edit these. Perhaps we should commit the image sources somewhere safe. Or could Unity even render SVGs directly?
MusicLibraryIcons
RedBackButton
SelectedSourceIconBackground


wyrdough:
Can you render the icons that are individual icons at 128x128 or some other multiple of 32? (it makes GPUs happier, supposedly)

Ape:
New version:
- Rebased
- Switched 100px textures to 128px
- Added my art sources here: https://github.com/Ape/YARG.ArtSource

polson (approved):

PR #1359, opened 2026-02-28
FIx kick fret color

Frickitickitavi:
Restores an accidentally-removed piece of `FretArray::Initialize` that sets the color for the kick frets.

This is technically a slight kludge since it uses a magic `0` rather than properly getting the `Kick` value from the proper enum. As of now, we only support two enums (4L and 5L) when using kick frets and they both have kick at `0`, so I don't think it's worth passing in some extra argument just so that we can know which enum we're supposed to check to ultimately arrive at `0` either way.


wyrdough:
Add a comment noting the inconsistency and I'll merge this

Frickitickitavi:
> Add a comment noting the inconsistency and I'll merge this

Done

PR #1362, opened 2026-03-02
Un-literalize posterize, give mirror its colors

Mekkabit:
Honestly been meaning to get around to this for a *while* and just kind of forgot until a couple days ago, whoops- but this brings posterize to, despite its name, its intended black-crushing only and as a nice bonus, trails aren't immediately cut off. Mirror also gets its color changes rather than being JUST the mirroring effect.


wyrdough (approved):
Approved pending testing

PR #1365, opened 2026-03-04
Fix lyric bar behaviour in replays when scrubbing

Supersonicboss1:
I ended up mostly just switching to using a list, which means that we don't need to remove and rebuild the data when scrubbing, we can just move to any arbitrary time in the song, and all the data should be updated to match.

### DIFF
--- a/Assets/Script/Gameplay/HUD/LyricBar.cs
+++ b/Assets/Script/Gameplay/HUD/LyricBar.cs
@@ -38,7 +38,8 @@ namespace YARG.Gameplay.HUD
         private const double MAX_TRANSITION_DURATION   = 0.3;
         private const double FADE_DURATION             = 0.5;
 
-        private bool ShouldEnable => !(GameManager.IsPractice || SettingsManager.Settings.LyricDisplay.Value == LyricDisplayMode.Disabled);
+        private bool ShouldEnable =>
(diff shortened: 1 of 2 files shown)


Assets/Script/Gameplay/HUD/LyricBar.cs
```diff
@@ -92,39 +93,31 @@ protected override void OnChartLoaded(SongChart chart)
...
-                {
-                    continue;
-                }
-
                 var phraseData = new LyricBarPhrase.PhraseTransitionData
                 {
                     Phrase = currentPhrase,
                 };
 
                 // Can't use i == 0 since we may not be building from the start
```
polson: This comment says don't do this, are you sure it's ok to do?

Supersonicboss1: oh man, thought i removed that comment. Yeah it's fine, this code should only be called during init now so we don't need to worry about where we're starting from.

Supersonicboss1: Okay, removed the comment and also renamed the function to not use the word "queue" since it isn't a queue anymore

wyrdough (approved):
Thanks for addressing this so quickly!

PR #1367, opened 2026-03-05
Increase Size of Play Button

Dansla116:
In `Sidebar`:
- Removed duplicate Favorite button
- Expanded height/width of Play button
- Made tap text (e.g. "Play Song") 50% larger than hold text (e.g. "(Hold) Add to Setlist")
- Matched border-radius of play button to album-art
- Fine-tuned position/alignment
  - Button is middle aligned between album art and source box
  - Line spacing is smaller *(in other words: text is vertically closer)*
  - Text is *middle aligned* within button.  *(note: with the font we use, parenthesis hang lower than all capital letters, so I actually offset all the text slightly higher)*
  - Distance from the button border to the green icon is now the same as the distance from the green icon and the left side of the text
  
**The changes detailed above and the "After" picture below are the latest pushed version.**
  
Before:
image

After:
image


polson (approved):

polson (approved):

polson:
Thoughts on making these the same width? 
CleanShot 2026-03-09 at 10 01 23

Dansla116:
They **are** the same width, right?  At least, I matched up the padding here:

image

And when I photoshop in lines, it's right.  So maybe it's either an issue with different border radius (aka rounded corners) or maybe the different colors are playing tricks on the eyes?
image

I know we've gotten some pushback on rounded corners.  So maybe reducing those and trying to get them the same would help?

polson:
My eyes deceive me, something about optical illusions going on here.  Thanks for checking

PR #1369, opened 2026-03-06
feat(sacn): configurable FPS, pulse duration and fixes

Nort1346:
Improves compatibility with DMX interfaces and software. The original code sent trigger channels for only a single frame (~23ms), too short for most hardware to catch.

`DMXTargetFPS` - configurable sACN update rate, defaults to 44 Hz

`DMXPulseDuration` - how long trigger channels stay on, defaults to 60ms (at least 2 DMX frames). Set to 0 for original single-frame behavior

`InvokeRepeating` -> `Update` - avoids timing drift

Fixed `KillSacn` - `OnChannelSet` was never unsubscribed


polson (approved):
Code looks good

PR #1370, opened 2026-03-07
Migrate venue pp to RenderGraph

theli-ua:


Assets/Art/Shaders/UberPP.shader
```diff
@@ -325,14 +327,14 @@ Shader "Artificial Artists/Universal Render Pipeline/AA_UberPost"
...
-            return half4(color, alpha);
+            return YargVenuePP(color, uvDistorted, uv);
 
-            #if _ENABLE_ALPHA_OUTPUT
-            // Saturate is necessary to avoid issues when additive blending pushes the alpha over 1.
-            return half4(color, saturate(inputColor.a));
-            #else
-            return half4(color, 1);
-            #endif
+            // #if _ENABLE_ALPHA_OUTPUT
```
polson, line 332: Uncomment or remove?

polson (approved):
One comment, but works great, nice job

Assets/Art/Shaders/UberPP.shader
theli-ua: Well, the return statement is right above this actualy. I commented so that it would be more clear that this code isn't being used. but left it there to see easier where we add our changes in the uberpp if we need to update it

PR #1373, opened 2026-03-08
Slightly enhance trails

Mekkabit:
It's kind of stuck in between pure frame blending and what I was actually going for, but it makes the look of trails a *little* more distinct at least?


Mekkabit:
I forgot about the PR that's gonna merge the post processing, oops- I'll just hold off to add it to that

PR #1376, opened 2026-03-09
support Univrm Mtoon shader outlines

4dispel:
just added the feature to the high renderer, i don't know why the diff is so noisy, hope it didn't break stuff.
added it to only the high renderer, because I'm not sure what the difference between low and high is and adding it to that one seemed to work for me.

### DIFF
--- a/Assets/Settings/High_Renderer.asset
+++ b/Assets/Settings/High_Renderer.asset
@@ -13,8 +13,8 @@ MonoBehaviour:
   m_Name: ScreenSpaceAmbientOcclusion
   m_EditorClassIdentifier: 
   m_Active: 1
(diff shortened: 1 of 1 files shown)


4dispel:
for vrm characters that were previously exported as a yargchar this will work without re-exporting if they had the outline settings in the material

polson (approved):

wyrdough:
We should probably verify there is no (significant) performance impact to this before merging. I would expect that the render pass just wouldn't do anything if there are no outline shader materials to render and so would only have a minimal effect when not using a character that uses the material, but I don't actually *know* that.

polson:
Testing outdoor conecrt on my m1 macbook pro, FPS looks about the same on dev vs. this branch.  Didn't test any special characters though

PR #1377, opened 2026-03-09
Fix ListView align to middle in Music Library and History menus + History menu fixes

poliwhirlge:
This PR brings back the previous behavior of the Music Library where as you scroll the selected ListView stays vertically in the middle of the menu.

It also fixes the History menu, which has been just broken in general, especially on 4K:
https://github.com/user-attachments/assets/4b4a7fec-c7b9-452c-9774-098b005135ea

I also snuck in some tweaks to the History menu to bring it closer to the style used by the Music Library (monospace on the score display and color tweaks) but this is just temporary to get it to an okay state while I plan other improvements.


polson:
Can't really tell what this did without before and after screenshots, but looks fine to me I guess

polson (approved):

poliwhirlge:
lol yeah, here you go:

before:
image

after:
image

but mostly it's that the selected item stays in the middle, and also doesn't freak out like it does in the video in the initial post

after:
https://github.com/user-attachments/assets/efb89110-720e-462f-9a56-b05a8872ce63

PR #1378, opened 2026-03-09
Misc lyric bar fixes

Supersonicboss1:
This fixes the lyric bar behavior when there is less than 3 but more than 0 phrases. For reference this is what happened before:
image

I also changed the lyric text prefab to automatically change font size based on width, set vertical alignment to center to prevent it from looking weird when shrunk, and reduced the horizontal margins slightly to prevent the text from going off of the edge of the default background. Example:

https://github.com/user-attachments/assets/53446ff6-3e6f-467b-b202-df80f3f57fa0


polson (approved):
Code looks good

PR #1379, opened 2026-03-09
Make trails less full-screen-smear-y

Mekkabit:
Nooww I get to do this one since it won't be immediately overwritten by the pp merge, hooray!

### DIFF
--- a/Assets/Art/Shaders/VenueShaders/VenuePP.hlsl
+++ b/Assets/Art/Shaders/VenueShaders/VenuePP.hlsl
@@ -133,7 +133,9 @@ half4 YargVenuePP(half3 col, float2 uvDistorted, float2 uv)
     if(_YargTrailLength > 0)
     {
         half3 prevCol = SAMPLE_TEXTURE2D(_YargPrevFrame, sampler_LinearClamp, uv).rgb;
-        col = col + (1.0 - _YargTrailLength) * prevCol;
+        float luma = dot(prevCol, float3(0.2126, 0.7152, 0.0722));
+        float mask = step(0.15, luma);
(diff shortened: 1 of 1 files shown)


Assets/Art/Shaders/VenueShaders/VenuePP.hlsl
```diff
@@ -133,7 +133,9 @@ half4 YargVenuePP(half3 col, float2 uvDistorted, float2 uv)
     if(_YargTrailLength > 0)
     {
         half3 prevCol = SAMPLE_TEXTURE2D(_YargPrevFrame, sampler_LinearClamp, uv).rgb;
-        col = col + (1.0 - _YargTrailLength) * prevCol;
+        float luma = dot(prevCol, float3(0.2126, 0.7152, 0.0722));
+        float mask = step(0.15, luma);
+        col = max(col.rgb, prevCol.rgb * (mask - _YargTrailLength / 2));
```
theli-ua: I know I had `.rgb` in my example (and existing sampling line above uses it too) but they are not necessary, since at some point I decided that all venue PP is done without doing anything to alpha. and those colors are already only having 3 components (they are `half3`)

It also probably makes little sense to use `float` here since everything else already uses half precision floats. (so change those from `float` to `half )

theli-ua (changes requested):

Assets/Art/Shaders/VenueShaders/VenuePP.hlsl
Mekkabit: Oh yeah alright I would've never even questioned any of that-

theli-ua (approved):

PR #1384, opened 2026-03-11
Alternate hands on drum animations

polson:
The drummer now uses both hands for rolls and quick hits on the same drum.  This applies only to auto generated drum animations.

- We disable alternating hands on blue cymbal as this causes YARG drummer to twist excessively, which looks funny
- Repeated slow snare hits use right hand instead of default left which looks more natural
- Green cymbal is randomly mapped to Crash1 or Crash2 to make use of both crash cymbals

Video:

https://github.com/user-attachments/assets/c330e5d0-84c1-48ab-83e6-738498ed18e5

### DIFF
--- a/Assets/Script/Venue/Characters/CharacterManager.cs
+++ b/Assets/Script/Venue/Characters/CharacterManager.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using Discord;
 using UnityEngine;
 using YARG.Core;
 using YARG.Core.Chart;
--- a/Assets/Script/Venue/Characters/CharacterManager.cs
+++ b/Assets/Script/Venue/Characters/CharacterManager.cs
@@ -483,7 +483,7 @@ namespace YARG.Venue.Characters
                 if (!character.IsAnimating())
                 {
                     character.StartAnimation(_currentTempo.SecondsPerBeat);
-                    return;
+                    break;
                 }
 
                 // If next note is more than secondsPerBeat away, stop animating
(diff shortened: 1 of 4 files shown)


Assets/Script/Venue/Characters/CharacterManager.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using Discord;
```
wyrdough: ¿Que?

polson: lmao no idea

Assets/Script/Venue/Characters/VenueCharacter.AnimationLookups.cs
```diff
@@ -640,7 +640,7 @@ private AnimationStateType GetDrumAnimationForNote(DrumNote child)
                 FourLaneDrumPad.Kick => AnimationStateType.Kick,
                 FourLaneDrumPad.YellowCymbal => AnimationStateType.HihatRightHand,
                 FourLaneDrumPad.BlueCymbal => AnimationStateType.RideRh,
-                FourLaneDrumPad.GreenCymbal => AnimationStateType.Crash1RhHard,
```
polson, line 643: This should be removed whoops

wyrdough:
All looks reasonable to me, not that you need my approval. I was just interested in how you'd implemented it, so I figured I may as well take a look ;)

polson:
> All looks reasonable to me, not that you need my approval. I was just interested in how you'd implemented it, so I figured I may as well take a look ;)

Thanks for reviewing, gotta clean up a few things and also the commit history looks a little borked

PR #1385, opened 2026-03-12
Display instruments played for each entry in history menu

poliwhirlge:
Displays the instruments played for each game on the history menu:
image

Had to make one extra db call to get all scores (and one linq to sort them) but it should be okay since it's just once per menu load?

### DIFF
--- a/Assets/Script/Menu/History/HistoryView.cs
+++ b/Assets/Script/Menu/History/HistoryView.cs
@@ -1,8 +1,13 @@
-﻿using Cysharp.Text;
+﻿using System.Collections.Generic;
+using Cysharp.Text;
 using TMPro;
 using UnityEngine;
+using UnityEngine.AddressableAssets;
 using UnityEngine.UI;
+using YARG.Core;
+using YARG.Helpers.Extensions;
 using YARG.Menu.ListMenu;
+using YARG.Scores;
 
 namespace YARG.Menu.History
 {
@@ -19,6 +24,12 @@ namespace YARG.Menu.History
         private TextMeshProUGUI _bandScore;
         [SerializeField]
         private StarView _starView;
+        [SerializeField]
+        private GameObject _instrumentsContainer;
+        [SerializeField]
+        private Image[] _instrumentIcons;
+        [SerializeField]
+        private TextMeshProUGUI _additionalInstrumentsText;
 
         public void OnClick()
         {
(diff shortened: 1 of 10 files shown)


Assets/Script/Scores/ScoreDatabase.cs
```diff
@@ -227,6 +227,15 @@ ORDER BY Date DESC"
             );
         }
 
+        public List<PlayerScoreRecord> QueryAllPlayerScoreRecords()
```
polson, line 230: I kind of feel like this should be a more focused query.  This is what claude gave me, might be an easy win?

```
SELECT * FROM PlayerScores 
          WHERE GameRecordId IN (
              SELECT Id FROM GameRecords 
              ORDER BY Date DESC
          )
          ORDER BY GameRecordId, Id;
```

Otherwise this might pull in a ton of data.  Which might not really matter but if it's easy enough to do your filtering in the query, then we might as well.

Assets/Script/Menu/History/HistoryView.cs
```diff
@@ -50,11 +61,63 @@ public override void Show(bool selected, ViewType viewType)
...
             else
             {
                 _scoreContainer.SetActive(false);
+                _instrumentsContainer.SetActive(false);
             }
         }
+
+        private void PopulatePlayerInstrumentIcons(List<PlayerScoreRecord> playerScoreRecords)
+        {
+            // _instrumentIcons[0] is used for nPlayers > 5 and shares a spot with [1]
```
polson: What is instrumentIcons[0]?  Might be nice to have a comment explaining what this is doing because I don't really get it

Assets/Script/Scores/ScoreDatabase.cs
poliwhirlge: Unless the user has manually messed with the scores.db database by deleting rows in GameRecords, that query will also pull all the records just as mine does, there's no filtering done if we select all players from all plays.

As it is now, the replay menu pulls every game record from the beginning of time and displays them - which we probably should do something about, maybe by capping the number of replays we load (e.g. last 5000 games played) which is easy to do but that seems like a separate PR.

polson: Ok if you're not worried about the query then I'm not either :)

PR #1386, opened 2026-03-18
Update Settings content to remembered tab

Dansla116:
As of 2/17 in `Nightly`, there was an enhancement to remember what tab you were on in the Settings Menu.  This was necessary for navigating to the "Songs" tab specifically from the Music Library Menu when a song required a rescan.
However, this introduced a bug.  When leaving and returning to the Settings Menu, the tab was remembered, but the content would still show "General" *(see image)*, leaving users in a mixed state where they couldn't go to the tab they wanted unless they left first.  In other words, you couldn't click the tab you were *on* but not *on*.

image

Resolves threads:
https://discord.com/channels/1086048856678084609/1483461594070519919
https://discord.com/channels/1086048856678084609/1483562050197127168 *(duplicate)*

### DIFF
--- a/Assets/Script/Menu/Settings/SettingsMenu.cs
+++ b/Assets/Script/Menu/Settings/SettingsMenu.cs
@@ -134,9 +134,20 @@ namespace YARG.Menu.Settings
 
             if (CurrentTab == null)
             {
-                CurrentTab = SettingsManager.DisplayedSettingsTabs[0];
-                _searchBarContainer.SetActive(false);
-                Refresh();
+                var tabId = !string.IsNullOrEmpty(_pendingTabName)
+                    ? _pendingTabName
+                    : _headerTabs.SelectedTabId;
+
+                if (!string.IsNullOrEmpty(tabId))
+                {
+                    SelectTabByName(tabId);
+                }
+                else
+                {
+                    CurrentTab = SettingsManager.DisplayedSettingsTabs[0];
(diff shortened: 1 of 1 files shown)


Assets/Script/Menu/Settings/SettingsMenu.cs
```diff
@@ -134,9 +134,20 @@ private void OnEnable()
...
-                CurrentTab = SettingsManager.DisplayedSettingsTabs[0];
-                _searchBarContainer.SetActive(false);
-                Refresh();
+                var tabId = !string.IsNullOrEmpty(_pendingTabName)
+                    ? _pendingTabName
+                    : _headerTabs.SelectedTabId;
+
+                if (!string.IsNullOrEmpty(tabId))
+                {
+                    SelectTabByName(tabId);
```
wyrdough, line 143: Not at my computer now, but I assume `SelectTabByName` does a refresh on its own?

Dansla116: Yes, but in an indirect way.  `SelectTabByName()` calls `_headerTabs.SelectTabById()`, which fires `TabChanged`.  That calls `OnTabChanged()`, which invokes `SelectTab()`, and that's where `Refresh()` is.

polson:
Looks fine to me and this bug is super annoying so I'm gonna merge

PR #1392, opened 2026-03-23
Fix venue output alpha when alpha is disabled

theli-ua:


polson (approved):

polson:
Looks good but not sure how to test this

wyrdough (approved):

PR #1394, opened 2026-03-23
Use default keyboard binds when not bound to profile

polson:
The Problem:

- In order to use the keyboard to navigate the menus, you must attach a keyboard to a profile.  This is confusing for new players.  Additionally, if the user disconnects the profile which has the keyboard attached (which is not obvious) then menu binds stop working again, and now it's even more confusing.

This adds default keyboard bindings so the keyboard can always be used to navigate the UI.

If any connected profile has a keyboard device associated with it, the menu binds for that profile will override the default binds entirely.  That is, the default keyboard binds will be disabled completely in favor of the explicit menu binds on the profile.  If the profile has keyboard menu binds, but is disconnected, we use the default binds instead.

The default binds are: 

| Action | Key(s) |
|--------|--------|
| Green | 1, Enter |
| Red | 2, Escape |
| Yellow | 3 |
| Blue | 4 |
| Orange | 5 |
| Start | Space |
| Select | Backspace |
| Up/Down/Left/Right | Arrow keys |

### DIFF
--- a/Assets/Script/Input/Bindings/DefaultKeyboardMenuBindings.cs
+++ b/Assets/Script/Input/Bindings/DefaultKeyboardMenuBindings.cs
@@ -1,0 +1,78 @@
+using System;
+using System.Collections.Generic;
+using UnityEngine.InputSystem;
+using YARG.Core.Input;
+
+namespace YARG.Input
+{
+    public sealed class DefaultKeyboardMenuBindings : IDisposable
+    {
+        private static readonly (MenuAction action, Key key)[] DefaultBindings =
+        {
+            (MenuAction.Green,  Key.Digit1),
+            (MenuAction.Red,    Key.Escape),
+            (MenuAction.Red,    Key.Digit2),
+            (MenuAction.Green,  Key.Enter),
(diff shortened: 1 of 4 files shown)


Assets/Script/Input/Bindings/DefaultKeyboardMenuBindings.cs
```diff
@@ -0,0 +1,78 @@
+using System;
+using System.Collections.Generic;
+using UnityEngine.InputSystem;
+using YARG.Core.Input;
+
+namespace YARG.Input
+{
+    public sealed class DefaultKeyboardMenuBindings : IDisposable
+    {
+        private static readonly (MenuAction action, Key key)[] DefaultBindings =
```
polson, line 11: Need to fix this whoops

Assets/Script/Menu/Navigation/Navigator.cs
```diff
@@ -143,6 +143,11 @@ private void Start()
             UpdateHelpBar().Forget();
         }
 
+        private void OnDestroy()
```
polson: Not sure if we need this but seems like it should be here

PR #1397, opened 2026-03-26
New Results Screen

poliwhirlge:
New results screen:
- More stats added (avg multiplier, # of star power activation and time in star power
- Display intensity of selected parts for each player
- Displays the selected controller type (icon in the background of the header) separately from the selected part player (the icon with the difficulty rings)

Future work:
- Add breakdown by section
- Add detailed engine parameters for custom engines

FC score card:
image

One player:
image

\>4 Players:
image

With non-default engine / modifiers active:
image


Assets/Script/Gameplay/GameManager.cs
```diff
@@ -611,7 +611,8 @@ private bool EndSong()
                 {
                     IsHighScore = player.Score > player.LastHighScore,
                     Player = player.Player,
-                    Stats = player.BaseStats
+                    Stats = player.BaseStats,
+                    AverageMultiplier = (float) player.BaseStats.StarScore / player.BaseEngine.BaseScore
```
polson: Do we need to guard for division by 0 here?

Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
```diff
@@ -139,6 +146,9 @@ private void OnEnable()
 
             _sourceIcon.sprite = SongSources.SourceToIcon(song.Source);
 
+            var cancellationToken = new CancellationTokenSource();
```
polson: This token should probably be stored as a field and canceled/disposed in `onDisable`

Assets/Script/Menu/ScoreScreen/ScoreCards/GuitarScoreCard.cs
```diff
@@ -20,9 +21,14 @@ public override void SetCardContents()
         {
             base.SetCardContents();
 
-            _overstrums.text = WrapWithColor(Stats.Overstrums);
-            _hoposStrummed.text = WrapWithColor(Stats.HoposStrummed);
-            _ghostInputs.text = WrapWithColor(Stats.GhostInputs);
+            // Set background icon
+            _instrumentIcon.sprite = Addressables
+                .LoadAssetAsync<Sprite>($"InstrumentIcons[guitar]")
```
polson: Will this show guitar icon even if we are playing the bass part?

Assets/Script/Menu/MusicLibrary/DifficultyRing.cs
```diff
@@ -167,7 +167,8 @@ private void UpdateIconColor()
                 _instrumentIcon.color = Color.white.WithAlpha(INACTIVE_OPACITY);
                 return;
             }
-            if (_songSearchingField.HasInstrumentFilter(_instrument))
+
+            if (_songSearchingField is not null && _songSearchingField.HasInstrumentFilter(_instrument))
```
polson: We have to use != null on unity objects instead of pattern matching

Assets/Script/Menu/MusicLibrary/DifficultyRing.cs
```diff
@@ -177,6 +178,11 @@ private void UpdateIconColor()
 
         public void OnPointerClick(PointerEventData eventData)
         {
+            if (_songSearchingField is null)
```
polson: Use `== null`

Assets/Script/Menu/ScoreScreen/ScoreCards/ScoreCard.cs
```diff
@@ -159,31 +189,55 @@ public virtual void SetCardContents()
...
 
-            // Set background icon
-            _instrumentIcon.sprite = Addressables
-                .LoadAssetAsync<Sprite>($"InstrumentIcons[{Player.Profile.CurrentInstrument.ToResourceName()}]")
-                .WaitForCompletion();
+            // Set engine preset tag
+            var enginePresetId = Player.EnginePreset.Id;
+            if (enginePresetId == EnginePreset.Default.Id)
+            {
+                _enginePresetTag.SetValues("Default Engine", ColoredPillElement.ColoredPillPreset.Default);
```
polson: Should these strings be localized?

polson (changes requested):

Assets/Script/Menu/ScoreScreen/ScoreCards/GuitarScoreCard.cs
poliwhirlge: Yes, it's supposed to represent the controller/profile type of the player and probably makes more sense if the player is using a guitar to play a keys part and vice versa, but I did hesitate since a guitar controller is also a bass controller - maybe we change the icons according to part if it's guitar/bass and maybe drums/prodrums.

Assets/Script/Gameplay/GameManager.cs
poliwhirlge: It would be a problem if we got this far and have a part with 0 base score lol but added the guard.

Assets/Script/Menu/ScoreScreen/ScoreCards/ScoreCard.cs
poliwhirlge: Added localization keys for those strings, but I guess they should match the prepackaged presets the game ships with, which are currently not localized.
I'm also not super sure about where to put these keys so if you have a better idea we can move them.

poliwhirlge:
Think that's everything, ready for re-review - thanks!

Assets/Script/Menu/ScoreScreen/ScoreCards/ScoreCard.cs
```diff
@@ -116,6 +139,13 @@ public virtual void SetCardContents()
             _instrument.text = Player.Profile.CurrentInstrument.ToLocalizedName();
             _difficulty.text = Player.Profile.CurrentDifficulty.ToDisplayName();
 
+            if (_difficultyRing is not null)
```
polson: We need to use `!= null` here because unity is weird

https://docs.unity3d.com/6000.1/Documentation/Manual/null-reference-exception.html

polson (approved):
One last thing

Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
```diff
@@ -119,7 +120,13 @@ private void OnEnable()
             // Set text
             _songTitle.text = song.Name;
             _artistName.text = song.Artist;
-            _bandScoreNotSavedMessage.gameObject.SetActive(
+
+            var scoreNotSavedText = Localize.Key("Menu.ScoreScreen.BandScoreNotSaved")
+                .Replace("(", string.Empty)
+                .Replace(")", string.Empty);
```
wyrdough: Why not remove the parentheses from the localization string to begin with?

Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
wyrdough, line 135: It doesn't really matter in this particular case, but for future reference you're better off using StringBuilder rather than string concatenation since every string modification implicitly creates a new string, generating garbage.

wyrdough (changes requested):
Couple of thoughts and also this problem surfaced during testing, though I'm not exactly sure where it's coming from:

Image

The stack trace indicates it's happening in `ImageExtensions.LoadAlbumCover`, but it's unclear to me exactly why. As I understand it that message indicates that you're trying to destroy something that's part of a prefab rather than an instance.

poliwhirlge:
> Couple of thoughts and also this problem surfaced during testing, though I'm not exactly sure where it's coming from:
> Image
> 
> The stack trace indicates it's happening in `ImageExtensions.LoadAlbumCover`, but it's unclear to me exactly why. As I understand it that message indicates that you're trying to destroy something that's part of a prefab rather than an instance.

Thanks for the catch. I think what's happening is LoadAlbumCover is mainly used in the music library where a lot of album covers are loaded one after another so it destroys the current texture before loading the next one, but I had a default texture referenced in the prefab so it was trying to delete that. I removed it and this error no longer pops up. Similar to how the pause menu (which also loads album art) is set up.

Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
poliwhirlge: This is because I thought the localization files came from crowdin and didn't want to mess with that lol.

Assets/Script/Menu/ScoreScreen/ScoreScreenMenu.cs
poliwhirlge: Yeah, I think I'll just leave this alone for now, this one is at most once per song, will definitely use it for loops though.

wyrdough (approved):

wyrdough:
I went ahead and merged, but if you want to bring back the default album art like we have on the music library, the way to do it is probably to have the default art in the prefab with the dynamic one drawing over it when it exists. (pretty sure that's what music library does)

PR #1398, opened 2026-03-26
Reduce flicker on gameplay start

polson:
Currently there's a few things that make gameplay look janky on start

1. Gray screen before venue pops in
2. HUD elements flicker

This PR fixes these issues by positioning HUD elements correctly, and also fading in the venue once it is loaded

Before:

https://github.com/user-attachments/assets/e3336532-c2eb-41ea-b014-2991b209b201

After:

https://github.com/user-attachments/assets/3c64fcad-2e8c-4c48-93ba-deb5c94f9d49

### DIFF
--- a/Assets/Script/Gameplay/VenueCameraRenderer.cs
+++ b/Assets/Script/Gameplay/VenueCameraRenderer.cs
@@ -1,5 +1,3 @@
-using System;
-using System.Threading;
 using UnityEngine;
 using UnityEngine.Rendering;
 using UnityEngine.Rendering.RenderGraphModule;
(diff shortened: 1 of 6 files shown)


Assets/Script/Gameplay/VenueCameraRenderer.cs
```diff
@@ -47,6 +43,7 @@ public class VenueCameraRenderer : MonoBehaviour
 
         public static float ActualFPS;
         public static float TargetFPS;
+        public static bool IsRendered { get; private set; }
```
polson, line 46: I don't love using a static here but I couldn't think of any other way to do this without a ton of extra complexity.  IMO its fine.

Assets/Script/Gameplay/VenueCameraRenderer.cs
```diff
@@ -203,7 +200,6 @@ private void Update()
             {
                 RecreateTextures();
                 _needsInitialization = false;
-                // Force a render this frame to avoid flickering when resizing
```
polson, line 206: need to restore comments whoops

PR #1401, opened 2026-03-28
Rework star and band star calculations

Supersonicboss1:
See YARC-Official/YARG.Core#346

Most of the code here is just to deal with the changes in the Core PR, and to fix replay saving when playing with bots, and adjusting the star cutoff multipliers to match the `BaseScore` calculation changes.

### DIFF
--- a/Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
+++ b/Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
@@ -111,8 +111,6 @@ public override bool ShouldUpdateInputsOnResume => true;
         public override float[] StarMultiplierThresholds { get; protected set; } =
             GuitarStarMultiplierThresholds;
 
-        public override int[] StarScoreThresholds { get; protected set; }
-
         public float WhammyFactor { get; private set; }
 
         private int _sustainCount;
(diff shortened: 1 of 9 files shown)


Supersonicboss1:
This also now makes the score screen use the `BaseNoteScore` stat for more accurate average multiplier. It should now be completely correct, with the exception of dynamics on drums, which are not factored into either `BaseScore` or `BaseNoteScore`, but are factored into `CommittedScore` if hit.

Assets/Script/Gameplay/Player/FiveLaneKeysPlayer.cs
```diff
@@ -80,12 +80,12 @@ public int GetLanePosition(FiveFretGuitarFret fret)
 
         private static float[] GuitarStarMultiplierThresholds => new[]
         {
-            0.21f, 0.46f, 0.77f, 1.85f, 3.08f, 4.52f
+            0.06f, 0.12f, 0.2f, 0.47f, 0.78f, 1.15f
```
wyrdough, line 84: It is less obvious to me what these numbers actually *mean*. With the old system, it was clear that the threshold was a required average multiplier. Maybe add a comment somewhere explaining what the numbers correspond to so people can know without digging through the entire calculation.

wyrdough (changes requested):

PR #1402, opened 2026-03-29
Add toggle to reduce flashing lights in venues

polson:
Corresponds to: https://github.com/YARC-Official/YARG.Core/pull/374

This reduces flashing lights, camera cuts, and post processing effects for photosensitive players

Toggle screenshot:
CleanShot 2026-03-29 at 17 13 02@2x

Video:

https://github.com/user-attachments/assets/fc3c61d8-90e8-4221-9103-5564ae01c2de

### DIFF
--- a/Assets/Script/Venue/Lights/LightManager.cs
+++ b/Assets/Script/Venue/Lights/LightManager.cs
@@ -1,9 +1,11 @@
+using System;
 using System.Collections.Generic;
 using UnityEngine;
 using YARG.Core.Chart;
 using YARG.Core.Extensions;
 using YARG.Gameplay;
 using YARG.Playback;
+using YARG.Settings;
 using Random = UnityEngine.Random;
 
 namespace YARG.Venue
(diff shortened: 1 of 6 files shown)


polson:
One question is: do we want to further reduce PP effects to the ones that are less intense?  Or should we just rely on the player turning off PP if they don't like it

Edit: ended up reducing PP and lighting

Assets/Script/Venue/Lights/LightManager.cs
```diff
@@ -147,6 +159,57 @@ protected override void GameplayDestroy()
...
+            if (events.Count == 0)
+            {
+                return events;
+            }
+
+            var filtered = new List<T>();
+            foreach (var e in events)
+            {
+                var hasItems = filtered.Count > 0;
+                var timeSinceLast = e.Time - filtered[^1].Time;
```
wyrdough: The first time through the loop this fails (and thus there are never any more times through the loop) since there aren't yet any items added to `filtered`.

wyrdough (changes requested):
I think if you initialize filtered with the first item in events in LightManager like you did in CameraManager this will be just what the doctor ordered. (well, maybe not quite, lol)

polson:
@wyrdough this is good to go now, cleaned up a bunch of this to be simpler

Assets/Script/Venue/Lights/LightManager.cs
polson: Redid this to be a bit more readable

wyrdough (approved):
Didn't test it, but I see nothing wrong here. I might quibble about exactly what lighting and postprocessing events are being replaced and with what, but that's really just a judgement call and I think your choices are well within the realm of reason.

polson:
I'm totally fine with adjusting how we map pp and lighting events if you think there's a better mapping

PR #1409, opened 2026-04-03
feat(gameplay): sync video rewind and add background fades

Nort1346:
Previously, when rewinding, the video would just sit there on a single frame and wait for the audio and gems to catch up. Now, the video actually rewinds along with everything else. I've also added fade in/out animations to the background so the transitions aren't so abrupt

Additionally I removed the blocking WaitUntil in SongRunner.RewindAndResume to keep things moving smoothly

> [!NOTE]
> It's probably worth looking at YARG.Core (BaseEngine.Generic.cs line 355) to tweak the logic there otherwise it might spam some errors in the logs during the rewind


Assets/Script/Playback/SongRunner.cs
```diff
@@ -748,11 +747,7 @@ public async UniTask<bool> RewindAndResume(double seconds, double? overrideTarge
             SetSongTime(targetRewindTime - (AudioCalibration * SongSpeed), 0);
             Resume();
 
-
-            var waitCanceled = await UniTask.WaitUntil(() => SongTime > targetResumeTime, cancellationToken: token)
```
wyrdough, line 752: My recollection is that this was necessary to deal with pausing again during the rewind/resume cycle...

(The exact details escape me at the moment, sorry)

Nort1346: The reason I removed it is that with the new logic, the video now seeks to the target time simultaneously with the audio.

Previously the video stay paused for that 1 second rewind duration, so we had to 'wait' for the notes to catch up to the video's frame. Now that they stay in sync during the jump. There's no longer a gap to wait for

wyrdough: If you look at the `await` chain, you'll note that `GameManager.Resume()` waits on `GameManager.RewindAndResume()` so that `ResumeCore()` doesn't get called until after the `WaitUntil` has completed and we abort the resume there if the player pauses again before we reach the point in the song where the game was paused. 

This is necessary because we *only* want the visual stuff getting updated until we've reached the time of the original pause.

I suspect what you want to do here is to have the call to `BackgroundManager.SetPaused()` that currently happens in `ResumeCore()` happen earlier in the resume process. The issue is making sure that only happens for video backgrounds. I'd imagine one possibility would be to have `BackgroundManager.SetPaused()` take a 3 state enum or something along those lines so it can do the right thing depending on the phase of the resume process and the type of background being used.

Nort1346: You're right that WaitUntil was important for aborting resume if player pauses again before reaching target time.

However in the new implementation:
1. Video stays paused during the entire rewind phase (0.5s animation)
2. Video seeks to audio position **after** rewind completes via `SetTime()`
3. If player pauses again during rewind, `_rewindSource.Cancel()` aborts the rewind and `ResumeCore()` is never called
4. The WaitUntil was specifically to handle the 1 second gap where video was frozen while audio rewound. That gap no longer exists since video now rewinds simultaneously

wyrdough: The delay in calling ResumeCore also prevents the engines from being updated and inputs queued inappropriately until we reach the time of the original pause, ensuring that as far as the engine is concerned there was never any discontinuity.

Nort1346: Is this a theoretical concern or has there been a specific bug in the past? I have tested thoroughly and can't find any issues. What should I look for?

polson: @Nort1346 We don't want to just gloss over the warnings that get thrown when inputs come in during the rewind lead-in.  That's why the `waitUntil` was there, so we can't send inputs out-of-time during the lead-in.  I don't think you needed to remove it unless I'm missing something

PR #1412, opened 2026-04-04
Move vocals into highway renderer

theli-ua:
This removes need for separate vocals camera and improves fps with vocals+highways and default bg by ~10% as measured on steam deck / vulkan


wyrdough:
looks promising...

A quick test reveals an odd interaction with pro keys tracks, at least playing with 2 track players:

Screenshot 2026-04-04 035331

Interesting that the same effect isn't seen on other instrument types, even when using a longer than default track length. I should also note that it doesn't happen with 4 track players, just in case it's helpful..

theli-ua:
oh, thanks. I haven't considered this interaction at all!!!!! (and probably wouldn't even notice it). Although I'm not sure why would this not always be triggered :thinking:

theli-ua:
I have fixed the highways rendering under the vocals track

wyrdough:
#1471 was merged, so this doesn't need to be open

Issue #1262, opened 2026-01-13
Enahnce drums filtering


mrd77:
Currently, there is no way to filter by **any** drums or **if the drum type that I have is available in the song, but is not the main one.**

You currently have to select a specific type of drums when doing it by the search field (like fourlanedrums:; fivelanedrums:;,...).
If you try to filter by clicking on the instrument icon on the library, it will also only filter by that specific type. Every song will only show one main type of drums, even if others are available in the song. It would be great if one could:

- Filter by if the song has any drums chart at all
- If the song has a specific type of drums' chart available, even if it's not the song's main one and/or
- If the song has your type of drums' chart (depending on your selected profile)

Dansla116:
We'll have some form of this as of the next Nightly when #1413 goes in.  Make sure your controller is set to **MIDI Drumkit** and let me know how it works for your library.

PR #1413, opened 2026-04-04
Aggregate Drums Score/Sort/Filter

Dansla116:
When controller is set to **MIDI Drumkit**, show scores for all drum types and treat them all as one instrument.  Order of preference is Elite > Pro > 5-Lane > 4-Lane.  _(see image 1)_

This resolves #1262.

image

_note:_ This originally included the changes in #1451, #1452, and #1453.  I've split those out as they were only very loosely related.


wyrdough:
All sounds reasonable except that I do have concerns about old hashes being removed from playlists since it's possible they might come back later. Some people do store their libraries on removable disks, after all.

Dansla116:
That's a fair concern and not something I considered.  Dare I suggest to add a toggle?

Dansla116:
This was getting unwieldy and I realized there were like four different things going on.  So I've cherry-picked just the changes related to the _Aggregated Drums_ changes and force-pushed to this branch.  I've started new branches and opened new PRs for the other changes.

Assets/Script/Menu/MusicLibrary/SongSorting.cs
```diff
@@ -474,34 +503,63 @@ private static void SortByInstruments(SongCache cache, SortedSongs sorted)
                 {
                     foreach (var entry in list.Value)
                     {
-                        if (DisallowedByRating(entry.SongRating))
```
polson, line 477: Don't we need this?

Assets/Script/Menu/MusicLibrary/SongSorting.cs
```diff
@@ -474,34 +503,63 @@ private static void SortByInstruments(SongCache cache, SortedSongs sorted)
...
+            {
+                foreach (var entry in list.Value)
+                {
+                    var preferredInstrument = MidiDrumkitHelper.GetPreferredInstrumentForSong(entry);
+                    if (!preferredInstrument.HasValue) continue;
+
+                    int intensity = entry[preferredInstrument.Value].Intensity;
+                    if (intensities == null)
+                    {
+                        lock (sorted.AggregateDrums)
```
polson: Not sure we need this lock here since we aren't doing Parallel.ForEach like in the `SortByInstruments `

polson (changes requested):

Dansla116:
@polson, good catch on both of those.  I'm blaming it on cherry-pick/merge problems.

PR #1417, opened 2026-04-07
support song specific textures in custom characters

4dispel:
moved code around and use renderers from bgInstance instead of bg

this let's shaders on custom characters also use the song specific textures.
motivated by the character i posted  in #venue-artist

I've only tested on windows, and i don't know if this will have any effect on mac shader bundles.

### DIFF
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -200,22 +200,11 @@ namespace YARG.Gameplay
             // Breaks things for other platforms, because Unity
             var bg = (GameObject) await bundle.LoadAssetAsync<GameObject>(
                 BundleBackgroundManager.BACKGROUND_PREFAB_PATH.ToLowerInvariant());
-            var renderers = bg.GetComponentsInChildren<Renderer>(true);
 
             // Load Metal shaders, if necessary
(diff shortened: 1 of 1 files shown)


theli-ua:
I start to think that a better way to approach this would to simply do Shader.SetGlobalTexture once on track load and remove on scene unload. (And r move current material iteration all together)

And as far as texture sizes - simply use album art/icon/video(video could also use screen res). 

Using any smaller sizes for those does not matter at all.

theli-ua:
Forgot to update this as we discussed this on discord - those textures are not global as already used by existing venues. 
Although I still think that for future usage we should rather set them unconditionally and global

4dispel:
the global texture approach makes a lot of sense the textures we set whenever we find them.
for the background specific textures we'd be loosing the option of a fallback set by the venue author, so i think we'll also need 2 global boolean shader parameters for whether an image background or video background was found to keep feature parity with what is currently possible.

4dispel:
setting the textures unconditionally is also not perfect for performance, since the textures are only created when a material uses them, we don't run the calculations for the audiovisualizer for no reason. with your proposed full screen resolution to render the video and no checks we would end up with the video of a song being rendered at full screen resolution to a render texture that is never used.
my pc is a bit too powerful to notice, but i heard from some friends that video backgrounds in venues can have a significant impact on performance, so i'd rather not have that overhead when the venue doesn't actually use it.

PR #1424, opened 2026-04-12
Music Library Polish

poliwhirlge:
Mostly UI polish for the following items:

1. Better sidebar gradient blending so it looks less like poo especially for light album colors
2. Adjust sidebar colors to a closer color to the concept and make sure the blue part of the sidebar persists across the non-songs too
3. Align text for the genre and subgenre in the sidebar by making them the same font size
4. Pill that shows percentage adjusts size based on whether that setting to put a decimal is active or not instead of just being long all the time
5. Align the back button, text, and search box better in the header (and also instrument select)
6. Remove light blue bar on the top (rip)
7. Remove random song ButtonView - it's in the context menu and also not very reachable after the first random song
8. Add icon to recommended songs
9. Add "help text" to the action views
10. There was like half a pixel of the background that bled into the menu on the left side of the "sort by artist" header bar which I covered up
11. Removed select keybind to go to next sorting because its annoying if your guitar has select bound to tilt starpower (most of them) and changed the icon away from the select button
12. Sidebar green button when on a CategoryView no longer fills up when held (it has no held action)

After:
image
image
image

Before:
image
image


Assets/Art/Menu/MusicLibrary/SidebarAlbumGradient.png
wyrdough, line 1: Can you rename this to not have spaces? Spaces can be a pain on some OSes

Assets/Script/Menu/Navigation/NavigationScheme.cs
```diff
@@ -38,31 +38,22 @@ public readonly struct Entry
...
-            public Entry(MenuAction action, string localizationKey, Action<NavigationContext> handler,
-                float holdSeconds, Action<NavigationContext> onHoldHandler,
-                Action<NavigationContext> onHoldOffHandler = null, bool hide = false)
+            public Entry(MenuAction action,
+                string localizationKey,
+                Action handler,
+                Action onHoldOffHandler = null,
+                Action onHoldHandler = null,
+                float holdSeconds = 0,
+                bool hide = false) : this(action, localizationKey, _ => handler?.Invoke(), _ => onHoldHandler?.Invoke(), _ => onHoldOffHandler?.Invoke(), holdSeconds, hide)
```
wyrdough: I think by changing this to use lambdas we may be inadvertently creating multiple copies of the class (to capture `onHoldHandler` and `onHoldOffHandler`) rather than just one as the previous version did. Was there a particular reason to change the way this constructor signature works?

My thinking here is that while the previous version did still use lambdas to mutate `Action` into `Action<NavigationContext>` it was not then calling another version of the constructor that needs a closure to keep a reference to the handler actions.

wyrdough (changes requested):
Couple of minor tweaks and I think we'll be good...

Assets/Script/Menu/Navigation/NavigationScheme.cs
poliwhirlge: The main thing I wanted to do here was to fold the third constructor back into the second because it was getting messy and some room for errors down the line one was like (... Action onHoldOff) vs (.. float onHoldSeconds, Action onHold) and if you forgot to specify the seconds it would be interpreted as calling the other constructor, now there's only one ordering of parameters across the two constructors and it's consistent.

Happy to defer to you about the other issue.

poliwhirlge: oh and the constructor chaining had a bug in it so thanks for making me look again lol

PR #1426, opened 2026-04-13
Reduce memory allocations in a number of places

jstnlef:
I did a bit of frame profiling in Unity and saw that there were some low hanging fruit with respect to memory allocations. Fewer allocations means less GC pressure and hopefully fewer GC collection cycles. I was seeing around 80 allocations per frame at 3.9KB on my test song. Now I'm seeing something closer to 23 allocations per frame at 1.4KB

### DIFF
--- a/Assets/Script/Gameplay/HUD/VocalsPlayerHUD.cs
+++ b/Assets/Script/Gameplay/HUD/VocalsPlayerHUD.cs
@@ -36,11 +36,11 @@ namespace YARG.Gameplay.HUD
 
         private bool _shouldPulse;
         private bool _hudShowing = true;
-        private int _displayedMultiplier = int.MinValue;
+        private TextMeshProUGUI[] _textCache;
 
         public void Initialize(EnginePreset enginePreset)
         {
-            _displayedMultiplier = int.MinValue;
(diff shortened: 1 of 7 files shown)


Assets/Script/Gameplay/HUD/VocalsPlayerHUD.cs
```diff
@@ -91,13 +94,18 @@ public void UpdateInfo(float phrasePercent, int multiplier,
...
+            if (_displayedMultiplier != multiplier)
             {
-                _multiplierText.SetTextFormat("{0}<sub>x</sub>", multiplier);
-            }
-            else
-            {
-                _multiplierText.text = string.Empty;
+                _displayedMultiplier = multiplier;
+
+                if (multiplier != 1)
```
theli-ua: For the regular highway player I did it differently to avoid calling SetTextFormat each frame - I simply pre-created 8 TMP objects and I would just made sure only the correct one is visible in a given frame

jstnlef: I like it. I'll put it in. 👍

jstnlef: Made the change

wyrdough (approved):

PR #1427, opened 2026-04-13
Avoid unnecessary calling GetObjectsOfType in `Start` method

theli-ua:
It is expensive. And Start can be called multiple times if the object is being disabled and re-enabled in which case this causes stutter on low performance devices.
Move those into Awake instead.


theli-ua:
Actually, no. it will be called if it is enabled, but only once. I've been misinformed on this one. The doc clearly states that it should only be called once in a lifetime. 

So where are those damn FindObjectsByType calls are coming from then :thinking:

PR #1418, opened 2026-04-08
Refactor layering

theli-ua:
This incorporates #1412 and supersedes it.

highlights:

* Move vocals into highway renderer, this brings about 10% perf increase when both highway and vocals track are visible (on steam deck)
* Fix particle effects not showing unless above the track
* Remove all the UI full-screen rawimage layering and dimming
  * highways render directly into the backbuffer
  * venue is blitted into backbuffer as a render pass, while applying the dimming
  * ui is rendered within highway's "render single camera" stack, instead of separate one
* Track/Graphics preview now applies background dimming (so you get instant feedback on how opaque it is)

The UI change actually barely has any perf impact. (it does improve FPS by couple frames more /second but since I haven't done proper extensive measurement it could be chalked to an error)

PR #1429, opened 2026-04-13
Revert "Re-enable HDR"

wyrdough:
Reverts YARC-Official/YARG#1425 (perf regression)


Ape:
How bad was the perf hit?

wyrdough:
20260413capframex

wyrdough:
Not necessarily from this PR alone, btw, #1418 was also involved

PR #1435, opened 2026-04-15
Remove more memory allocations

jstnlef:
I think this is the remaining low hanging fruit w.r.t. needless allocation every frame. I'm only seeing around 100 bytes a frame now for the "typical frame" in the VenueCameraRender which I believe is happening inside of Unity library code. It's a far cry from the 4k per frame we were seeing originally. Unfortunately, there are still burst-y allocations as songs play but those can be tackled in a future PR.

### DIFF
--- a/Assets/Script/Gameplay/HUD/ScoreBox/ScoreBox.cs
+++ b/Assets/Script/Gameplay/HUD/ScoreBox/ScoreBox.cs
@@ -79,6 +79,8 @@ namespace YARG.Gameplay.HUD
         private bool _easterEggTriggered;
         private bool _vocalsOnly;
         private bool _singlePlayer;
+        private int _displayedCountUpSeconds = -1;
+        private int _displayedCountDownSeconds = -1;
 
         private Tween _multiplierShowTweener;
 
@@ -104,6 +106,8 @@ namespace YARG.Gameplay.HUD
             _scoreText.text = SCORE_PREFIX + "0";
             _bandComboText.text = SCORE_PREFIX + "0";
             _songTimer.text = string.Empty;
+            _displayedCountUpSeconds = -1;
+            _displayedCountDownSeconds = -1;
 
             _songProgressBar.SetProgress(0f);
         }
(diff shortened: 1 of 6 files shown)


Assets/Script/Gameplay/HUD/ScoreBox/ScoreBox.cs
```diff
@@ -145,7 +147,7 @@ private void Update()
             if (GameManager.BandScore != _bandScore)
             {
                 _bandScore = GameManager.BandScore;
-                _scoreText.SetTextFormat("{0}{1:N0}", SCORE_PREFIX, _bandScore);
+                _scoreText.SetText(SCORE_TEXT_FORMAT, _bandScore);
```
wyrdough: I understood `SetText` (and our extension `SetTextFormat`) to not be garbage-generating since they modify the backing array rather than allocating a new string.

Apparently TMP does generate garbage in editor when doing this because it creates a new string for the inspector, but that shouldn't happen in a build.

wyrdough (commented):
Just one small question/comment and I'm happy with this (pending testing, of course)

wyrdough (commented):
Just one small question/comment and I'm happy with this (pending testing, of course)

wyrdough (commented):
Just one small question/comment and I'm happy with this (pending testing, of course)

wyrdough (commented):
Just one small question/comment and I'm happy with this (pending testing, of course)

Assets/Script/Gameplay/HUD/ScoreBox/ScoreBox.cs
jstnlef: I've reverted this change. You were right.

Assets/Script/Gameplay/HUD/SoloBox.cs
```diff
@@ -54,6 +57,9 @@ public void StartSolo(SoloSection solo)
 
             _solo = solo;
             _soloEnded = false;
+            _displayedHitPercent = int.MinValue;
+            _displayedNotesHit = int.MinValue;
+            _displayedNoteCount = int.MinValue;
```
wyrdough, line 62: It only just now occurs to me that note *count* should never change, but I don't think it's a big enough deal to hold this up...

wyrdough: Actually, scratch that, it would change between solo sections, just not within a solo section...sometimes the obvious escapes me until slightly too late

wyrdough (approved):

PR #1436, opened 2026-04-15
Optionally control venues with animators (and add bass fretting)

Mekkabit:
This allows venue authors to do things 'the hard way' and hand over everything I could reasonably think of save for character animations over to animators rather than somewhat hardcoded and limited behavior, including many things otherwise entirely unavailable to venues like live gameplay stats (crowd states/happiness) and all expert chart notes. All this opens up the possibility for completely customizable lighting presets with unique animations per cue, in-game chart visualization (could potentially create an animusic-like venue that plays the charts), moving the band around the stage when camera angles change, changing crowd animations to reflect players' performance and the song's mood, and..whatever someone might think of to do with knowing what post processing cue is currently active, that one kind of stumps me but it's there anyway! ..Hopefully without being any real dent to performance while active or leaving garbage in memory, if those are an issue I won't have the slightest idea how to change that.

Also, the venue character script's hand map handling was ignoring the bassist, so I fixed that.


Assets/Script/Venue/Animator/VenueAnimator.cs
```diff
@@ -0,0 +1,1403 @@
...
+					{
+						_harmony2NoteList.Add(h2phraseNote);
+					}
+				}
+			}
+		}
+
+		private void Update()
+		{
+			if (_lightingEnable && _lightingAnimator == null || _postProcessingEnable && _postProcessingAnimator == null
```
wyrdough: Also note to self: There has got to be a better way to do this...

PR #1437, opened 2026-04-16
Fix particles disappearing unless above the highway

theli-ua:
By including unlit objects in the alpha mask calculation


wyrdough:
Can you correct this in light of #1435?

theli-ua:
> Can you correct this in light of #1435?

rebased

PR #1438, opened 2026-04-16
Revert drum BREs to single-zone, move visual BRE logic out of Core

Frickitickitavi:
[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/360)

Reverts drums to their previous behavior where they functionally have just one big scoring zone. This encourages players to play something musical, since drum BREs actually produce sound - it's also what RB does, I suspect for the same reason.

This also takes the visual "last hit" logic entirely out of YARG.Core and lets each Player class handle it. Each instrument can define arbitrary sets of scoring zones (just one for drums, five for guitar/5LK, three or four for PK depending on the current range), and can then arbitrarily map actions to those zones.

BRE visuals now prioritize visual feedback on the player's actions rather than trying to convey 1:1 information about the scoring system. This means that each drum hit will always light up only its own lane, even in split view, because there are no separate scoring zones to notate anyway. 5LK also separately flashes the Open and Green lanes if the dedicated open lane is enabled, despite them sharing a scoring zone under the hood. It does still flash the Green lane for Open inputs when there _isn't_ an open lane, to explain why they're still gaining points from the open inputs.

### DIFF
--- a/Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
+++ b/Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
@@ -41,7 +41,7 @@ namespace YARG.Gameplay.Player
         };
 
         // Record of the most recent time that each BRE lane has been lit up by any of the actions that map to it
-        private static Dictionary<FiveFretGuitarFret, double> _fretToMostRecentTime = new()
+        private Dictionary<FiveFretGuitarFret, double> _fretToMostRecentTime = new()
         {
             { FiveFretGuitarFret.Green,     0 },
             { FiveFretGuitarFret.Red,       0 },
--- a/Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
+++ b/Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
@@ -32,6 +32,25 @@ namespace YARG.Gameplay.Player
 
         public const int LANE_COUNT = 5;
 
+        private static Dictionary<GuitarAction, FiveFretGuitarFret> _actionToFret = new() {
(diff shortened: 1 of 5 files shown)


Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
```diff
@@ -32,6 +32,25 @@ public sealed class FiveFretGuitarPlayer : TrackPlayer<GuitarEngine, GuitarNote>
...
+        private static Dictionary<GuitarAction, FiveFretGuitarFret> _actionToFret = new() {
+            { GuitarAction.Fret1,    FiveFretGuitarFret.Green},
+            { GuitarAction.Fret2,     FiveFretGuitarFret.Red},
+            { GuitarAction.Fret3,   FiveFretGuitarFret.Yellow},
+            { GuitarAction.Fret4,     FiveFretGuitarFret.Blue},
+            { GuitarAction.Fret5,   FiveFretGuitarFret.Orange},
+        };
+
+        // Record of the most recent time that each BRE lane has been lit up by any of the actions that map to it
+        private static Dictionary<FiveFretGuitarFret, double> _fretToMostRecentTime = new()
```
wyrdough: If this is static won't multiple guitar players overwrite each other's state?

wyrdough (commented):
One small question..

Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
Frickitickitavi: ...yup lol. Will fix

PR #1442, opened 2026-04-16
Add ability to fail and resume

Supersonicboss1:
This mostly updates the code used for resuming to allow arbitrary rewinds, fixes up the highway raising and lowering to work more than once per song, and updates the failing code to freeze gameplay immediately to prevent missed notes as much as possible.

I'm not sure about pausing the SongRunner outside of normal pausing? Technically it makes PauseInfo invalid, but I don't think this will cause any issues since the replays are already invalid.

Demo:

https://github.com/user-attachments/assets/b0dadb49-2600-4ed4-b1ba-8c458d04de3e

### DIFF
--- a/Assets/Script/Gameplay/Player/TrackPlayer.cs
+++ b/Assets/Script/Gameplay/Player/TrackPlayer.cs
@@ -480,6 +480,10 @@ namespace YARG.Gameplay.Player
             {
                 _didLowerTrack = true;
                 CameraPositioner.Lower(isSongEnd);
+            } else if (_didLowerTrack && !shouldLowerTrack)
+            {
+                _didLowerTrack = false;
+                CameraPositioner.Raise(false);
             }
         }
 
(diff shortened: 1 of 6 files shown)


Assets/Script/Gameplay/Player/TrackPlayer.cs
```diff
@@ -480,6 +480,10 @@ protected override void UpdateVisuals(double visualTime)
             {
                 _didLowerTrack = true;
                 CameraPositioner.Lower(isSongEnd);
+            } else if (_didLowerTrack && !shouldLowerTrack)
```
wyrdough: Nitpicky formatting thing: else goes on its own line

wyrdough (approved):
I need to do some testing and double check that what I think is true about what SongRunner  does when it is told to pause is actually true, but unless that uncovers some issues I'm good with this.

wyrdough:
Oops, there was *one* tiny change needed lol

wyrdough:
We should probably at some point despawn already missed notes when resuming, but that was a pre-existing issue, so I'll go ahead and merge

PR #1443, opened 2026-04-17
Music Library Content Rating Icons

poliwhirlge:
Compared to my last attempt I just took out the background and made them larger. If we want them to be just flat icons I can do that too.

Examples:

image
image
image


wyrdough:
Hm, I'm not quite sure what I think of this. All the other metadata is gray/white (and flat). Gonna have to get some feedback/sit with this for a minute to see if it grows on me.

poliwhirlge:
Yeah just leave it for now, I'll try flattening some and see how that is

wyrdough:
It seems that the icon sheet isn't being imported as a sprite atlas

PR #1447, opened 2026-04-18
Clean up multiplier text cache implementation and fix vocals bug

Supersonicboss1:
Switched to using a dictionary over a raw array, and moved the methods to their own helper.

I don't know if this even helps performance on vocals since they are TextMeshProUGUI which I think draws directly to the canvas instead of being a mesh, but this at least fixes vocals for the time being.

### DIFF
--- a/Assets/Script/Gameplay/HUD/VocalsPlayerHUD.cs
+++ b/Assets/Script/Gameplay/HUD/VocalsPlayerHUD.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections;
+using System.Collections.Generic;
 using Cysharp.Text;
 using DG.Tweening;
 using TMPro;
@@ -7,6 +8,7 @@ using UnityEngine;
 using UnityEngine.UI;
 using YARG.Core.Game;
 using YARG.Helpers.Extensions;
+using YARG.Helpers.UI;
 using YARG.Localization;
 using YARG.Player;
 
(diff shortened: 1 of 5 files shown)


Assets/Script/Gameplay/HUD/VocalsPlayerHUD.cs
```diff
@@ -34,13 +36,13 @@ public class VocalsPlayerHUD : GameplayBehaviour
 
         private Coroutine _hudCoroutine;
 
-        private bool _shouldPulse;
-        private bool _hudShowing = true;
-        private TextMeshProUGUI[] _textCache;
+        private bool                             _shouldPulse;
+        private bool                             _hudShowing = true;
+        private Dictionary<int, TextMeshProUGUI> _textCache;
```
wyrdough: I think we'd be better off keeping this an array. It'll waste 2 elements (for the nonexistent 5x and 7x), but save the dictionary lookup overhead and memory penalty.

wyrdough (changes requested):

Assets/Script/Gameplay/HUD/VocalsPlayerHUD.cs
Supersonicboss1: Good point, I didn't realise there was any significant penalty for a dictionary, and figured it would make the code easier to understand - fixed

PR #1452, opened 2026-04-21
Remove Dead Playlist Hashes

Dansla116:
When deleting or modifying a song on disc _(and therefore, removing/changing song hash)_, that inaccessible, *dead* hash was still on the playlist file.
For normal users, this resulted in empty spaces in the playlist manager in game.  You could press left/right to re-order your playlist, and it would sometimes look like nothing is happening because you're moving songs around the *dead* hashes that no longer exist.
For power users/developers, the playlist in game will say one number, and the playlist file on disc will include all the *dead* hashes and show a higher number.
~~Made a change so when rescanning your library, it removes the *dead* hashes from existing playlists.~~ (see feedback item below)

_note:_ This change was originally part of #1413.  I've split it out as they were only very loosely related.

### DIFF
--- a/Assets/Script/Playlists/Playlist.cs
+++ b/Assets/Script/Playlists/Playlist.cs
@@ -47,20 +47,14 @@ namespace YARG.Playlists
             if (!ContainsSong(song))
             {
                 SongHashes.Add(song.Hash);
-                if (!Ephemeral)
-                {
-                    PlaylistContainer.SavePlaylist(this);
-                }
+                SaveAfterManualEdit();
             }
         }
 
         public void RemoveSong(SongEntry song)
         {
             SongHashes.Remove(song.Hash);
-            if (!Ephemeral)
-            {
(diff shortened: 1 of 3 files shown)


wyrdough:
I'm going to hold off on this until after 0.15 since I have no good ideas yet...

Dansla116:
**Feedback Item:**
There was some concern about players who use removable storage.  When rescanning, they would lose songs in their playlists, and not get them back when plugging their storage back in.
I'm open to ideas to work around this.  Perhaps lock this behind a toggle?

Made a change so when manually editing a playlist, it removes the *dead* hashes from only that playlist.  Editing can include re-ordering (either with left/right, or using the "Sort by: ..." from the popup menu), adding or removing a song from the playlist.

wyrdough (approved):
I can live with this since it's only purging dead hashes when the playlist is edited somehow. May have to revisit if we get complaints, but isn't that always the case?

PR #1453, opened 2026-04-21
Sidebar State Fixes

Dansla116:
- Fix sidebar _Difficulties Display_ still showing when navigating to `FiltersMenu`.
- Fix sidebar display after rescan from Music Library when header or button highlighted.

_note:_ This change was originally part of https://github.com/YARC-Official/YARG/pull/1413. I've split it out as they were only very loosely related.


wyrdough:
I went ahead and attached a related bugfix because it's easier this way. Hope you don't mind...

PR #1454, opened 2026-04-21
Save and Load Collapsed Headers Between Music Library Visits

Dansla116:
Without this commit, the collapsed headers would get re-expanded every time the Music Library loads.  This saves collapsed headers, and restores them when returning.

Tested scenarios include:
- Instrument/Difficulty Selection _(including changing instrument)_
- Playing a Song
- Main Menu
- Re-scanning Library _(new headers added will always default expanded)_
- Filters Menu, and changing Filters _(but not **Sort**)_

This **does not** save to disc.  Closing and re-opening the game will re-expand all collapsed headers.

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.SortFilter.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.SortFilter.cs
@@ -42,6 +42,7 @@ namespace YARG.Menu.MusicLibrary
         public bool HasSortHeaders { get; private set; }
 
         private SongCategory[] _sortedSongs;
+        private SongCategory[] _collapsedHeaderTargets;
         private SortAttribute _playlistSort = SortAttribute.Name;
         private bool _playlistSortAscending = true;
         private static readonly Dictionary<SortAttribute, HashSet<SongCategory>> _collapsedHeaders = new();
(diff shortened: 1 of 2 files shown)


Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.SortFilter.cs
```diff
@@ -44,11 +44,61 @@ public partial class MusicLibraryMenu
...
         private bool _playlistSortAscending = true;
+        private static readonly Dictionary<SortAttribute, HashSet<string>> _collapsedHeaders = new();
 
         private List<int> _sectionHeaderIndices = new();
         private int _primaryHeaderIndex;
         private int _recommendedHeaderIndex = -1;
 
+        private static string GetCategoryCollapseKey(SongCategory category)
+        {
+            return $"{category.CategoryGroup ?? string.Empty}\u001F{category.Category ?? string.Empty}";
```
wyrdough: This is going to generate a bunch of garbage from all the string concatenation and will fail if there somehow end up being multiple categories with the same `CategoryGroup` and `Category` strings.

Since this state doesn't persist across program launches, I think a better approach might be to set a custom `GetHashCode` on `SongCategory` like `public override int GetHashCode() => HashCode.Combine(Category, CategoryGroup, Songs);` and use that as the key for the `HashSet`.

wyrdough (changes requested):
The change requested here is the minimum required. We should try to avoid allocating strings in loops like that.

I had a different idea that was easier to just write code to explain rather than write it out here, so I PRed your PR branch. Feel free to discuss, of course.

Dansla116:
Thanks for the suggestion.

I cherry-picked your change.  There was an issue with preserving the collapsed headers when changing filters and I fixed that up.  However, changing sorts (and therefore, changing all headers) still puts you in a state where everything is re-expanded.  This is an expected limitation.

I saw your comment about the initialize block and moved that to it's own function to make `Awake()` cleaner.

wyrdough (approved):

PR #1455, opened 2026-04-22
Hook up the calibration scene back button

jstnlef:
I added an onclick handler to the button. It seems like the scene serialized differently due to the change. It was mentioned here https://discord.com/channels/1086048856678084609/1496246906043498707


wyrdough (approved):
Not quite the *correct* behavior, but fixing things so that it returns to the settings menu is a task for another day.

PR #1456, opened 2026-04-23
Back button on Calibrator should bring player back to settings

jstnlef:
This PR does the following:

- Calibrator now owns its navigation scheme more safely: it tracks whether it pushed a scheme and clears it on state changes/destruction.

- Settings button rows now close their focused navigation scheme before invoking button actions, preventing stale nav state.

- Opening the calibrator from settings now sets a one-shot flag so returning to the main menu reopens settings.

- Settings menu tab restore logic now handles the case where the selected header tab is already selected but CurrentTab was cleared. This fixes a bug where the Player could go into settings, leave settings, go back into settings and hit "show advanced" which would end up clearing all the settings on that page until navigating to a new tab

### DIFF
--- a/Assets/Script/Menu/Calibrator/Calibrator.cs
+++ b/Assets/Script/Menu/Calibrator/Calibrator.cs
@@ -51,6 +51,7 @@ namespace YARG.Menu.Calibrator
 #nullable disable
 
         private bool _wasWhammyEnabled;
+        private bool _hasNavigationScheme;
 
         private void Awake()
         {
@@ -65,6 +66,7 @@ namespace YARG.Menu.Calibrator
         private void OnDestroy()
         {
             InputManager.MenuInput -= OnMenuInput;
+            ClearNavigation();
             _mixer?.Dispose();
         }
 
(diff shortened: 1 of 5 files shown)


wyrdough:
Thanks! Not sure if I'll be able to get to this tonight, but definitely appreciate you taking this on!

jstnlef:
Not a problem. Seemed needed.

wyrdough:
It's 3:30 in the morning, so I'm just guessing as to the cause, but with this PR the main menu is shown for a frame when opening the calibrator. I suspect it has something to do with clearing the navigation stack, as navigator has some logic that causes it to wait a frame before updating so as not to upset the music player (which it owns).

jstnlef:
> It's 3:30 in the morning, so I'm just guessing as to the cause, but with this PR the main menu is shown for a frame when opening the calibrator. I suspect it has something to do with clearing the navigation stack, as navigator has some logic that causes it to wait a frame before updating so as not to upset the music player (which it owns).

I made the fix 👍

PR #1457, opened 2026-04-23
Force venue output alpha channel to 1.0 even if post processing is disabled

theli-ua:
This fixes older visualisers that didn't write alpha channel at all

### DIFF
--- a/Assets/Art/Shaders/VenueShaders/VenueAlphaFix.shader
+++ b/Assets/Art/Shaders/VenueShaders/VenueAlphaFix.shader
@@ -1,0 +1,31 @@
+Shader "Hidden/YARG/VenueAlphaFix"
+{
+    HLSLINCLUDE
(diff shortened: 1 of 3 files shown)


wyrdough:
This busted the trails effect. Also it seems like the venue render FPS is always significantly below the target FPS, but I haven't actually checked how much it changed from `dev`.

It does, however, fix the old shaders 100%.

theli-ua:
Trails (and other effects) fixed. I've also fixed the slow fps effect when the cap is set to 0.

theli-ua:
I've also added change to improve fps limiting and get results that are much closer to desired fps

wyrdough:
I'm going to go ahead and merge this since it fixes fixes what it set out to fix and makes the venue FPS target much more accurate. However, one minor problem is that low FPS is a little more janky than it used to be with this, presumably because of an interaction with the FPS smoothing. It has an expectedly low framerate for a second, then renders at a higher framerate for a little while, then pops back to low framerate, repeating.

PR #1459, opened 2026-04-24
Various replay viewer fixes

Supersonicboss1:
- Fixes the BandMultiplier becoming very large on seeking
- Fixes SuccessCount mismanagement on seeking
- Fixes band combo becoming very large sometimes on seeking
- Fixes band combo and band multiplier displays depending on active profiles for behaviour during replays

This also changes the way the band combo multiplier determines units - instead of 1 if non-vocals present, 10 otherwise, it picks the smallest combo units from the available engines, which should do the same thing but work better with future engine types which may have different combo unit values.

I did not fix the star display behavior on seeking backwards during a replay, because I truly have no idea how the Unity animator works, and every time I tried to fix it, it just broke even more somehow.

Requires YARC-Official/YARG.Core#370


Supersonicboss1:
nevermind, I did the stars. This is a more significant change since I essentially just ripped out the animator and animations and replaced them with DOTween Sequences, so if you want me to not include this change, that's totally fine.

Assets/Script/Gameplay/HUD/ScoreBox/ScoreBox.cs
```diff
@@ -77,7 +77,7 @@ public class ScoreBox : GameplayBehaviour
         private string _timeFormat;
 
         private bool _easterEggTriggered;
-        private bool _vocalsOnly;
+        private int _bandComboUnits;
```
wyrdough: You should default this to 1 to avoid the divide by zero when `_bandComboObject` is inactive. (Either that or check that it isn't 0 before calling `_bandComboText.SetTextFormat()` below, but that would be a waste of a branch IMO)

wyrdough (changes requested):
I also did some testing of the star stuff and it seemed to work but I'm going to need to dig into the code a bit more before I sign off on that. It's entirely up to you whether you separate out the star changes to get the rest of it merged more quickly.

Supersonicboss1:
> I also did some testing of the star stuff and it seemed to work but I'm going to need to dig into the code a bit more before I sign off on that. It's entirely up to you whether you separate out the star changes to get the rest of it merged more quickly.

No rush, all of this has been broken at least since unisons and band multipliers were added (v0.13?), and while I would *like* for it to be merged before v0.15, the bugs are pretty insignificant overall, since all of them only occur if you seek backwards in a replay, and nobody other than me seems to have noticed anyway lol. I'll leave it in here since the stars fixes are pretty much impossible to test without the other changes from this PR anyway.

PR #1463, opened 2026-04-26
Vocals visual fixes

jstnlef:
In playing YARG, I always found the vocal lyrics to be a bit blurry. This is especially true for songs with faster lyrics. These should be relatively subtle changes which I believe make a difference.

- Pixel-snaps the vocal camera viewport before assigning Camera.rect, reducing fractional render texture sampling for the vocal track.
- Moves pooled gameplay visual element updates from Update() to LateUpdate() so visual positions are updated after all of the gameplay logic has been updated.
- Applies a vocal-specific TMP _Sharpness value to scrolling and static lyric text material instances to make lyric edges render more crisply.

Feel free to let me know if you don't think these are worth it. At the very least, I think the Update() to LateUpdate() change is worth it since we can ensure that all of the game logic is executed first before trying to move any of the visuals during that frame.

### DIFF
--- a/Assets/Script/Gameplay/Player/Vocals/VocalTrack.cs
+++ b/Assets/Script/Gameplay/Player/Vocals/VocalTrack.cs
@@ -210,19 +210,18 @@ namespace YARG.Gameplay.Player
             var imageAspectRatio = vocalsSize.width / vocalsSize.height;
             var clampedAspectRatio = Math.Clamp(imageAspectRatio, MIN_ASPECT_RATIO, MAX_ASPECT_RATIO);
 
-            float heightPixels = vocalsSize.height;
-            float widthPixels = heightPixels * clampedAspectRatio;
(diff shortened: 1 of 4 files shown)


wyrdough:
I ran a quick test of this and didn't see any issues, but I also didn't really see any difference, sooo...

jstnlef:
> I ran a quick test of this and didn't see any issues, but I also didn't really see any difference, sooo...

It would be pretty subtle regardless. I just generally find the text blurry as it goes by in YARG far more than I ever noticed in Rock Band. What do you think about moving BaseElement to `LateUpdate()` from `Update()`. This guarantees that all the visual elements update after all the game logic updates in the frame.

jstnlef:
Closing this for the more focused change [here](https://github.com/YARC-Official/YARG/pull/1469)

PR #1472, opened 2026-04-30
Add custom theme support

theli-ua:
Same as usual, as asset bundle with metal shader subbundle. 
Themes dropped in custom/themes show up as options in profiles. Can have multiple active at the same time


Assets/Script/Themes/Authoring/ThemeComponent.cs
```diff
@@ -89,8 +97,176 @@ VisualStyle.FiveFretGuitar or
...
+#if UNITY_EDITOR
+        private const string THEME_PREFAB_NAME = "ThemeRoot";
+        private const string THEME_META_NAME = "theme_meta";
+        private const string THEME_PREFAB_PATH = "Assets/" + THEME_PREFAB_NAME + ".prefab";
+        private const string THEME_META_PATH = "Assets/" + THEME_META_NAME + ".asset";
+        private const string BUNDLE_OSX_SUFFIX = "_metal.bytes";
+        private const string BACKGOUND_OSX_MATERIAL_PREFIX = "_metal_";
+
+        [ContextMenu("Export Theme")]
+        public void ExportTheme()
```
theli-ua, line 112: There is code duplication between this and venue/character mostly because venue/character share scene

theli-ua: I've refactored this to not have any duplication

PR #1475, opened 2026-05-01
Fix vocal percussion orientation

theli-ua:


theli-ua:
https://discord.com/channels/1086048856678084609/1499595366335774720

PR #1479, opened 2026-05-03
Fix slow fps effect

theli-ua:
while using actualfps makes sense for trails, for fps it does not as it becomes a feedback loop - lowering targetfps lowers actualfps which lowers targetfps which lowers actualfps which lowers target fps which... hopefully the pattern is clear here.

Instead always target the intended 60/divisor capped to venue fps cap.


theli-ua:
https://discord.com/channels/1086048856678084609/1499748970753425430

Supersonicboss1 (approved):
Code seems correct, and it fixes the issue for me (tested in both the editor and an actual build)

PR #1488, opened 2026-05-13
Add ability to view the score screen for a replay

Supersonicboss1:
Things I have changed in this PR that are not obvious from the title:
- Analyze has been banished to test builds and the editor, and therefore removed from nightly.
- AverageMultiplier is no longer a parameter needed to create the score screen, as it should be present in BaseStats. 
- This uses ReplayAnalyzer to generate more detailed information about the replay. You might be able to get it from the replay directly, but this was simpler and didn't require a super significant rewrite of anything to accomplish. If you don't like this approach, let me know and I'll find another way to do it.
- You can view a replay from the replay score screen. I considered removing the view replay button entirely, and making viewing the score screen the default (green option), and then you press yellow from a replay score screen to view, but I left that unchanged for now.

I should note, this only lets you view the score cards of replays where you could watch them, that is, where the chart is present and the replay version is current. I would like to change this to not require that, but it would require building a subset of both the score screen and score cards which only includes data that is present in replays, and it would have no background image which sucks.

Requires YARC-Official/YARG.Core#380


Assets/Script/Menu/History/ViewTypes/ReplayViewType.cs
```diff
@@ -158,6 +160,58 @@ public override void Shortcut1()
...
+            }
+            var results = ReplayAnalyzer.AnalyzeReplay(chart, _entry, data);
+            var playerScoreCards = new PlayerScoreCard[results.Length];
+            for (int i = 0; i < _entry.Stats.Length; i++)
+            {
+                var playerResult = results[i];
+                var playerEntry = _entry.Stats[i];
+                playerScoreCards[i] = new PlayerScoreCard
+                {
+                    IsHighScore = false, // No way to know this from a replay
```
wyrdough: I am guessing without checking that the ID of the score record isn't saved in the replay so we can't query the score db to figure out whether something was a high score?

wyrdough (commented):
One thing I think merits discussion: Should we refuse to generate score cards for replays that don't pass validation?

Supersonicboss1:
> One thing I think merits discussion: Should we refuse to generate score cards for replays that don't pass validation?

I would say no, or at least not now? Right now, there is a fair number of replays which don't pass validation, but where the score is still reasonably "correct" (e.g. it's just some tiny difference with star power time stuff), and I wouldn't want users to be unable to view the scores of these replays, even if they may be slightly off. Maybe it would be better if we just showed the inconsistent replay results popup (or one similar to it, but specific to viewing the score screen on a replay)?

Assets/Script/Menu/History/ViewTypes/ReplayViewType.cs
Supersonicboss1: Turns out you can do this! To be honest, I totally forgot the score db existed, but it's been implemented now.

poliwhirlge:
late to this but imo don't even need to bring the inconsistent popup again, just add another little pill thing up in the top right with "INCONSISTENT REPLAY" in it where we usually have "SCORE NOT SAVED"

Supersonicboss1:
> late to this but imo don't even need to bring the inconsistent popup again, just add another little pill thing up in the top right with "INCONSISTENT REPLAY" in it where we usually have "SCORE NOT SAVED"

Sorry for taking an entire month to get to this lol, but yeah, that's a better idea, so I've implemented it here.

wyrdough (approved):
YOLO

PR #1494, opened 2026-05-22
Fix vocals scaling 

polson:
This fixes issues with vocals scaling when resizing the window 

We also add back in the fade effect on the sides of the vocals highway

[Kooha-2026-05-22-10-55-47.webm](https://github.com/user-attachments/assets/fdaef4d4-4661-43c0-9314-e275a75efa5d)


theli-ua (approved):
this is fine

PR #1495, opened 2026-05-22
Fix vocal percussion song state

jstnlef:
Fixes vocal percussion visibility and practice-mode note handling.
   •  Hide the percussion fret on initialization so it no longer appears at song start
   •  Trigger the fret show animation when entering a percussion phrase, including when the
      first phrase is percussion.
   •  Hide the percussion fret again, if active, after the final vocal phrase.
   •  Reset percussion visual state during practice-section resets.
   •  Avoid a NullReferenceException when hitting percussion notes after jumping into
      practice mode before the note visual has spawned.

### DIFF
--- a/Assets/Script/Gameplay/Player/VocalsPlayer.cs
+++ b/Assets/Script/Gameplay/Player/VocalsPlayer.cs
@@ -516,8 +516,7 @@ namespace YARG.Gameplay.Player
             while (_phraseIndex < NoteTrack.Notes.Count &&
                 (_phraseIndex == -1
                     ? NoteTrack.Notes.Count > 0 &&
-                        (NoteTrack.Notes[0].Time <= time ||
-                            NoteTrack.Notes[0].ChildNotes.Any(note => note.IsPercussion))
+                        (NoteTrack.Notes[0].Time <= time || HasPercussion(NoteTrack.Notes[0]))
                     : NoteTrack.Notes[_phraseIndex].TimeEnd <= time))
             {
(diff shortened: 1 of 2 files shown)


Assets/Script/Gameplay/Player/VocalsPlayer.cs
```diff
@@ -513,36 +514,44 @@ private void UpdatePercussionPhrase(double time)
             // the end times of phrases instead (where the phrase lines are). Problem
             // with this is that we still gotta account for the first phrase, so use
             // an index of -1 for that.
-            while (_phraseIndex == -1 ||
-                (_phraseIndex < NoteTrack.Notes.Count && NoteTrack.Notes[_phraseIndex].TimeEnd <= time))
+            while (_phraseIndex < NoteTrack.Notes.Count &&
```
polson: This is making my head spin, could we extract some booleans here to make it more readable?  Or at least add comments explaining what we're checking here

polson (approved):
Looks ok to me, just one cleanup thing

Assets/Script/Gameplay/Player/VocalsPlayer.cs
jstnlef: Makes sense. I'll do that when I get home later tonight

jstnlef: I've done the refactoring

polson (approved):

PR #1508, opened 2026-06-08
Load default album texture in venue if none to prevent errors

Supersonicboss1:
Fixes the vibe desk if the song has no album art present.


theli-ua:
Looks fine to me, although I wish you'd minimize scene file changes like these:

```patch
-m_Name: 
+m_Name:
```

theli-ua (approved):

PR #1509, opened 2026-06-08
Hide missed notes on rewind

Supersonicboss1:
Mildly janky, but it does work? Not sure about the whole `INoteElement` thing, but otherwise it does not seem to be able to recognize `NoteElement` class types because of the type parameters, so aside from reflection or duplicating the same method 5 times I don't know of a good alternative?

https://github.com/user-attachments/assets/c2082c78-63e2-4a08-b3d5-ebcc10858613

### DIFF
--- a/Assets/Script/Gameplay/Visuals/TrackElements/NoteElement.cs
+++ b/Assets/Script/Gameplay/Visuals/TrackElements/NoteElement.cs
@@ -45,7 +45,6 @@ namespace YARG.Gameplay.Visuals
 
         private   bool _lastStarPowerState;
         private   bool _lastStarPowerActiveState;
-        private   bool _wasMissed;
         protected bool IsStarPowerVisible => CalcStarPowerVisible();
 
         public abstract void SetThemeModels(ThemeDict models, ThemeDict starPowerModels);
@@ -117,14 +116,13 @@ namespace YARG.Gameplay.Visuals
         /// </summary>
         public virtual void MissNote()
         {
-            _wasMissed = true;
             SustainState = SustainState.Missed;
             OnNoteStateChanged();
         }
 
         public void OnRewind()
         {
-            if (_wasMissed)
+            if (NoteRef.WasMissed)
             {
                 ParentPool.Return(this);
             }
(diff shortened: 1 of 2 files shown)


Assets/Script/Gameplay/Visuals/TrackElements/NoteElement.cs
```diff
@@ -38,9 +43,9 @@ public abstract class NoteElement<TNote, TPlayer> : TrackElement<TPlayer>, IThem
 
         public override double ElementTime => NoteRef.Time;
 
-        private bool _lastStarPowerState;
-        private bool _lastStarPowerActiveState;
-
+        private   bool _lastStarPowerState;
+        private   bool _lastStarPowerActiveState;
+        private   bool _wasMissed;
```
wyrdough: This needs to be reset in `InitializeElement` (assuming it sticks around), otherwise the state will be wrong when the object gets reused from the pool.

wyrdough (changes requested):
In general we want to avoid interfaces in gameplay stuff because resolving an interface method call is slow compared to a direct method call, but since this is just rewind so we're not calling it 20 times a frame or anything I think defining an interface is a perfectly reasonable way of handling this.

The only (slight) concern I have with this is adding the missed state to the `NoteElement` itself rather than getting it from the `NoteRef`, which risks getting the missed state desynchronized from the engine. I can imagine a scenario where someone looking to optimize replay seeking ends up attempting to keep a `NoteElement` alive instead of reinitializing it, for example.

Supersonicboss1:
> In general we want to avoid interfaces in gameplay stuff because resolving an interface method call is slow compared to a direct method call, but since this is just rewind so we're not calling it 20 times a frame or anything I think defining an interface is a perfectly reasonable way of handling this.
> 
> The only (slight) concern I have with this is adding the missed state to the `NoteElement` itself rather than getting it from the `NoteRef`, which risks getting the missed state desynchronized from the engine. I can imagine a scenario where someone looking to optimize replay seeking ends up attempting to keep a `NoteElement` alive instead of reinitializing it, for example.

Turns out there is just a `WasMissed` parameter on `NoteRef` - there was no need to define my own variable here, so there should be no worry about desync.

PR #1524, opened 2026-06-24
Add 'Background FPS Cap' setting

daviditsygin:
Adds a "Background FPS Cap" setting (Graphics → Display, default 10) that caps the framerate while the window is unfocused to lower CPU/GPU load, restoring the configured framerate on refocus. Set to 0 to disable. VSync is disabled while capped so the limit applies, and restored from the live setting on refocus. Independent of Pause on Focus Loss.

### DIFF
--- a/Assets/Scenes/PersistentScene.unity
+++ b/Assets/Scenes/PersistentScene.unity
@@ -2348,6 +2348,7 @@ GameObject:
   - component: {fileID: 788257161}
   - component: {fileID: 788257162}
   - component: {fileID: 788257167}
+  - component: {fileID: 788257168}
   m_Layer: 0
   m_Name: Global Variables
   m_TagString: Untagged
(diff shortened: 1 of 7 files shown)


wyrdough:
Thanks for the PR! It'll be a few days before anything happens on this because we're just about to release 0.15. Speaking of which, if you want to be listed in the in-game credits in 0.15, please open a PR against the [Contributors repo](https://github.com/YARC-Official/Contributors) to add yourself.

wyrdough (changes requested):
If we're going to have an increasing amount of focus-related state, I think I'd prefer to refactor `OnApplicationFocus` into a new `MonoSingleton` named something like `FocusHandler`, just for organizational reasons.

Other than that, this looks good to me!

daviditsygin:
> If we're going to have an increasing amount of focus-related state, I think I'd prefer to refactor `OnApplicationFocus` into a new `MonoSingleton` named something like `FocusHandler`, just for organizational reasons.
> 
> Other than that, this looks good to me!

Updated to match that pattern!

wyrdough (approved):

PR #1527, opened 2026-06-26
add wind parameter to character manager and pause VRM springs during pause

4dispel:
adds a new Vector Setting to the character manager, which sets a wind that will be applied to VRM-character's springbones.
a slight wind could look good on outside venues :)
i made it check for updates to the wind so that animating the wind vector on the character manager can be possible

### DIFF
--- a/Assets/Script/Venue/Characters/CharacterManager.cs
+++ b/Assets/Script/Venue/Characters/CharacterManager.cs
@@ -21,7 +21,8 @@ namespace YARG.Venue.Characters
 
         [SerializeField]
         private Vector3 _wind;
-        private Vector3 _lastUpdatedWind = Vector3.zero;
+        private Vector3 _lastUpdatedWind     = Vector3.zero;
+        private bool    _lastKnownPausedState = false;
 
         private readonly Dictionary<VenueCharacter.CharacterType, VenueCharacter> _characters = new();
         public Dictionary<VenueCharacter.CharacterType, VenueCharacter> Characters => _characters;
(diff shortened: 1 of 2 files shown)


4dispel:
now pauses springbones during gameplay pauses, before they would move very weirdly during pauses.
the way springbone pausing is implemented in UniVRM is just stopping the updates to the transform so they will still move weirdly it just won't be visible. on unpause they teleport to the location they reached during the pause.
also rebased to resolve conflicts with dev

Assets/Script/Venue/Characters/CharacterManager.cs
```diff
@@ -236,7 +241,20 @@ private void Update()
...
+                }
+
+                if (_wind != _lastUpdatedWind && character is VRMCharacter)
+                {
+                    ((VRMCharacter) character).SetWind(_wind);
+                }
             }
+
+            _lastKnownPausedState = GameManager.Paused;
+            _lastUpdatedWind = _wind;
```
wyrdough, line 257: If these go inside their respective `if` blocks, wouldn't we avoid unnecessarily setting them every frame?

wyrdough (changes requested):

Assets/Script/Venue/Characters/CharacterManager.cs
4dispel: they can't go into the if blocks that are already there, since those are iterated over for each character, so changing these before the if blocks for the other characters would make those not update.
they could be put into their own if blocks, but i don't know if an assignment every frame or an extra comparison would be a performance difference.

PR #1535, opened 2026-06-30
Add the ability to correct audio offset for individual songs

asimard1:
Adds a button on the score screen that applies the average hit-timing offset (ScoreScreenStats.AverageOffset, computed from non-bot players) to the song's delay value in song.ini. The button toggles between "Add Song Offset" and "Remove Song Offset" so the change can be undone with a second press. Also adds AverageOffset to ScoreScreenStats, populated as the mean GetAverageOffset() across non-bot players.
Bound to the Select action.


asimard1:
What I'll do next is:
- Load this .json file and actually use the offset as a delay when loading the song so it affects gameplay (this could eventually be hidden behind a setting, but I don't know how to do that yet)
- Move some code to new files, as I don't want to clutter ScoreScreenMenu.cs
Now works for CON and SNG packages as well :)

wyrdough:
I was going to reply to your question from the other PR about where to save the JSON file, but it seems you have it well in hand. Normally things go in a subdirectory, but there's not really any place it obviously fits, so where you have it now is fine.

As for how to add a toggle setting, the general idea is to define the settings in `SettingsManager.Settings.cs`, then expose it to the user in `SettingsManager.cs` where ever in the `DisplayedSettingsTabs` list makes sense for it to live. You can then access the value of the setting with something like `SettingsManager.Settings.YourSettingName.Value`.

I hate adding other stuff that you'll have to figure out, but it would probably be best to make the button require a long press to activate so people won't be as prone to hitting it accidentally. It's not the most helpful pointer since it is large and spread out over several files, but `MusicLibraryMenu` should have an example of how to make a navigation button require a long press in there somewhere.

asimard1:
This still requires just a bit of testing, but if anyone could test this with more than one player it would be great. Appart from that everything should be done. (Tested, works fine except when playing online because of the delay.)

edit: Thank you wyrdough for your help with finding the necessary files, everything was easy to adapt :)

edit: Tested with more than one player and it works perfectly

asimard1:
The last bug that I see is that YargLogger.LogFormatInfo doesn't seem to make anything appear, but that might just be on my side. Anyway, I don't think it came from my code.

Also when doing a setlist, buttons won't fit on screen :(
image

**Also really important** to do when Online gets added to the game (conclusions after testing with the online alpha from pythonology): Only consider local players for the average, or else some horrible delays are considered from the online delay. I added a comment for this and tested it with the onlive dev build available, works perfectly.

asimard1:
I will add the quick option in game to manually calibrate this value. I find that having the option only at the end of songs is not really useful for multiple songs, and I would like to have the ability to solve a song's offset directly at the start so that I can enjoy it fully :)

asimard1:
I confirmed that the quick settings feature works well.

I think the main use of this PR, as discussed on Discord, is that it will greatly help to make perfect some songs which weren't synced properly in a chart editor. From my own testing, most songs have a delay that is quite important, and some have offsets that go up to 200 ms (which are almost impossible to play without this addition). [A few of the offsets I have recorded.](https://github.com/user-attachments/files/30572003/song_offsets.json)

I feel like this will be quite important as well for the leaderboards, especially for  https://github.com/YARC-Official/YARG/pull/1539 if it ever gets merged.

Also, this will work great with the late/early/perfect indicator for hits, since as soon as you start a song you'll see quickly if you always hit late or early, calibrate the song offset directly (if your calibration is already good for other songs), and then you'll hit much better scores.

I hope you consider this feature :D

polson:
This branch has compilation errors that need to be fixed

asimard1:
@polson should be fixed

polson:
Seeing this when I try to run:

Screenshot 2026-08-13 at 12-06-57


Also:

```
[<a href="Assets/Script/Localization/LocalizationManager.cs" line="129">LocalizationManager.cs:TryParseAndLoadLanguage:129</a>] Newtonsoft.Json.JsonReaderException: After parsing a value an unexpected character was encountered: ". Path 'Menu.ScoreScreen.RemoveSongOffset', line 521, position 12.
  at Newtonsoft.Json.JsonTextReader.ParsePostValue (System.Boolean ignoreComments) [0x0019a] in <761cf2a144514d2291a678c334d49e9b>:0 
  at Newtonsoft.Json.JsonTextReader.Read () [0x0005a] in <761cf2a144514d2291a678c334d49e9b>:0 
  at Newtonsoft.Json.Linq.JContainer.ReadContentFrom (Newtonsoft.Json.JsonReader r, Newtonsoft.Json.Linq.JsonLoadSettings settings) [0x001f2] in <761cf2a144514d2291a678c334d49e9b>:0 
  at Newtonsoft.Json.Linq.JContainer.ReadTokenFrom (Newtonsoft.Json.JsonReader reader, Newtonsoft.Json.Linq.JsonLoadSettings options) [0x00030] in <761cf2a144514d2291a678c334d49e9b>:0 
  at Newtonsoft.Json.Linq.JObject.Load (Newtonsoft.Json.JsonReader reader, Newtonsoft.Json.Linq.JsonLoadSettings settings) [0x0006a] in <761cf2a144514d2291a678c334d49e9b>:0 
  at Newtonsoft.Json.Linq.JObject.Parse (System.String json, Newtonsoft.Json.Linq.JsonLoadSettings settings) [0x0000c] in <761cf2a144514d2291a678c334d49e9b>:0 
  at Newtonsoft.Json.Linq.JObject.Parse (System.String json) [0x00000] in <761cf2a144514d2291a678c334d49e9b>:0 
  at YARG.Localization.LocalizationManager.ParseAndLoadLanguage (System.String cultureCode) [0x00047] in /Users/admin/Development/YARG/Assets/Script/Localization/LocalizationManager.cs:147 
  at YARG.Localization.LocalizationManager.TryParseAndLoadLanguage (System.String cultureCode) [0x00025] in /Users/admin/Development/YARG/Assets/Script/Localization/LocalizationManager.cs:117 

[<a href="Assets/Script/Localization/LocalizationManager.cs" line="62">LocalizationManager.cs:LoadLanguage:62</a>] Failed to parse and load the default language! (no fallback)


```

asimard1:
@polson Sorry, there was a comma missing in the localization file, so the two problems were related. Should be fixed now!

Sorry for tagging you, I don't know if that's the proper way to do things :')

polson:
Found a few issues:

Problem:

**1. Replay viewing can change the saved song offset**

1. Turn on automatic song-offset calibration.
2. Finish a song and open its replay instead of starting a new song.
3. Watch at least 20 replayed notes.
4. Compare the saved song offset before and after watching the replay.

**Expected:** Replay playback does not change the saved offset.

**2. Changing specific song delay can make music, notes, and other parts fall out of sync**

1. Start a song with metronome enabled, crowd claps enabled, and a background video.
2. Pause the song and change its specific delay by a noticeable amount, such as 100 milliseconds.
3. Resume the song
4. Change song specific delay back to 0, resume

**Expected:** Music, notes, metronome, crowd claps, and video all stay in sync, including after a live change.  Crowd claps are already a little off normally, but make sure that the offset changes reset these

**3. Failed offset loads can erase saved offsets**

1. Create a song-offset file with valid entries.
2. Make the file contain invalid JSON.
3. Open any score screen and leave it without changing an offset.
4. Check the song-offset file again.

**Expected:** A failed load preserves existing saved data and does not replace it.

Also add the two missing Unity `.meta` files

asimard1:
@polson 

So for the 3 problems:
1. I moved the auto calibration turning off outside of the if statement it was in, that way it also turns off when loading a replay.
2. I added reschedule() for both metronome and claps, and made sure to call them when offset is changed. However, this adds a delay everytime the offset is adjusted, so I will have to check that out and report back... This delay makes calibration very awkward, so it's imperative that it is fixed. :(
3. For this, I figured the best way to do this was to add a fallback, and so I added a regex reading of the json file to recover offsets in case of a mismatch. Also, I added a check where the json file is only written when there actually is a change.

Edit: the delay is now fixed, so all three problems are gone.

asimard1:
Okay replaced a throw with a return and now it actually works.

asimard1:
Okay, calibration for songs without video is now perfect. However, when there is a video background, adjusting it takes too long and there is still a rather big lag (makes auto calibration really awkward). I will try to fix this right away.

asimard1:
@polson I think I'm finally done solving this problem, I have to say that when the offset is actually updated, there is a small change in the metronome, clapping and video background that can be distracting, but at least it stays well scheduled.

Let me know if there's anything else to look at, I know I made a lot of changes in the last three days so sorry about that 😅

Assets/Script/Playback/SongRunner.cs
```diff
@@ -248,23 +261,50 @@ public class SongRunner : IDisposable
...
+        /// </summary>
+        public void SetSongOffsetOverride(double songOffsetOverride)
+        {
+            SongOffsetOverride = songOffsetOverride;
+            SongOffset = -(_chartSongOffset + SongOffsetOverride);
+            AnchorTimeline(InputTime);
+        }
+
+        ~SongRunner()
+        {
```
polson: Any reason for doing this?  probably better not to I think

Assets/Script/Settings/SettingsManager.cs
```diff
@@ -52,6 +52,10 @@ private sealed class StartupSettingValues
                 nameof(Settings.AudioCalibration),
                 nameof(Settings.VideoCalibration),
                 new FieldMetadata(nameof(Settings.AccountForHardwareLatency), true, isAdvanced: true),
+                // Not sure how to use nameof?
```
polson: no need to, remove this comment

polson:
Looking good.  One thing about videos, we're still using the chart offset but not factoring in the saved offset.

This should be a pretty easy fix, just expose `GetAudioPlaybackTime` from `GameManager` and then use that for the time in `BackgroundManager.Update`

asimard1:
I'll get that fixed tomorrow, thank you ! :)

PR #1537, opened 2026-07-01
Prefer per-engine high scores and show engine ring

Ape:
With this commit we now prefer scores from the active engine preset when querying high scores, percentages, and stars. 

This commit also adds a engine-colored ring around the music library difficulty badge.

In this screenshot we have custom engine, precision engine, and default engine:
presets

### DIFF
--- a/Assets/Art/Menu/Common/DifficultyIcons.png
+++ b/Assets/Art/Menu/Common/DifficultyIcons.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7d27c18287d0868e24fae2d03ab25819b3af0b6102c81eb1bde166c10cb11cd8
-size 457654
+oid sha256:22098c722af6c2a68c7e0c0bbeb4d8ef446545c2fb935b49efa7c8a5912c2507
+size 505545
--- a/Assets/Art/Menu/Common/DifficultyIcons.png.meta
+++ b/Assets/Art/Menu/Common/DifficultyIcons.png.meta
@@ -207,6 +207,28 @@ TextureImporter:
       edges: []
       weights: []
     - serializedVersion: 2
+      name: Ring
+      rect:
+        serializedVersion: 2
+        x: 1024
+        y: 0
+        width: 512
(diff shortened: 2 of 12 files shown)


wyrdough:
Interesting idea! I'm not so sure about it from a UX and UI perspective, though. I think it could be unnecessarily confusing and not look great with the overall score screen design. People already sometimes get confused about what score is or is not showing. That said, I may very well be the only one with an objection, so if you want to bring it up on Discord to get wider feedback, you're more than welcome.

I do think at the very least this work could be useful on the upcoming more info screen where we intend to show local (and remote) leaderboard standings for each song.

Ape:
> I'm not so sure about it from a UX and UI perspective, though. 

Is that concerning the change about which score to show or the colored ring visualization?

wyrdough:
Both. The score because people already get confused about what score is being shown and why and the colored ring because it doesn't seem to blend well with the overall vibe of the library menu. (This isn't to say we *can't* do this, just that I'd like some more discussion/feedback first)

polson:
Perhaps this should be settings-ized?

> Both. The score because people already get confused about what score is being shown and why and the colored ring because it doesn't seem to blend well with the overall vibe of the library menu. (This isn't to say we _can't_ do this, just that I'd like some more discussion/feedback first)

polson:
Can you double check that History view shows the selected engine preset correctly?

Like:
 1. Get a high score with default engine
 2. Switch engine to casual and play same song, get a lower score.

Check engine ring in History

polson:
@Ape If you make this a setting, I'll merge it

Ape:
> @Ape If you make this a setting, I'll merge it

Something like?
- Prefer High Scores from Current Engine
- Show High Score Engine Indicator

I assume both should default to false for unchanged behavior.

polson:
I think a single setting would be better, and less complicated for users.  Maybe just `Prefer Current Engine for High Scores` which would show the indicator and prefer scores from current engine

Whatever you decide to call it is fine but I do think we should just do a single setting with the indicator.

Issue #1275, opened 2026-01-22
Implement an alternate timing engine flagged in the settings that enables multiple timing windows


adam-f-miller:
As a player, it would be fun to optionally focus on precision in addition to accuracy when playing a song.

Proposal: implement a new scoring engine and note timing engine for hit-based instruments that captures multiple degrees of precision, and represent this precision on a results screen at the end of a song.

Ape:
I think it would be neat to have the accuracy to affect scoring, perhaps like this or similar:
- <10 ms offset => 100% points
- \>70 ms offset (or whatever the hit window is) => 50% points
- linear interpolation for points between those two points

Combo would be broken only if the hit is outside the hit window (as before).

This kind of scoring system would really separate expert players from good players. Some other values might need to be readjusted so that the experience for beginner player doesn't get worse (things like star rating threshold maybe).

PR #1539, opened 2026-07-01
Add Tournament Engine

Ape:
This commit adds a new Tournament Engine. This has been a long-standing wishlist item for me and many other players. The engine is intented for players who care about competitive precision playing. Unlike any of the existing engines, this one can easily separate high-level players even on songs where they are likely to full combo.

This PR is my take on this feature. I have been play testing this implementation and I think it fits my ideal vision well. I'm open to comments of course!

This engine is not available on vocals since I don't even know what it would do there.

Needs accompanying Core PR https://github.com/YARC-Official/YARG.Core/pull/407.

# Details
## Star power is disabled
- In my opinion, star power doesn't really fit high-level competitive play. I mostly care about drums, though. (Maybe guitar players still like SP?)
- I wanted to have one engine for the leaderboards where nobody has star power since it obviously affects scoring.

## Score is based on timing accuracy
- Hitting the note within +-10ms will give the full points
  - That window exactly corresponds to the middle 5 bars in the current note offsets histogram.
- Barely registering a hit will give 50% points
- Anything in between these is scored with a linear interpolation

## The scoring percentage is based on the score
- For competitive play, it's easier to compare the percentages than absolute scores
- In this mode, the percentage is calculated by dividing the score by the potential maximum score for that song

## Multiplier (1x..4x) is unchanged
Some people have called out for removing the multiplier for competitive play, but I think it's still an effective and important way to punish overhits / overstrums.

We could consider limiting the multiplier to go up to 2x only.

## In-song accuracy indicator will be implemented separately
We will need some way to visualize the timing accuracy during play, but that's out of scope of this PR and can be done separately.

# Implications in practice
For songs that I could previously get around 99.7%, I can get something around 90% in Tournament Engine.

With this engine it's possible to FC with a score like 95%. That's fine. FC remains a motivating achievement.

# Related discussion
- https://discord.com/channels/1086048856678084609/1103199806291181629
- https://docs.google.com/document/d/161E3atVfryE3YvKe9rBzx-8Ee4rGVmutd43r6xdY6Dk/view
- #1275


Assets/Script/Menu/ScoreScreen/ScoreCards/ScoreCard.cs
```diff
@@ -267,6 +280,68 @@ public virtual void SetCardContents()
...
+                return;
+            }
+
+            int siblingIndex = section.GetSiblingIndex();
+            if (siblingIndex <= 0)
+            {
+                return;
+            }
+
+            var previousSibling = section.parent.GetChild(siblingIndex - 1);
```
polson: Best to use a [SerializeField] reference instead of finding by name.  Otherwise if someone removes this separator or renames it this will break

polson (changes requested):

Ape:
Rebased and now using `[SerializeField]` as suggested.

wyrdough:
Sorry for taking so long to get back to you. I think this is interesting, but making it a profile-selectable engine preset isn't really the way I'd prefer to go with this. I think this would work better as a separate game mode in the upcoming Game Modes submenu. It would still use the normal music library, setting a flag to indicate to the players that they need to create the engine using the special tournament preset.

I'm thinking the plan should be to first get career mode done (I should have it done in the next few days to a week, although I said that last weekend lol) so we have a game modes menu, then you do some slight rework to not expose the tournament preset as selectable in profiles and add "Tournament" (not sure that will be the permanent name, but it's fine for now) as a selection in the game modes menu and set a flag somewhere the players can use to know they should create the engine with the tournament mode preset and (in someone else's PR) use the streak/accuracy elements.

PR #1540, opened 2026-07-02
feat/library-export-enhancement: add web song browser export

XaiaX:
## Summary

Adds a fourth **"Web Page"** button to Settings → File Management → Export Songs (alongside Ouvert JSON / Plain Text / CSV) that exports the song library as a single self-contained HTML file: an offline song browser with search, per-instrument filter chips, sortable columns, and a tap-for-detail view. All data is embedded — no server, no network requests, no external assets — so the file can be opened anywhere or shared directly.

The generator streams a static template in four segments (`WebBrowserTemplate.Seg0`–`Seg3`, split at static init around the `/*DATA*/`, `/*GENRES*/`, `/*META*/` markers) interleaved with the JSON payloads, so the full HTML document is never materialized in memory; only the records JSON string (~1 MB for a large library) is held at once.

## Why

Users want to browse and share their song libraries outside of YARG — on phones, tablets, or machines without the game installed. The existing exports (Ouvert JSON, plain text, CSV) are data dumps that need other software to be useful; a self-contained HTML page is viewable everywhere with zero setup.

### Compact schema

Song records use short JSON keys: `a` artist, `t` title, `al` album, `y` year, `l` length (seconds), `vp` vocal-parts count, `p` parts code, `d` difficulty code, `g` genre index. `al`, `y`, and `g` are omitted entirely when absent.

- `d` is always exactly 5 characters in fixed V,G,D,K,B order: `.` = part absent, `?` = present but unknown tier, otherwise a base-36 digit (`0`–`9` for tiers 0–9, `a`–`z` for 10–35 headroom). Each slot aggregates its sub-instruments by max known tier (Guitar = 5/6-fret, Drums = 4-lane + Pro + 5-lane, Keys = Keys + ProKeys, Vocals = Vocals + Harmony, Bass = 5/6-fret).
- `p` is the *subset* of present parts, not fixed-width (e.g. `"VGD"`, `"VGk"`); Keys is uppercase `K` when ProKeys is present, lowercase `k` for 5-lane-only.
- Genres are interned into a separate `GENRES` array and referenced by index.
- Records are pre-sorted by (artist, title), invariant culture, ignore-case.

A real 9,366-song library exports to ~1.13 MB.

### Script-injection hardening

Artist/title/album/genre are user-controlled and flow into an inline `<script>` block. Newtonsoft escapes `"` and `\` but not `<`, so a literal `</script>` or `<!--` in a song name could terminate the script early. `WebBrowserExport` escapes every `<` to `\u003C` in the DATA and GENRES payloads before writing; JSON has no structural `<`, and the JS engine decodes the escape back, so values round-trip unchanged while the page cannot be broken out of. (This is more to prevent weird metadata values from accidentally breaking the output than an actual injection attack)

### Template

The page is a full single-page app embedded as one C# verbatim string:

- **Virtualized rendering** — only visible rows are in the DOM; handles 5000+ songs without jank (stress-tested at 10k rows).
- **Filter chips** — per-instrument part filters (V/G/D/K/B) with brand-colored active states; collapse to emoji + initial at narrow widths.
- **Gear-icon search settings** — popover with independent artist/title/album search-field toggles (artist + title on by default).
- **Sortable columns** — click any header to sort, with ▲/▼ indicators; artist ascending is the default.
- **Detail bottom sheet** — tap a row for full metadata, per-instrument tiers with numeric prefix and tier coloring (orange at tier 5, red at 6+), and vocal-parts info (`N Harmony Parts` / `Solo Vocals` / `None`).
- **Instrument badges** — the V badge shows "H" when harmonies are present (`vp >= 2`); grey dash for absent instruments.
- **Dark/light theme** — sun/moon toggle, persisted in localStorage (with graceful fallback when unavailable).
- **Progressive mobile hiding** — breakpoints at 900/720/540 px hide columns right-to-left; artist, title, and length always visible.

The static initializer validates the three markers (present, unique, in document order) and throws otherwise, so a bad template edit fails fast at startup instead of silently emitting broken pages. (Useful for a potential future test harness)

### Unity constraint: plain class, not record

`WebBrowserRecord` is a plain `class` with `{ get; set; }`. Unity 6000.3's C# runtime lacks the `IsExternalInit` shim, so `record` / `init` accessors fail with CS0518 (the first Unity compile of this branch caught exactly that). The type is a constructed-once DTO serialized immediately, so mutability is harmless.

## Changes

| File | Change |
|------|--------|
| `Assets/Script/Song/WebBrowserExport.cs` | Streaming generator — writes template segments interleaved with escaped JSON payloads (DATA/GENRES as raw JSON literals with `<` escaped to `\u003C`, META raw) |
| `Assets/Script/Song/WebBrowserTemplate.cs` | Verbatim HTML/CSS/JS template (999-line file), split into Seg0–Seg3 at static init with marker validation |
| `Assets/Script/Song/WebBrowserRecord.cs` | Compact record DTO with single-letter JSON keys, parts/difficulty code builders, (artist, title) sort |
| `Assets/Script/Song/WebBrowserDifficulty.cs` | Pure base-36 tier encoder + multi-sub-instrument slot aggregation (`.` / `?` / digit) |
| `Assets/Script/Song/WebBrowserGenreTable.cs` | Genre interner — deduplicates trimmed genre strings, returns index or null for empty |
| `Assets/Script/Song/SongExport.cs` | Adds `WebBrowser` to `ExportFormat`, wires save-dialog callback (`.html`) |
| `Assets/Script/Settings/SettingsManager.Settings.cs` | Adds `ExportSongsWeb` button handler |
| `Assets/Script/Settings/SettingsManager.cs` | Adds the fourth button to the Export Songs `ButtonRowMetadata` |
| `Assets/StreamingAssets/lang/en-US.json` | `"ExportSongsWeb": "Web Page"` label |

## Testing done

The Unity project has no test assembly (out of scope for this PR), so verification was layered:

- **Standalone compile + logic tests (C#):** `WebBrowserDifficulty`, `WebBrowserGenreTable`, and the `WebBrowserRecord` builder were compiled and run in a throwaway `dotnet` project with stub `SongEntry`/`Instrument`/`RichTextUtils` types and the real Newtonsoft.Json — 20/20 encoder cases and 8/8 record-builder cases passed, plus an end-to-end generator run confirming marker replacement, raw GENRES/META literals, omitted optional fields, `?` emission, `</script>` neutralization, and empty library → `[]`.
- **Template behavior (browser, Playwright):** verified against fixture pages in the fork's local tooling workspace (not part of this PR) covering tiers 0–6 and base-36 tiers ≥ 10, `?` unknown, `.` absent, harmony (`vp` = 2, 3), solo (`vp` = 1), instrumental (`vp` = 0), all-parts (VGDKB), keys-without-pro (lowercase `k`), Japanese text, long titles, minimal fields, and an empty-DATA page — 18 display acceptance criteria, virtualization stress-tested at 10k rows, zero network requests confirmed.
- **Byte-identity:** the verbatim string in `WebBrowserTemplate.cs` reconstructs the reference `template.html` exactly (33,713 chars, re-verified 2026-07-01).
- **Unity compile gate:** the branch compiles clean in Unity 6000.3 (the compile loop surfaced and fixed the `record`/`init` CS0518 issue described above).
- **In-game:** exported a real 9,366-song library via the new Settings button, opened the result in a browser, and confirmed search, filter chips, sorting, detail sheet, theme toggle, and mobile responsive behavior. Re-verified 2026-07-02 on the final revision (raw-literal DATA + review-pass fixes) after a clean Unity recompile.
- **C#↔Python parity:** record-by-record diff against the reference Python generator on the same 9,366-song library — 9,339 exact matches; every divergence explained (25× the intentional `?`-vs-clamped-0 asymmetry, 2× trailing whitespace since fixed by the trim).
- **Reference test specs:** [`2026-07-01-web-browser-export-unity-test-specs.md`](../../notes/2026-07-01-web-browser-export-unity-test-specs.md) enumerates the unit cases to implement if/when a Unity test harness exists.

## Notes

- Unity-side only — no YARG.Core changes, consistent with the existing exporters in `SongExport`.
- The template is a compile-time string constant: no runtime compilation, Addressables, or asset-pipeline dependency. The `.cs` compiles regardless of the HTML content.
- Line endings inside the verbatim string are LF (the exported page's own bytes). `.editorconfig` enforces CRLF for `.cs` files; the verbatim string interior must not be reformatted, or the emitted HTML silently changes. An editor that normalizes line endings on save will break this.
- Only `en-US` gets the new label; other languages flow through Crowdin as usual.


wyrdough (changes requested):
Tiny request: Move the export scripts (both new and existing) into a new subfolder called `Exporters` or something so we can keep things tidy...

XaiaX:
Added the folder and also put them into a namespace for `YARG.Song.Exporters`. I considered `YARG.Settings.Exporters` but they're really only under settings because we don't have a more specific category like "Library Management" or something like that.

XaiaX:
Actually, just a sec, noticed a couple minor issues on the css (sort indicator on length is running up against the "Y" in "year".

XaiaX:
Ok, done, tweaked the spacing on the length/year columns a little and changed the justification on the length column to the right so that the times line up better. (plus a little right-padding to create some breathing room to the next column)

mutantx22:
Can you add a column that shows the HASH and the source type (ini or CON file) values (like in the CSV export)?

That would be useful for  comparing and sharing songs between people

PR #1541, opened 2026-07-02
Play-a-show Toggle

asimard1:
Add a button to toggle the Play a show feature, for those who don't wish to use it. It can get a bit confusing in parties when people press it on accident, and it's a menu that is not easy to escape (Red is not mapped to back).

Obviously another good options would be to map Play a show to a holdable button, I might add that also.


wyrdough (changes requested):
Functionally this looks good to me, you just need to rework the navigation scheme setup a bit so that yellow comes before blue. (we want to keep them in GRYBO order as much as possible)

Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
```diff
@@ -348,14 +348,32 @@ public void SetNavigationScheme(bool reset = false)
...
-                    new NavigationScheme.Entry(MenuAction.Yellow, "Menu.MusicLibrary.StartSet", StartSetlist) :
-                    new NavigationScheme.Entry(MenuAction.Yellow, "Menu.MusicLibrary.PlayShow", EnterShowMode),
                 new NavigationScheme.Entry(MenuAction.Blue, "Menu.MusicLibrary.Filters", OpenFilters),
                 new NavigationScheme.Entry(MenuAction.Orange, "Menu.MusicLibrary.MoreOptions",
                     OnOrangeHit, OnOrangeRelease),
-            }, false));
+            };
 
+            // Give yellow the same behaviour as green: press to add to set, hold to start the set
+            if (SettingsManager.Settings.EnablePlayAShow.Value)
```
wyrdough: Probably the easiest way to get the ordering correct is something like this, added before the `var entries...`

```
NavigationScheme.Entry yellowEntry;

if (SettingsManager.Settings.EnablePlayAShow.Value)
{
    yellowEntry = new NavigationScheme.Entry(...);
}
else
{
    yellowEntry = new NavigationScheme.Entry(...);
}
```

and then you can change the `var entries` declaration to have the following in the middle:

```
new NavigationScheme.Entry(MenuAction.Red, "Menu.Common.Back", Back, hide: true),
yellowEntry,
new NavigationScheme.Entry(MenuAction.Blue, "Menu.MusicLibrary.Filters", OpenFilters),
```

Longer term, it would probably be best to rework the NavigationScheme system to allow inserting entries since there are several places where we have to do some gymnastics to deal with buttons that change function based on context, but that's a me problem not a you problem lol

asimard1: Okay I'll do the order change and update the PR.

asimard1:
Sorry I requested a review I don't know I should have done that D:

wyrdough (approved):

PR #1545, opened 2026-07-06
Unison Display

Supersonicboss1:
Unison display stuff. I tried to code this in a similar way to the lyric bar, since when it appears and disappears is based entirely on the chart, but this ended up being significantly more complicated since it relies on a bunch of player input stuff. Code should be relatively performant, no LINQ outside of init, limited use of DOTween, and most updates are event-driven.

Subscribing to events kind of sucks due to no useful base class, let me know if there's a more clean way to do this.

Code also includes some changes to call sites in the various [instrument]Player for the core changes I made.

Unison display should be fully localizable, including the settings. In single-player, when enabled, it is also draggable.

Requires YARC-Official/YARG.Core#409


wyrdough:
Pls rename "Unison Box" etc to "UnisonBox" to fit with the way all the other files are named.

Assets/Script/Gameplay/HUD/UnisonBonus/BaseUnisonObject.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections.Generic;
+using YARG.Core;
+
+namespace YARG.Gameplay.HUD
+{
+    public abstract class BaseUnisonObject : GameplayBehaviour
+    {
+        protected Dictionary<int, bool> ParticipantFailState  = new();
+        protected Dictionary<int, int>  ParticipantTotalNotes = new();
+        protected Dictionary<int, int>  ParticipantNotesHit   = new();
```
wyrdough: I'm not sure how much it really matters in this case, but just in general array indexed access is faster than a dictionary lookup, so when you're doing `Dictionary<int,something>` and the range of possible indexes is reasonably small (or well packed) it's better to just use an `Array<something>`.

Assets/Script/Gameplay/HUD/UnisonBonus/UnisonBar.cs
```diff
@@ -0,0 +1,116 @@
...
+        private void Update()
+        {
+            if (Mathf.Approximately(_fill.fillAmount, Progress))
+            {
+                return;
+            }
+
+            // Lerp to new progress
+            _fill.fillAmount =
+                DOVirtual.EasedValue(_fill.fillAmount, Progress, Time.deltaTime * 15, Ease.OutCubic);
```
wyrdough: Hm, normally I'd say this needs to check that `_hitNotes` actually changed, but I guess that doesn't work so well since you're not actually using a tween here.

Assets/Script/Gameplay/HUD/UnisonBonus/UnisonBar.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            if (_isFailed)
+            {
+                return;
+            }
+
+            if (_fill.fillAmount >= 0.99f)
+            {
```
wyrdough: This should probably also check whether `_hitNotes == _totalNotes` just to cover the pathological case of a 100+ note unison phrase.

Assets/Script/Gameplay/HUD/UnisonBonus/UnisonDisplay.cs
```diff
@@ -0,0 +1,596 @@
...
+            }
+
+            ResetState();
+        }
+
+        private void SubscribeToEngineEvents(EngineManager.EngineContainer engineContainer)
+        {
+            int engineId = engineContainer.EngineId;
+
+                        switch (engineContainer.Engine)
```
wyrdough: Minor whitespace issue here

Assets/Script/Gameplay/HUD/UnisonBonus/BaseUnisonObject.cs
Supersonicboss1: I assume the same applies to Lists? Using arrays is kind of a headache because I check the length of the list to get the participant count, but I can switch if it's significantly different.

Supersonicboss1: Nevermind, ignore me. using lists with `Insert` was dumb and arrays is the saner solution.

wyrdough (approved):

PR #1547, opened 2026-07-07
Bring back vocals track fade

theli-ua:
Was lost likely when vocals were incorporated into highway renderer.
Vocals track shader used UV coords to apply gradual transparency on the edges. This no longer works as edges of the track extend past the screen. A fix is to simply use screen coordinations instead so that it would apply regardless of tracks' stretching or positioning.

Fixes https://discord.com/channels/1086048856678084609/1521934526954078271

This only swaps input node from UV to screen position. But as usual saving anything in Unity comes with a lot of re-serialization baggage


theli-ua:
Fixed:

image

Dev:

image

PR #1556, opened 2026-07-11
dmx: Add configurable local sACN IP address

Ape:
The default behavior of automatic interface selection remains, but now you can override it if it didn't choose the correct network.

### DIFF
--- a/Assets/Script/Settings/Types/IPv4Setting.cs
+++ b/Assets/Script/Settings/Types/IPv4Setting.cs
@@ -8,17 +8,28 @@ namespace YARG.Settings.Types
     {
         public override string AddressableName => "Setting/IPv4";
 
-        public IPv4Setting(string value, Action<string> onChange = null) : base(onChange)
+        private readonly string _defaultValue;
+
+        public bool AllowEmpty { get; }
+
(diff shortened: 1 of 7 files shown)


Assets/Script/Settings/Types/IPv4Setting.cs
```diff
@@ -8,17 +8,28 @@ public class IPv4Setting : AbstractSetting<string>
...
+            {
+                _value = string.Empty;
+                return;
+            }
+
             if (!IsValidIPv4(value))
             {
-                // If it's invalid, just default to this
-                _value = "255.255.255.255";
+                _value = _defaultValue;
```
wyrdough, line 32: If I'm reading this right, `_defaultValue` will either end up being an empty string (in the case of a brand new install) or whatever value we loaded from JSON. Is that really what we want?

Ape: `_defaultValue` always comes from the setting declaration; JSON deserialization updates only `_value` on the existing setting instance. So for `DMXLocalIP`, `_defaultValue` is always an empty string, while a valid value loaded from JSON becomes the current `_value`.

I’ve renamed the constructor parameter from `value` to `defaultValue` and made the call site explicit to clarify that distinction.

PR #1557, opened 2026-07-11
Support lanes on Medium, Easy, and Beginner

Frickitickitavi:
[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/413)

Previously, `.MID` only supported lanes on X and H (other than on PK). Non-PK lanes in `.MID` are X-only by default, and can be configured to XH by setting the lane marker's velocity in the range `[41,50]`. This behavior is inherited from RB3.

CH introduced an extension to this system, whereby the range `[31,40]` applies a lane to XHM and `[21,30]` applies it to XHME. In theory, introducing this without hiding it behind a text event feature flag could cause issues with legacy charts, but in practice this doesn't seem to have happened, so the cat's out of the bag as far as I'm concerned.

In addition to implementing those two new velocity ranges, this PR also implements wildcard lanes on Beginner. These were already theoretically possible due to `.CHART` supporting per-difficulty lanes on drums, but I considered them an incidental edge case and [previously introduced logic to strip them out of Beginner charts](https://github.com/YARC-Official/YARG.Core/pull/375). Now that we have more robust and intentional support for lanes on Easy across all parts and formats, I'd like to properly support Beginner lanes.

In gameplay, wildcard lanes act as you would expect - any input counts toward keeping the lane going, and it's completely impossible to overhit during them (and during the little proximity forgiveness windows, for that matter). On the parsing side, I introduced a little new logic to turn E trills into B tremolos, since B is all wildcards and a trill of all one note isn't valid.

Wildcard lanes use the default color profile value (a pale green). Could be cool to make them pulse in rainbow colors, but that would require a whole new shader and extra logic, which is beyond both my capabilities and what's worth doing for such a niche feature.

Architecturally, this isn't the cleanest thing I've ever written, so feel free to nitpick my encapsulation and inheritance. I've thoroughly tested all of the gameplay logic though, and everything seems to be in order.

### DIFF
--- a/Assets/Script/Gameplay/Player/DrumsPlayer.cs
+++ b/Assets/Script/Gameplay/Player/DrumsPlayer.cs
@@ -40,7 +40,7 @@ namespace YARG.Gameplay.Player
         private bool _greenCymbalHasLane = false;
 
         public int NumberOfDedicatedKickLanes { get; private set; } = 0;
-
+        public int CenteredPosition => (LaneCount - 1) / 2;
 
         public float NoteScaleFactor = 1f;
         private float _baselineLaneCount => _fiveLaneMode ? 5f : 4f;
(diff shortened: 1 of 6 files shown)


Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -386,7 +386,20 @@ protected override void InitializeSpawnedNote(IPoolable poolable, DrumNote note)
...
+            HighwayOrderingInfo highwayOrderingInfo;
+
+            if (_fiveLaneMode && note.Pad is (int)FiveLaneDrumPad.Wildcard)
+            {
+                highwayOrderingInfo = new(CenteredPosition, (int) FiveLaneDrumPad.Wildcard);
+            }
+            else if (!_fiveLaneMode && note.Pad is (int) FourLaneDrumPad.Wildcard)
+            {
+                highwayOrderingInfo = new(CenteredPosition, (int) FourLaneDrumPad.Wildcard);
+            }
```
wyrdough, line 398: I would complain about allocating here, but I'm fairly certain that since this is a struct that isn't getting passed anywhere it will get allocated on the stack and not make it to the heap, so it's fine.

wyrdough (approved):
lgtm

PR #1562, opened 2026-07-13
Add Skip to Score Screen feature

Ape:
image

### DIFF
--- a/Assets/Prefabs/Gameplay/HUD/Pause/QuickPlayPause.prefab
+++ b/Assets/Prefabs/Gameplay/HUD/Pause/QuickPlayPause.prefab
@@ -317,6 +317,9 @@ PrefabInstance:
       addedObject: {fileID: 5062416941184210017}
     - targetCorrespondingSourceObject: {fileID: 8918172421510936374, guid: d0426e4d1cafc1c489d1df380508ec96, type: 3}
       insertIndex: -1
+      addedObject: {fileID: 7300000000000000002}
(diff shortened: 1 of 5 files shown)


polson:
Can you add a screenshot of the menu item?

Ape:
> Can you add a screenshot of the menu item?

Done.

polson:
When fail mode is on, skipping to score screen just goes to the fail screen.  I feel like this is not intended?

Ape:
> When fail mode is on, skipping to score screen just goes to the fail screen. I feel like this is not intended?

You can still use it at the end of the song.

polson:
> > When fail mode is on, skipping to score screen just goes to the fail screen. I feel like this is not intended?
> 
> You can still use it at the end of the song.

Sorry I'm not understanding.  My feeling is clicking "Skip to score screen" should go to the score screen, fail mode or not.  What do you think?

Ape:
> > > When fail mode is on, skipping to score screen just goes to the fail screen. I feel like this is not intended?
> > 
> > You can still use it at the end of the song.
> 
> Sorry I'm not understanding.  My feeling is clicking "Skip to score screen" should go to the score screen, fail mode or not.  What do you think?

Yeah, but it also shouldn't allow passing songs that you otherwise wouldn't. So I guess it should disable score saving and clearly label the score screen failed.

polson:
> Yeah, but it also shouldn't allow passing songs that you otherwise wouldn't. So I guess it should disable score saving and clearly label the score screen failed.

I like this idea better than going to fail screen personally.  We already have a "Score not saved" message on score screen for other cases, like playing with bots, too many pauses, etc.

PR #1564, opened 2026-07-14
Add the option to adjust the tint on the song progress overlay

RubyStarbloom:
Added a color-selection option to add a tint to the song progress overlay around the score counter.

I am colorblind and have had difficulty seeing this at a glance with the default blue. This just allows a player to customize to something with higher contrast.

### DIFF
--- a/Assets/Script/Gameplay/HUD/ScoreBox/ScoreBox.cs
+++ b/Assets/Script/Gameplay/HUD/ScoreBox/ScoreBox.cs
@@ -111,6 +111,7 @@ namespace YARG.Gameplay.HUD
             _displayedCountDownSeconds = -1;
 
             _songProgressBar.SetProgress(0f);
+            _overlayImage.color = SettingsManager.Settings.GraphicalSongProgressTint.Value;
         }
 
         protected override void OnChartLoaded(SongChart chart)
--- a/Assets/Script/Helpers/JsonUnityColorConverter.cs
+++ b/Assets/Script/Helpers/JsonUnityColorConverter.cs
@@ -1,0 +1,22 @@
+using System;
+using Newtonsoft.Json;
+using UnityEngine;
+using YARG.Helpers.Extensions;
+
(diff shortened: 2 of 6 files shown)


wyrdough (changes requested):
Thanks for the PR! In terms of the setting itself and how it functions, this is 100% spot on.

Unfortunately, you can't directly serialize a `ColorSetting` because it uses `Unity.Color` internally, which has recursive references that make NewtonSoft sad. Probably the easiest thing to do is to add another converter like `JsonColorConverter` that converts `Unity.Color` to `System.Drawing.Color` before serialization and back when deserializing. It should be relatively straightforward since we already have extensions to convert between the two, so you can just do things like `systemColor = unityColor.ToSystemColor()` and `unityColor = systemColor.ToUnityColor()` to swap between them. 

The part I'm not as clear on is how NewtonSoft would handle the case with multiple converters when the first converter converted a value to something the second converter would itself want to convert. I *presume* we would want `UnityColor` to get transmuted into a `System.Drawing.Color` and then run through the existing `JsonColorConverter` to do...whatever it is that does... If it really does just run through the list in order, it ought to work as desired, but I don't actually know that to be true.

In the future it would be helpful if you made sure your editor wasn't unnecessarily adjusting whitespace or end of line markers, as it makes the diffs on GitHub pretty useless since it looks like a wholesale replacement in the diff despite being just a minor change.

RubyStarbloom:
Thanks for the feedback - have added in a new `JsonUnityColorConverter` to handle this.

I have done some testing around chaining the converters and as far as I can tell yes, they are operating sequentially. Making a change to the existing `JsonColorConverter` to force a specific color value overwrites what was done by the new converter.

Have also fixed the whitespace issue, turns out my git client was just hiding this one. Should now be able to diff normally!

wyrdough (approved):
Thanks for the quick turnaround! Don't forget to PR the Credits repo with your info if you want to be listed in the game credits.

PR #1353, opened 2026-02-26
feat: library sorting options: percentage and score, fully localized

Nort1346:
This PR adds two new library sorting options:

- Percentage
- Score (raw score)

Both options are fully localized

This addresses a previous feature request: [Issue #675](https://github.com/YARC-Official/YARG/issues/675)

This is my first contribution to the project
Thank you for maintaining this project. I really appreciate the work you're doing

PR #1565, opened 2026-07-15
Add Current Difficulty Score Display Options and Sync Star Sorting

Dansla116:
**Current Difficulty Display**
- Added **Best percentage on current difficulty** and **Highest score on current difficulty options** to the **High Score History** setting.
- When selected, these options only show stars, score, and percentage for the current difficulty.
- "Current difficulty" is the difficulty most recently selected from the Select Instrument menu after choosing a song.
- Like the four existing options, these solo high-score display modes are ignored unless exactly one human player is connected.
image

**Star Sorting and Display**
- Based on feedback from #1353, adjusted the initial Sort by Stars behavior. That PR may now have conflicts with this one.
- Previously, the score display and star grouping could use different difficulty scopes. For example, the display could consider all difficulties while star grouping only considered the current difficulty.
image

- In the example above, I earned 4 stars on Easy and 5 stars on Hard for Therefore I Am. The display correctly shows the best stars from Hard, but the song was grouped by the 4-star Easy score because Easy was the current/last-selected difficulty.
- Updated star sorting so songs are grouped by the stars from the same score record selected by **High Score History**.
- Added refresh handling so displayed stars and star sorting update when profiles are added/removed, the selected difficulty or instrument changes, or the **High Score History** setting changes. This avoids stale display or grouping data.

_Note:_ This is built on top of #1548, which includes similar changes around reapplying filters. That PR should either be merged first, or both PRs should be merged together.

### DIFF
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -261,13 +261,20 @@ namespace YARG.Menu.MusicLibrary
             }
             Instrument currentInstrument = profile?.CurrentInstrument ?? Instrument.FiveFretGuitar;
             Difficulty currentDifficulty = profile?.CurrentDifficulty ?? Difficulty.Expert;
-            if (_needsReload ||
+            bool profileSortContextChanged =
                 currentInstrument != _lastInstrument ||
-                currentDifficulty != _lastDifficulty)
+                currentDifficulty != _lastDifficulty;
+
+            if (_needsReload || profileSortContextChanged)
             {
                 _lastInstrument = currentInstrument;
                 _lastDifficulty = currentDifficulty;
                 _needsReload = false;
+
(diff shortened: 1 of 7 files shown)


polson:
@Dansla116 should we get #1548 in before we do anything with this?  Or does this PR do the same things?

Dansla116:
Everything in #1548 is already included in #1565, but the two PRs solve different problems.

#1548 is a small fix that ensures active filters are reapplied whenever the Music Library opens.  I expect that one can be reviewed and merged fairly quickly.

While it was still waiting, I started #1565 on top of that branch, so its changes naturally carried over.  #1565 is primarily about the new current-difficulty score display options and keeping the displayed score, star sorting, and refresh behavior in sync.

So we have two reasonable options:

- Merge #1548 first, then I can rebase #1565 onto dev.
- Treat #1548 as superseded by #1565 and close it, since all of its changes are already present here.

They do not do the same thing, but merging both as they currently stand would be redundant.

Dansla116:
That looked disastrous for a minute. 😅  Okay, this should be good to merge now.

PR #1567, opened 2026-07-16
Vocals visual enhancements

Supersonicboss1:
This encompasses three things, based on unused textures I found in the game files (note: I don't know why they were unused, but they look good to me lol).

## Combo Rim
It is now coloured gold for an FC, similar to the tracks, and blue when groove is active (4x multiplier).
Screenshot 2026-07-16 225158
image

## Sunburst Effects
Again, similarly to the tracks, blue on a groove, yellow on SP.
(yes, I used the same screenshot again, sue me lol)
image
Screenshot 2026-07-16 225147

## Combo multiplier animation
This one was not found in the game files, but I tried to create an animation vaguely similar to RB when your multiplier increases (does not apply to increases from SP)

https://github.com/user-attachments/assets/ac2eb7e8-7811-48e2-9fbe-3bc08d16d2c8


XaiaX:
I like the animation but I think it's a little *too* prominent. I'm glad someone is adding the gold for FC, though, that's an odd gap.

PR #1570, opened 2026-07-18
Minor UnisonDisplay fixes

Supersonicboss1:
This fixes two bugs in UnisonDisplay
- Display would sometimes stay on screen when FPS is low
- Extremely low FPS or extreme lag could cause the UI to freeze up as a tween is not initialised (somehow?)

This does not fix the practice mode bug. That will come later once I actually figure out how I want to fix it


Supersonicboss1:
Okay, in investigating the practice mode issue it turns out EngineManager is kinda super broken in practice mode (and also may have memory leaks?). I don't feel like having to resolve merge conflicts with my own EngineContainer PR (YARC-Official/YARG.Core#414), so for now I have pushed a commit to disable UnisonDisplay in practice mode completely. When that gets merged I will look into fixing EngineManager.

PR #1571, opened 2026-07-20
Fix: Song progress overlay desynced from actual song progress

RubyStarbloom:
Really noticeable with a high contrast song progress overlay but the custom position in the mask was causing the bar to be offset, so songs would end with the bar having about 5% left to show.


wyrdough:
Thanks for keeping at it! 

Normally a change this small and pretty obvious in terms of what it's doing and how it is doing it I'd just merge and revert if it broke something, but since it affects the scene file itself it's super conflict-prone, so I need to actually test it first. I should have a chance to do that tomorrow...

PR #1573, opened 2026-07-20
Implement per-stem drum muting

DevPrice:
Today, when you miss any note on drums, all drums stems are muted. For tracks that have multiple drum stems, we can do better and only mute the stem(s) that are directly associated with the missed gem(s). This PR adds support for muting individual drum stems based on which gems are missed.

See Core PR: YARC-Official/YARG.Core#425

The key changes in this PR are:
* Move muting logic from `TrackPlayer` into the subclasses, since drums now needs to handle muting differently than other (single-stem) types
* In `DrumsPlayer`, determine which stem a note should mute based on its corresponding `DrumStem` value, the current difficulty, and what stems are present in the mixer.
* Broadly handle changes in Core, such as `ToSongStem()` becoming `ToSongStems()` and the new `SongStem` enum values

Which stems to use is determined solely by which stem files are present when the song is loaded; the mix events in the chart are ignored (except for the disco flip status).

I tested this in every configuration I could think of, specifically:
* The basic case -- missing a gem mutes the correct stem for songs with multiple drum stems, hitting a gem also unmutes the corresponding stem
* Lefty flip still mutes correctly
* Disco flip swaps red/yellow drum stems on 4 lane, but not on pro drums
* Songs with only two drum stems work correctly (drums1 and drums2, but no drums3)
* We fall back gracefully if a song only has a single drum stem
* On easy difficulty, yellow/blue/green gems will unmute the kick as well (and vice versa)
* On beginner difficulty, there is functionally no change. Missing a note mutes everything and hitting a note unmutes everything.


Frickitickitavi:
`drums4` is used in GH charts and some customs (I think maybe even some official YARG content; I forget - we'd certainly use it for any charts where we have the right stems). Should definitely be supported

DevPrice:
> `drums4` is used in GH charts and some customs (I think maybe even some official YARG content; I forget - we'd certainly use it for any charts where we have the right stems). Should definitely be supported

It was "supported" in the sense that it was just merged with the drum3 stem, but I went ahead updated the PR to handle it.

Issue #675, opened 2024-02-09
Feature Request: Sort by Percentage


LostChris894:
It would be nice if we had to sort our songs by Percentage instead of all of our songs scrambled to find them, as this feature was in Clone Hero

Dansla116:
This should be available in tonight's (8/22/26) nightly build.

PR #1580, opened 2026-07-22
Percentage and Score Sorting, Resolution Helper

Dansla116:
This is a continuation of #1353 by @Nort1346.

Adds two new library sorting options, **Percentage** and **Score**

image

I've made changes so this works with the new **High Score History** options and logic that was already applied to sorting by **Stars**.
image

Best percentage across all difficulties, sorted by Percentage:
image

Best percentage on current difficulty (Expert), sorted by Percentage:
image

I either don't have good examples of other options/sort combinations or they're redundant.

**Feedback Item:** I retained secondary sorts within the groups - first by intensity, then by song title.  This is how sorting by **Stars** works.  Although, with stars, every song within each group is the same.  But with **Percentage**, all songs between 90-99.9% are mixed together.  Likewise, for **Scores**, all songs between 300,000-399,999 points are also mixed together.  Let me know if songs within the groups should be ordered high-to-low.

This fixes: #675.

### DIFF
--- a/Assets/Script/Menu/Filters/FiltersMenu.cs
+++ b/Assets/Script/Menu/Filters/FiltersMenu.cs
@@ -650,7 +650,8 @@ namespace YARG.Menu.Filters
                     }
                     else
                     {
-                        if (sort != SortAttribute.Playcount && sort != SortAttribute.Stars)
+                        if (sort != SortAttribute.Playcount && sort != SortAttribute.Stars &&
+                            sort != SortAttribute.Percentage && sort != SortAttribute.Score)
                             SettingsManager.Settings.PreviousLibrarySort = sort;
 
                         SettingsManager.Settings.LibrarySort = sort;
--- a/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
+++ b/Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.cs
@@ -1236,7 +1236,8 @@ namespace YARG.Menu.MusicLibrary
 
         private static bool IsDynamicScoreSort(SortAttribute sort)
         {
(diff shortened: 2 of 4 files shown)


Dansla116:
Responding to my own **feedback item**, not sorting groups by exact percentage or score looked ugly and unintuitive, so I've corrected that.  Groups (e.g. 90-99%) will now have highest percentages at the top and lowest percentages at the bottom.
image

Kept 3rd sort by instrument intensity and 4th sort by track name that is already used for unplayed songs and sort by stars.

Added **IsFC** sort for sorting by percentage specifically for 100%.  That is, within the 100% group, FCs are on top, followed by non-FC 100%s.

wyrdough:
Wouldn't it be simpler to query the score db for the highest score/percentage plays and have it return them in the correct sorted order rather than having to have a comparer and all that? (Genuinely asking, it's been a while since I've looked, so it is possible that would actually be harder/slower/whatever)

Dansla116:
That's a fair question.

It already does query the DB for the best score and best percentage play for each song in bulk.  `ScoreContainer` caches those records by song checksum, so the loop here is doing cache lookups rather than querying once per song.

The DB could return the played records ordered by score/percentage, but that would only provide part of the intended ordering.  I still need to merge them with the loaded song library to include unplayed and no-part songs, divide everything into the displayed buckets, and break ties by instrument intensity and title.  Those last two values are song metadata and aren’t in the score DB.

We could add a separate ordered DB API and preserve its ordering during that merge, but I'm not sure it would be simpler than sorting the completed buckets here.  Performance-wise, this is one bulk DB load followed by in-memory dictionary lookups and sorting.

wyrdough (approved):

polson (approved):

PR #1581, opened 2026-07-22
Kick lanes

Frickitickitavi:
Implements kick lanes, my beloved

[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/428)

## Chart Notation
In `.mid`, kick lanes are marked with note 125, right below the other two lane marker notes (126 and 127). It uses the same velocity-based system for LD application (`[41,50]` for XH, `[31,40]` for XHM, `[21,30]` for XHME, else X-only).

In `.chart`, kick lanes use special phrase 67, right after the other two lane phrases (65 and 66).

Both of these values have no defined meanings and I've exhaustively scanned legacy charts to ensure that they don't appear anywhere. I've also confirmed with the CH and MS devs that no other planned features will collide with these. As such, I'm not hiding these behind a text event feature flag, although I'm not dead-set against doing so if we want to be extra careful.

## Chart Parsing
Kick lanes are parsed independently of other lane types; they can overlap in any arbitrary arrangement. They're parsed similarly to tremolo lanes, requiring a minimum of two kicks to constitute a valid lane. This can produce differences between what is and isn't a kick lane on X and X+ (although any such cases are pretty questionable on the part of the charter).

On Beginner, kick lanes become wildcard tremolo lanes. Overlapping kick and non-kick lanes are stitched together into unified wildcard tremolo lanes. God help whoever makes a chart that actually makes use of this.

## Engine Behavior
Kick lanes are handled independently from other lanes and have their own autohit timer.

The autohit timer uses the same engine configuration setting as regular lanes, but outside of X+ the timer increment is doubled (equivalently, the required input rate is halved) because the player only has one limb to use and the autohit timer is calibrated to a two-limbed input roll.

For the most part, the `Note` and `DrumNote` classes treat kick lanes as entirely separate from lanes - they're flagged using bespoke `DrumNoteFlags` instead of the usual `NoteFlags` and have their own `IsKickLane[Start/End]` fields. The existing `IsLane[Start/End]` fields and flags continue to refer only to regular "hand" lanes, and new code comments clarify this when relevant.

However, we do count kick and hand lanes together for the `LanedNotesHit` post-song statistic. This is accomplished by a new `virtual bool` field in `Note` called `IsAnyLane`. By default, it simply returns `IsLane`, but `DrumNote` overrides it to return `IsLane || IsKickLane`. All other existing instances of `IsLane[Start/End]` throughout YARG.Core remain untouched and continue to refer to hand lanes.

Along the way, `CurrentLaneIndex` and `LanesExist` started causing some trouble. I looked into how they were being used, and as far as I could tell they were really only being used for a few early-exit conditions, most of which seemed redundant anyway (e.g. "if no lanes remain, exit early, otherwise iterate through the remaining lanes then exit"). I ripped them out and nothing seems to have regressed as a result, but let me know if I overlooked something.

## Visuals
Kick lanes use "full width" lane visuals like open lanes and wildcard lanes. To address visibility concerns, I've configured all fullwidth lanes to apply a slight emissiveness equivalent to a 1/48th-charged BRE lane. This produces a faint colored background to the highway-wide lane, making it easier to notice.

When using a highway configuration that contains a dedicated kick fret, kick lanes appear as normal lanes.

When using a highway configuration that contains _two_ dedicated kick frets, we always spawn kick lanes for both frets, even if the player isn't playing X+ or if this particular lane has no 2x kicks. It looks weird sometimes, but that's what players who use split kick frets are signing up for, and they can configure their left kick fret to be X+-only if they dislike it.


Assets/Script/Gameplay/Player/DrumsPlayer.cs
```diff
@@ -35,6 +35,11 @@ public class DrumsPlayer : TrackPlayer<DrumsEngine, DrumNote>
         // indistinguishable from 1x kicks by pad number
         public const int DOUBLE_KICK_FRET_INDEX = int.MaxValue;
 
+        private float FRET_ARRAY_PADDING_CORRECTION => 0.97f * 4 / LaneCount;
```
polson: This used to be `5 / laneCount` , double check that this is correct

polson (changes requested):

Assets/Script/Gameplay/Player/DrumsPlayer.cs
Frickitickitavi: Oh yup, good catch - fixed

polson (approved):

PR #1582, opened 2026-07-23
Add guide pitch for vocals practice mode

DevPrice:
This PR adds a guide pitch option to practice mode when a vocals or harmony player is present. When enabled, it will play a pure sine wave matching the active vocals note.

Depends on YARC-Official/YARG.Core#430.

I've mapped toggling the guide pitch state to the vocalist's Orange menu button (matching Rock Band), but happy to consider another approach if one makes sense.

This PR implements a native sine synth DSP, a guide pitch controller that uses it, and some basic UI to expose it in practice mode.

Tested:
* See video for basic case
* Verified it works when changing sections and restarting
* Verified that toggling only cycles through parts in the chart (e.g., skips HARM3 if not present)
* Verified on both solo vocals and harmonies
* Works with pitch slides
* Works across phrase boundaries

https://github.com/user-attachments/assets/fda1b853-c83f-47d9-93c0-29805de0ccb6


Assets/Script/Audio/Effects/GuidePitchSynthDsp.cs
```diff
@@ -0,0 +1,218 @@
...
+{
+    /// <summary>
+    /// A mixer DSP processor that generates a sine wave at the pitch of the currently
+    /// active vocal note. Implements <see cref="IMixerDspProcessor"/> so it can be
+    /// attached to any <see cref="StemMixer"/> without knowledge of the audio backend.
+    /// </summary>
+    public sealed class GuidePitchSynthDsp : IMixerDspProcessor
+    {
+        /// <summary>Target duration for a full volume fade (0 to 1) in seconds.</summary>
+        private const float FADE_DURATION_SECONDS = 0.00145f;
```
polson: 1.45 ms fade?  Not sure if you intended this to be ms or seconds

polson:
Thanks for your PR.  I think this is a worthwhile feature.  I may need to do some adjustments to move the DSP to native C++ soon pending https://github.com/YARC-Official/YARG/pull/1602

Once that is merged I'll test this to see if the DSP still hitches audio on garbage collection, if not it's fine as is, otherwise I will convert to C++ for you

polson:
I get this error in the logs when I turn on the guide pitch and then exit practice: 

```
BassDspHandle: failed to remove DSP: Handle
```

polson (changes requested):

DevPrice:
@polson I split out more of the guide pitch logic from the dsp itself, which hopefully makes any C++ conversion simpler

Assets/Scenes/Gameplay.unity
```diff
@@ -8467,6 +8866,43 @@ Transform:
   m_Children: []
   m_Father: {fileID: 0}
   m_LocalEulerAnglesHint: {x: 50, y: -30, z: 0}
+--- !u!84 &1757749434
```
polson: Seems like unrelated changes to this PR?

polson (changes requested):

Assets/Scenes/Gameplay.unity
DevPrice: To the best of my knowledge all the changes here are for the practice mode UI; but I have no idea how to read these .unity files to verify it.

DevPrice: I think this is fixed now.

polson:
Right now I think the guide pitch is being output to all channels.

For a multichannel setup, I think it should only output to the default channels we set in experimental settings

Can you verify this is working correctly?  Might be hard to test without a real multichannel set up, but I'll probably just accept whatever fix you come up with

Otherwise everything looks good to me

polson:
I also do kind of feel like most people may not be able to figure out how to actually turn on the guide pitch.  Given that it's assigned to orange menu button I do kind of feel like it would be worth putting an orange button icon below the pitch or next to it.  

What do you think?

DevPrice:
@polson addressed your feedback; updated the output channels and add a (clickable) toggle UI
image

polson:
Looks great!

polson (approved):

PR #1584, opened 2026-07-23
Add auto downcharts feature

jstnlef:
This is pure "programmer art" implementation of the feature but it will allow for both the hand authored and generated charts to be accessible in the game simultaneously while we tweak the generation algorithm.

Depends on https://github.com/YARC-Official/YARG.Core/pull/415

### DIFF
--- a/Assets/Script/Gameplay/GameManager.Loading.cs
+++ b/Assets/Script/Gameplay/GameManager.Loading.cs
@@ -318,10 +318,7 @@ namespace YARG.Gameplay
         {
             try
             {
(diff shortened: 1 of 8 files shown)


neilweber:
Hey man, can you generalize this so it can be used for drum charts too?

jstnlef:
> Hey man, can you generalize this so it can be used for drum charts too?

There will be a bunch of follow up work to this including the ability to have downcharting for drums 👍. We'd rather get this in people's hands sooner rather than later such that we can get feedback for tuning.

jstnlef:
It seems like we're not going forward with this :(

PR #1586, opened 2026-07-23
Vocal Note Glow (SP & Normal)

Supersonicboss1:
Demo:

https://github.com/user-attachments/assets/34c9cd25-2213-4ecb-b867-5d19725274f5

This is done through a shadergraph shader. Glow is coloured based on the note's harmony part, or gold if it's in an SP phrase.


wyrdough:
I'm sorry to say that this just makes it look like the notes are out of focus. Perhaps starting the glow at half the intensity might help? Or maybe changing the notes themselves to have a small outline so it looks like the glow is coming from underneath the notes?

Supersonicboss1:
How's this? I made the glow 1/3rd of the strength when not SP:


https://github.com/user-attachments/assets/ed189a69-cd83-4e9b-9233-f0ae870b9120

PR #1591, opened 2026-07-26
Dim the music library when the search field is in focus

poliwhirlge:
Example:
image

Moved the z-order of some ui elements around, hopefully didn't overlap anything incorrectly.
Adding the rounded corners package seems frivolous here but I'll use it in a bunch of other screens soon.


polson:
would you be willing to fix this issue in this PR? https://discord.com/channels/1086048856678084609/1530569410065076334

poliwhirlge:
That issue should be fixed, also fixed a bug where the search box was showing over the filters menu.

PR #1593, opened 2026-07-26
feat(difficulty-select): polish difficulty select menus

XaiaX:
## Summary

Reworks the Quickplay/Practice Difficulty Select menu:

- **Adjustments menu.** The main menu's Modifiers row becomes a single **Adjustments** row opening an intermediate menu that nests the **Modifiers** and (new) **Accessibility** menus. Each nested row previews its active options in its body; the Adjustments row shows a combined list from both, at normal body size for 0–1 active options and header size for 2+. Back/Done step out one level at a time (OpenLane → Modifiers → Adjustments → Main).
- **Accessibility menu.** Groups accessibility-flavored options per game mode: Lefty Flip (new localized toggle — previously only reachable in the profile editor), a unified positive **No Range Shifts** toggle (drives both `profile.RangeEnabled` and, where the mode allows it, the `RangeCompress` modifier; reads active when either mechanism disables shifts), and the relocated modifiers Opens To Greens, No Kicks, and Unpitched Only. These are filtered out of the Modifiers menu (`ACCESSIBILITY_MODIFIERS`). The row is hidden for modes with no entries.
- **Instrument ring row.** The main menu's Instrument item shows every available instrument's `DifficultyRing` (icon + tier wheel, honoring the Classic/Expanded numbered-ring setting) on a centered second row. The selected instrument gets a runtime-generated solid circle backdrop (50% black, 4px rim per side) that rims the wheel and shows through segment gaps; non-selected instruments dim (icon 33%, filled segments 60%). This replaces showing only the current instrument beside its name, which broke with long localized names ("4-Lane Drums" etc.).
- **Instrument menu tier labels.** The instrument sub-menu lists each instrument with a "`5 - Nightmare`"-style tier line (smaller, dimmed) instead of a ring, reusing the filters menu's localized `Menu.Filters.Intensities.*` names (`FiltersMenu.GetIntensityLabelByIndex` made public). Unknown/inactive parts show "`? - Unknown`"; the label clamps at Impossible while the number stays exact. Tier 5 renders in the tier orange `#FF8400`, tier 6+ in the ring-segment red `#FB443F`. Harmony falls back to the lead-vocals tier on solo-only songs.
- **Keys open-note-lane sub-menu.** The five-lane-keys `OpenLaneDisplayType` gets an explicit three-option sub-menu (Never / Always / If chart contains open notes, reusing the profile editor's localized enum names) with a non-interactive title row, replacing a pair of dependent toggles whose phrasing depended on English word order. Summaries list "Open Lane Always" / "Open Lane When Charted"; "Never" counts as off.
- **Blue Done buttons.** Done buttons in the nested menus get the Ready treatment in the menu's pale blue (`#2ED9FF` text, near-black teal `#012227` while highlighted) via a runtime `DifficultyItem.SetAccentColors`, distinguishing them from Ready's green while reading as the same kind of control.

## Why

- The Difficulty Select menu previously gave no indication of any instrument's charted tier, no access to lefty flip without leaving the flow, and mixed accessibility-flavored modifiers into the general Modifiers list.
- Discord feedback: showing a single ring beside the localized instrument name overflows for long names; the two range-shift mechanisms (`RangeEnabled` vs the "No Range Shifts"-named `RangeCompress` modifier) confused users; the dependent open-lane toggle pair didn't localize.
- The Modifiers/Accessibility summary rows dominated the main menu vertically; the Adjustments grouping restores a compact five-row main menu.

## Changes

| File | Change |
|------|--------|
| `Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs` | Adjustments/Accessibility/OpenLane menu states and builders, summary builders, ring row with selection backdrop and dimming, tier labels, Done accenting, per-level back navigation |
| `Assets/Script/Menu/DifficultySelect/DifficultyItem.cs` | `AttachRingRow` (centered ring row as a layout row), `UseSmallBodyText`, `SetAccentColors`, `InitializeAsTitle`, `SetInteractable` (dim/disable plumbing; no caller passes `false` yet) |
| `Assets/Script/Menu/DifficultySelect/ModifierItem.cs` | Net-zero vs `dev` (interim `SetInteractable` added and later removed) |
| `Assets/Script/Menu/MusicLibrary/DifficultyRing.cs` | `SetIconColor`, `SetRingOpacity`, `ShowSelectionBackdrop` with a cached runtime-generated antialiased circle sprite (builtin Knob sprite fails to load, rendering a square) |
| `Assets/Script/Menu/Filters/FiltersMenu.cs` | `GetIntensityLabelByIndex` made public for reuse |
| `Assets/Prefabs/Menu/DifficultySelect/DifficultySelectMenu.prefab` | `DifficultyRing` template instance wired in (kept inactive, used only as clone template); header child order |
| `Assets/StreamingAssets/lang/en-US.json` | New `Menu.DifficultySelect` keys: `Adjustments`, `Accessibility`, `LeftyFlip`, `NoRangeShifts`, `DedicatedOpenLane`, `OpenLaneAlways`, `OpenLaneWhenCharted` |

## Design notes

- All dynamic UI is built from existing prefabs (`DifficultyItem`, `ModifierItem`, the `DifficultyRing` prefab as an inactive in-scene template); the only runtime-generated asset is the backdrop circle texture, cached statically.
- Vocal-modifier uniformity is preserved: the relocated Unpitched Only remains a `Modifier`, so the existing primary-player `CopyModifiers` at confirm still applies it to all vocal players.
- `RangeCompress` is folded into No Range Shifts and filtered from the Modifiers menu; the profile editor's Range Disable stays in sync because the toggle writes `RangeEnabled`.
- A hand-authored `DifficultyItemBlue` prefab variant was tried for the Done styling and dropped — the component reference didn't resolve in-game — in favor of the code-based `SetAccentColors`.

## Testing done

- Unity batchmode compile check (`-batchmode -nographics -quit`, `error CS` scan) run against every commit on the branch; all clean.
- In-game verification 2026-07-18/19 across guitar, drums, vocals, and pro-keys bot profiles: Adjustments/Modifiers/Accessibility nesting and per-level back, combined summary and its dynamic font size, keys open-lane sub-menu including summary labels, instrument ring row (backdrop, dimming, Classic-mode wheels), tier labels, blue Done styling. Screenshot-driven iteration on the backdrop shape/color/contrast and dim levels.
- No Core changes; Core test suite unaffected.

## Notes

- English-only strings added; other languages fall through Crowdin as usual.
- `DifficultyItem.SetInteractable` and the `interactable` parameter on `CreateItem` are plumbing for dimmed fixed choices (e.g. a forced single-chart selection); no current caller passes `false`.
- `FiltersMenu` is upstream (#1314), so the tier-label reuse has no fork-only dependency.


Assets/Script/Menu/DifficultySelect/DifficultyItem.cs
```diff
@@ -31,5 +33,111 @@ public void Initialize(string body, UnityAction action)
...
+                var color = selected ? selectedTextColor : textColor;
+                _header.color = WithAlphaOf(color, _header.color);
+                _body.color = WithAlphaOf(color, _body.color);
+            }
+
+            static Color WithAlphaOf(Color color, Color alphaSource)
+            {
+                color.a = alphaSource.a;
+                return color;
+            }
```
Supersonicboss1: Instead of creating a whole new method for this, use `ColorExtensions.WithAlpha()`, e.g. `_header.color = color.WithAlpha(_header.color.a);`

Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
```diff
@@ -446,43 +589,320 @@ private void CreateDifficultyMenu()
...
+                _menuState = State.Main;
+                UpdateForPlayer();
+            });
+        }
+
+        private void CreateDoneItem(UnityAction action)
+        {
+            CreateItem(LocalizeHeader("Done"), false, action)
+                .SetAccentColors(DONE_TEXT_COLOR, DONE_SELECTED_TEXT_COLOR);
+        }
```
Supersonicboss1, line 687: I think this would be better as it's own prefab, similar to the ready button, rather than creating code to mimic it through a generic menu item. In terms of design, I think it would be better if the entire done button was highlighted in the lighter blue used for the selection outline, similar to the way other special buttons work.

Assets/Script/Menu/MusicLibrary/DifficultyRing.cs
```diff
@@ -163,6 +163,88 @@ public void SetInfo(string assetName, Instrument instrument, PartValues values)
...
+
+        /// <summary>
+        /// Dims the filled ring segments (not the base or icon). Menus use this
+        /// to de-emphasize non-selected instruments. Call after
+        /// <see cref="SetInfo"/>, which resets the segment opacity.
+        /// </summary>
+        public void SetRingOpacity(float alpha)
+        {
+            _ringSprite.color = _ringSprite.color.WithAlpha(alpha);
+        }
```
Supersonicboss1, line 189: This does not appear to work when the difficulty rings are set to expanded

Assets/Script/Menu/MusicLibrary/DifficultyRing.cs
```diff
@@ -163,6 +163,88 @@ public void SetInfo(string assetName, Instrument instrument, PartValues values)
...
+            }
+
+            var texture = new Texture2D(SIZE, SIZE, TextureFormat.RGBA32, false);
+            texture.SetPixels32(pixels);
+            texture.Apply();
+
+            _circleSprite = Sprite.Create(texture, new Rect(0f, 0f, SIZE, SIZE),
+                new Vector2(0.5f, 0.5f));
+            return _circleSprite;
+        }
```
Supersonicboss1: I don't understand the point of this? For one, we do have circle sprites (Art/Menu/Common/Circle for one), but also if you need, I would rather you create one and store it as an asset somewhere rather than creating it with code at runtime

Assets/Script/Menu/Filters/FiltersMenu.cs
```diff
@@ -1659,7 +1659,9 @@ private static string GetIntensityLabel(SongEntry entry, Instrument instrument)
             return GetIntensityLabelByIndex(intensity);
         }
 
-        private static string GetIntensityLabelByIndex(int index)
+        // Public so other menus (e.g. difficulty select) can reuse the same
+        // intensity-tier naming.
+        public static string GetIntensityLabelByIndex(int index)
```
Supersonicboss1: I would prefer if functions like these were extracted to a helper class rather than reaching into the filters menu to do this.

Assets/Script/Menu/DifficultySelect/DifficultyItem.cs
```diff
@@ -31,5 +33,111 @@ public void Initialize(string body, UnityAction action)
...
+            var group = GetComponent<CanvasGroup>();
+            if (group == null)
+            {
+                group = gameObject.AddComponent<CanvasGroup>();
+            }
+
+            group.alpha = interactable ? 1f : 0.3f;
+            group.interactable = interactable;
+            group.blocksRaycasts = interactable;
+        }
```
Supersonicboss1, line 105: I would prefer if you just added a CanvasGroup to the prefab.

Assets/Script/Menu/DifficultySelect/DifficultyItem.cs
```diff
@@ -31,5 +33,111 @@ public void Initialize(string body, UnityAction action)
...
+                rt.sizeDelta = new Vector2(65f, 65f);
+                rt.localScale = Vector3.one * (size / 65f);
+                rt.anchoredPosition = new Vector2((i - (count - 1) * 0.5f) * (size + spacing), 0f);
+
+                ring.gameObject.SetActive(true);
+                rings[i] = ring;
+            }
+
+            return rings;
+        }
```
Supersonicboss1: Rather than doing this, I think it would be better to create a prefab variant of DifficultyItem, extend DifficultyItem with a new class, and add a new CreateItem overload. Feel free to push back if you disagree, or it isn't possible for some reason.

Assets/Script/Menu/DifficultySelect/DifficultyItem.cs
```diff
@@ -31,5 +33,111 @@ public void Initialize(string body, UnityAction action)
             _body.text = body;
             Button.SetOnClickEvent(action);
         }
+
+        /// <summary>
+        /// Tints the header/body text, switching to a darker accent while the
+        /// item is selected (the same treatment DifficultyItemGreen gets from
+        /// its NavigationTextColorizer, but applied at runtime). Each text's
+        /// own alpha is preserved.
```
Supersonicboss1: Not going to add this to every comment, but could you go through the comments in this PR and make them a bit more concise? e.g. pretty much the whole first sentence is kind of unnecessary imo.

Supersonicboss1 (changes requested):
Sorry about the number of comments lol - this is a very good change in terms of UX, just needs some code polish.

XaiaX:
Ok, I've updated the PR:

**`WithAlphaOf`:** removed; using `ColorExtensions.WithAlpha()` as suggested.

**`CanvasGroup`:** added to the `DifficultyItem` prefab; the runtime
`AddComponent` is gone.

**Expanded rings bug:** root cause: in expanded mode the *remainder* arc
carries the visible color (white at 6–10, red at 11–15), and
`SetRingOpacity` only dimmed the filled arc, so the dimming read as a no-op.
It now dims both arcs. While testing I also made the row's de-emphasis
focus-aware: when another field has focus (black row background),
non-selected rings dim deeper and the selected-instrument backdrop hides —
it's translucent black blended for the blue highlight, so it was invisible
on the black row, anyway.

**Circle sprite:** runtime generation deleted; the backdrop now uses a
serialized sprite wired to `Art/Menu/Common/Circle`.

**Intensity labels:** extracted to `YARG.Helpers.IntensityLabels` (keys +
lookup, read-only surface); `FiltersMenu` and difficulty select both use
it, no more reaching across menus.

**Done button:** now its own prefab, but deliberately generic:
`DifficultyItemColored` (a `DifficultyItem` variant) plus a
`DifficultyItemColorizer` class that takes an accent pair at runtime, so
it's parameterized rather than another per-color prefab like
`DifficultyItemGreen`/`Red` (each of which bakes its text colors, a
`NavigationTextColorizer`, and a per-color background PNG). One design
note: I went with the Ready-button treatment in blue (blue text normally,
near-black blue text on the highlight while selected) rather than a literal
full blue fill — the green fill is a baked sprite, and a full fill needs a
baked `DifficultyItemBlue.png` sibling under the current per-color
approach. If we add a white row sprite the selected-state fill
becomes a plain runtime tint, then the green/red items could
also migrate onto the same generic variant and drop their per-color
prefabs entirely.

**Ring row / prefab variant:**  as suggested: `DifficultyItemRings`
class on its own `DifficultyItem` prefab variant, plus a `CreateItem`
overload; `AttachRingRow` is removed from `DifficultyItem` itself. (The
class sits alongside the item rather than subclassing it, so the base
stays free of ring concerns either way.)

**Comments:** tightened them throughout.

Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
XaiaX: I kind of split the difference on this, and built a more generic prefab that we could re-use if in the future we wanted to add something like "Revert" or "Defaults" or whatever that was yellow or orange. Basically it's just a "this is a slightly special entry in the difficulty menu" prefab that takes some parameters for custom colors.  If we wanted we could handle ready/sit out/disconnect the same way, but the latter two are also slightly differently sized and not just "red". The color is being pulled from MenuData.Colors.NavigationBlue, so if that specific blue ever changes, it'll match.

Assets/Script/Menu/MusicLibrary/DifficultyRing.cs
XaiaX: Yes, there's another "remainder" wheel for the bits that stay white or stay red, fixed that and also tweaked the values to make them contrast a little more, at least on my display.

Assets/Script/Menu/MusicLibrary/DifficultyRing.cs
XaiaX: Probably just SwiftUI brain, I'm too use to constructing shapes programmatically vs. using existing assets. Fixed.

Assets/Script/Menu/DifficultySelect/DifficultySelectMenu.cs
Supersonicboss1: I don't really have an issue with this direction (though I think it should probably be generalised, and we could have two versions, one small and one regular, but you don't need to do that in this PR). However, I have some nitpicks about the art direction. My idea is to not use a sprite for the selection on these special buttons, and instead allow a color to be passed in which acts as the background color, and fills the entire button, similar to the ready/sit out buttons.
e.g. Unselected:
image
Selected:
image

The color being used here is #00D3FF.

polson:
image

open note lane should not highlight with mouse

polson (approved):
Code looks fine, fix the one issue

Assets/Script/Menu/DifficultySelect/DifficultyItem.cs
```diff
@@ -31,5 +33,55 @@ public void Initialize(string body, UnityAction action)
...
+        {
+            _body.fontSize = _header.fontSize;
+        }
+
+        /// <summary>
+        /// Dims the item and disables interaction (used to show a fixed, non-editable choice).
+        /// </summary>
+        public void SetInteractable(bool interactable)
+        {
+            var group = GetComponent<CanvasGroup>();
```
Supersonicboss1: In general (and here), avoid `GetComponent` outside of init functions, and add a `SerializedField` instead.

Assets/Script/Menu/DifficultySelect/DifficultyItem.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using TMPro;
 using UnityEngine;
 using UnityEngine.Events;
+using UnityEngine.UI;
```
Supersonicboss1, line 4: Unused import here

Assets/Script/Menu/DifficultySelect/DifficultyItemColorizer.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class DifficultyItemColorizer : MonoBehaviour
+    {
+        private DifficultyItem _item;
+
+        /// <summary>
+        /// Applies the accent text and selected text colors.
+        /// </summary>
+        public void SetButtonColor(Color accentColor, Color selectedAccentColor)
+        {
+            _item ??= GetComponent<DifficultyItem>();
```
Supersonicboss1, line 22: This is only called when creating the button, so I don't really care whether you replace this with a `SerializedField` or not, but I thought i'd bring it to your attention in case you prefer one way.

Supersonicboss1 (changes requested):
Sorry, couple more nitpicks, then we should be good to go code-wise (along with Bort's thing). Also, let me know what you think about the design stuff I posted above on one of the review comments (or implement them if you feel they're good).

XaiaX:
Done menu modified per @Supersonicboss1 suggestion.
Screenshot 2026-08-29 at 12 56 33

Mouse isn't showing in the screenshot, but it's no longer highlighting the header bit, per @polson 
Screenshot 2026-08-29 at 12 56 47

Assets/Script/Menu/DifficultySelect/DifficultyItem.cs
XaiaX: no longer unused, needed for _selectionFill

XaiaX:
Added back the "Unknown" label for bad difficulties that I missed when merging from upstream after the last update.

polson (approved):

XaiaX:
Fixed merge conflicts and validated that this works with the new 6 fret game mode.

PR #1600, opened 2026-08-01
Enable Free Harmonies gameplay (minimal, single-mic)

XaiaX:
Depends on [Add Free Harmonies: multi-part vocals engine + ...](https://github.com/YARC-Official/YARG.Core/pull/437)
## Summary

Wires up Free Harmonies as a selectable vocals instrument. When a player chooses "Free Harmonies" on a song with vocals, the game constructs a `PartyVocalsCoordinatorEngine` instead of `YargVocalsEngine`. The coordinator matches the mic's pitch against ALL HARM parts simultaneously and grades phrases based on how many parts were satisfied.

For solo-only songs, the coordinator runs against the solo vocal chart as a single-part track — same engine, one part to match.

## Why

The Core PR adds the engines but nothing in the game constructs them. This PR is the smallest change that makes Free Harmonies playable, so the multi-part matching and phrase scoring can be validated before building dedicated harmony meters and multi-mic support.

## Changes

| File | Change |
|------|--------|
| `Gameplay/GameManager.Loading.cs` | Vocal track visual uses `Chart.GetVocalsTrack()` so PartyVocals gets the correct track on any song. |
| `Gameplay/Player/VocalsPlayer.cs` | `CreateEngine()`: construct `PartyVocalsCoordinatorEngine` when `CurrentInstrument == PartyVocals`. Clones and modifies ALL vocal parts (not just primary) so modifiers apply to every HARM line. Forwards sub-engine visual events (needle, sing, hit) to the coordinator. Particle trail colors update dynamically via `DisplayedHarmonyIndex`. |
| `Menu/DifficultySelect/DifficultySelectMenu.cs` | PartyVocals added to vocals mode-matching and difficulty checks. |
| `Gameplay/Player/Vocals/VocalTrack.cs` | Multi-lane lyric display for PartyVocals. |
| `Gameplay/HUD/UnisonBonus/UnisonDisplay.cs` | PartyVocals counted as vocals for unison tracking. |
| `Gameplay/HUD/InstrumentIconProvider.cs` | PartyVocals uses harmony icons and colors. |
| `Helpers/Extensions/InstrumentExtensions.cs` | `ToSortAttribute` and `ToResourceName` handle PartyVocals. |
| `StreamingAssets/lang/en-US.json` | `Enum.Instrument.PartyVocals` = "Free Harmonies". |
| `YARG.Core` (submodule) | Pinned to `feat/party-vocals-coordinator`. |

## How it works

- Player selects "Free Harmonies" in the difficulty select menu (appears alongside Vocals and Harmony for any song with vocals).
- `VocalsPlayer.CreateEngine()` constructs the coordinator with the harmony chart's parts (or solo chart for songs without harmony).
- The coordinator runs its sub-engine lifecycle internally, matching the mic's pitch against all HARM parts simultaneously.
- Phrase scoring uses the coordinator's ambiguity allocation and `PhraseGrade` mapping.
- The existing vocals HUD works unchanged — needle, particles, phrase meter, lyrics, percussion all display correctly. Particle trail color changes dynamically to match the currently-sung harmony part.
- No dedicated harmony meters or phrase-grade banners yet — those are future work.

## Testing done

- Core: 512 unit tests pass (473 upstream + 39 new).
- Verified in-game with both live microphone and bot profile: needle tracks pitch, particle trail appears and changes color when switching parts, lyrics display, phrase meter fills correctly, percussion works, star power activates, practice restart works, modifiers (UnpitchedOnly, NoVocalPercussion) behave correctly, replay analysis matches live results (no "inconsistent replay" warning).
- Tested on both harmony charts and solo-only vocals songs.

## Notes

- **Single mic only.** Multi-mic requires input routing and a dedicated player class — future work.
- **No dedicated Free Harmonies HUD.** The coordinator emits `OnPartyVocalsPhrase` (phrase grades, per-part results) but nothing subscribes yet. The existing vocals HUD is used as-is.
- **Profile serialization.** `CurrentInstrument` is serialized as a byte, so `PartyVocals = 42` persists without a version bump.

### DIFF
--- a/Assets/Script/Gameplay/GameManager.Loading.cs
+++ b/Assets/Script/Gameplay/GameManager.Loading.cs
@@ -526,11 +526,10 @@ namespace YARG.Gameplay
                             VocalTrack.transform.position = new Vector3(highwayIndex * TRACK_SPACING_X, 100, 0);
                             _trackViewManager.CreateVocalTrackView(highwayIndex);
 
-                            // Since all players have to select the same vocals
-                            // type (solo/harmony) this works no problem.
-                            var chart = player.Profile.CurrentInstrument == Instrument.Vocals
-                                ? Chart.Vocals
-                                : Chart.Harmony;
+                            // Use GetVocalsTrack so PartyVocals gets the correct track
+                            // (harmony when available, solo vocals fallback for songs
+                            // with no harmony parts).
+                            var chart = Chart.GetVocalsTrack(player.Profile.CurrentInstrument);
                             VocalTrack.Initialize(chart, player, Song.VocalScrollSpeedScalingFactor);
 
                             if (SettingsManager.Settings.KeepLyricBar.Value &&
--- a/Assets/Script/Gameplay/HUD/FailMeter.cs
+++ b/Assets/Script/Gameplay/HUD/FailMeter.cs
@@ -152,11 +152,11 @@ namespace YARG.Gameplay.HUD
                 _needleSliders[i].handleRect.DOAnchorPosX(SPRITE_INITIAL_OFFSET, 0.125f).SetAutoKill(false).SetLink(_needleSliders[i].gameObject);
 
                 var handleImage = _playerSliders[i].handleRect.GetComponentInChildren<Image>();
-                var spriteName = _players[i].GetInstrumentSprite();
+                var spriteName = _players[i].GetInstrumentSprite(_gameManager.Song);
 
                 var sprite = Addressables.LoadAssetAsync<Sprite>(spriteName).WaitForCompletion();
                 handleImage.sprite = sprite;
-                handleImage.color = _players[i].GetHarmonyColor();
+                handleImage.color = _players[i].GetGameplayIconColor();
 
                 _playerSliders[i].value = 0.01f;
                 _needleSliders[i].value = 0.01f;
--- a/Assets/Script/Gameplay/HUD/InstrumentIconProvider.cs
+++ b/Assets/Script/Gameplay/HUD/InstrumentIconProvider.cs
@@ -1,7 +1,9 @@
 ﻿using UnityEngine;
 using YARG.Core;
 using YARG.Core.Engine;
+using YARG.Core.Engine.Vocals.Engines;
 using YARG.Core.Logging;
+using YARG.Core.Song;
 using YARG.Gameplay.Player;
 using YARG.Helpers.Extensions;
 using YARG.Player;
@@ -21,6 +23,36 @@ namespace YARG.Gameplay.HUD
             return GetInstrumentSprite(player.Profile.CurrentInstrument, player.Profile.HarmonyIndex, isMissingDevice);
         }
 
+        // A party vocals container is built on the harmony chart's parts, so its
+        // instrument reports as Harmony; identify it by its coordinator engine instead.
+        private static bool IsPartyVocals(this EngineManager.EngineContainer container)
+        {
+            return container.BaseEngine is PartyVocalsCoordinatorEngine;
+        }
+
+        // Song-aware overloads for gameplay HUDs (player banner, fail meter):
+        // party vocals is part-agnostic, so its icon reflects the song's vocal
+        // part count instead of the player's harmony index.
+        public static string GetInstrumentSprite(this EngineManager.EngineContainer container, SongEntry song)
+        {
+            if (song != null && container.IsPartyVocals())
+            {
+                return GetPartyVocalsSprite(song.VocalsCount);
+            }
+
+            return GetInstrumentSprite(container);
+        }
+
+        public static string GetInstrumentSprite(this YargPlayer player, SongEntry song)
+        {
+            if (song != null && player.Profile.CurrentInstrument == Instrument.PartyVocals)
+            {
+                return GetPartyVocalsSprite(song.VocalsCount);
+            }
+
+            return GetInstrumentSprite(player);
+        }
+
         private static string GetInstrumentSprite(Instrument instrument, int harmonyIndex, bool isMissingDevice)
         {
             if (isMissingDevice)
@@ -36,6 +68,17 @@ namespace YARG.Gameplay.HUD
             return $"InstrumentIcons[{instrument.ToResourceName()}]";
         }
 
+        private static string GetPartyVocalsSprite(int vocalsCount)
+        {
+            // Part-count based icon: solo mic, two mics, or harmony mics
+            return vocalsCount switch
+            {
+                >= 3 => "InstrumentIcons[harmVocals]",
+                2 => "InstrumentIcons[twoVocals]",
+                _ => "InstrumentIcons[vocals]",
+            };
+        }
+
         public static Color GetHarmonyColor(this YargPlayer player)
         {
             return GetHarmonyColor(player.Profile.CurrentInstrument, player.Profile.HarmonyIndex, player.IsMissingInputDevice || player.IsMissingMicrophone, player.SittingOut);
@@ -44,6 +87,27 @@ namespace YARG.Gameplay.HUD
         public static Color GetHarmonyColor(this EngineManager.EngineContainer container)
         {
             return GetHarmonyColor(container.Instrument, container.HarmonyIndex, false, false);
+        }
+
+        // Color for gameplay HUD icons: party vocals is part-agnostic, so it stays untinted
+        public static Color GetGameplayIconColor(this YargPlayer player)
+        {
+            if (player.Profile.CurrentInstrument == Instrument.PartyVocals)
+            {
+                return Color.white;
+            }
+
+            return GetHarmonyColor(player);
+        }
+
+        public static Color GetGameplayIconColor(this EngineManager.EngineContainer container)
+        {
+            if (container.IsPartyVocals())
+            {
+                return Color.white;
+            }
+
+            return GetHarmonyColor(container);
         }
 
         private static Color GetHarmonyColor(Instrument instrument, int harmonyIndex, bool isMissingDevice, bool isSittingOut)
--- a/Assets/Script/Gameplay/HUD/InstrumentIconProvider.cs
+++ b/Assets/Script/Gameplay/HUD/InstrumentIconProvider.cs
@@ -28,7 +28,7 @@ namespace YARG.Gameplay.HUD
                 return $"NoInstrumentIcons[{instrument.ToResourceName()}]";
             }
 
-            if (instrument == Instrument.Harmony)
+            if (instrument == Instrument.Harmony || instrument == Instrument.PartyVocals)
             {
                 return $"HarmonyVocalsIcons[{harmonyIndex + 1}]";
             }
@@ -57,7 +57,7 @@ namespace YARG.Gameplay.HUD
             {
                 return Color.gray;
             }
-            if (instrument != Instrument.Harmony)
+            if (instrument != Instrument.Harmony && instrument != Instrument.PartyVocals)
             {
                 return Color.white;
             }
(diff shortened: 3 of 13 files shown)


XaiaX:
Discovered an edge case with songs where there are phrases that are only harm2/harm3, and getting weird results in the summary like "SP Phrases 7/5", and visually registering notes when there is no harm1 part but not counting towards or against the combo meter/phrase count, so I'm going to fix those real quick.

XaiaX:
The Free Harmonies option showing up in the instrument select for a vocals profile.
Screenshot 2026-08-29 at 13 46 23

What it looks like when Free Harmonies is selected.
Screenshot 2026-08-29 at 13 46 39

Bot profile singing the Harmony 1 part as normal.
Screenshot 2026-08-29 at 13 47 09

Bot profile automatically singing the Harmony 2 part when there's no Harmony 1 part available:
Screenshot 2026-08-29 at 13 47 36

XaiaX:
The combo meter icon at the start of a song now reflects how many parts are charted, and also in the fail meter
3 parts:
3-parts

2 parts:
2-parts

1 part: (identical to "solo", for completeness)
1-part

XaiaX:
There are a couple small bugs/annoyances still present here which I have identified but I think the fix will be too large to keep in this same PR. The vocal needles are a bit "sticky" towards the lowest index harmony available, so it is hard to sing harm2 or harm3 if there is harm1 or harm2 present, respectively.  Separately there's an issue allocating to notes that overlap and especially when there's a short phrase that constitutes the entirety of a vocal phrase for one harmony. E.g., just a "yeah!" or something like that while a long note sequence is going. So it may produce some confusion or frustration for players that notice it, but overall it's not a blocker to the main feature. The best test case for this is songs where there are distinct alternating / call and response parts, where now a vocalist can sing any part of that they wish.

I'll work on the behavioral change either as a separate patch for this, or roll it into the next stage if it makes sense.

polson:
Can you check this?

#### Practice mode guide pitch controls do not support Free Harmonies
When entering Practice mode on Free Harmonies, the guide pitch HUD indicator is hidden and the guide pitch toggle shortcut does not work to cycle through harmony parts.
1. Select a song with harmony vocals and choose Free Harmonies in the difficulty select menu.
2. Enter Practice mode.
3. Observe that the Guide Pitch HUD widget is missing from the practice overlay.
4. Press the guide pitch shortcut button and observe that no guide tone activates.
**Expected:** The guide pitch HUD indicator and toggle shortcut should function during Free Harmonies practice sessions.
**Actual:** The guide pitch display is hidden and the toggle shortcut is ignored.

XaiaX:
Ok, it works in practice now, had to get the practice mode to recognize Free Harmonies as a separate instrument. It shows up with the harmonies UI (switch between all available pitches) when harmonies are available, and falls back to the solo UI (just on/off) if the chart is solo-only.

--edit--
Also synced my .Core to master for the other part of this PR with no merge conflicts, so no changes there. other than where the core pointer is, now.

polson:
I see two harmony sorts

image

polson:
Another issue:

1. Play a song with solo vocals, but select free harmonies
2. Go back
2. In the song library, sort by Stars, Percentage, or Score.

Songs with a solo vocals part show up as NO PART

Seems wrong?

image

polson (changes requested):
See comments

XaiaX:
Were you testing this in a branch with https://github.com/YARC-Official/YARG.Core/pull/437 already merged in? (I haven't updated this PR for a bit because I was waiting on the other, but I'll update this in the morning in case you're looking at them together)

XaiaX:
I reproduced that "no part" issue on current nightly, I think it's been doing that for a while. I even started with a song that has harmonies and just quit the song as soon as it started, so I think that's just a gap in the existing sorting.

The difficulty/instrument select shows the difficulty wheel now.

Other things I fixed/added while addressing merge conflicts:

1. **Harmony-less song presentation:** Free Harmonies uses the lead-vocals tier and a one-part ring when no harmony chart exists, rather than presenting unavailable harmony-part information. (It was showing the difficulty tier as "? - Unknown" rather than using the solo tiering. There's not much point to using free harmonies on a solo vocals song right now, but when multiple mics on one profile are added it will matter)
2. **Practice readability:** with Guide Pitch enabled, unfocused harmony tubes, notes, and talkies are dimmed while the currently focused part remains clear. Example video/screenshot in [this](https://discord.com/channels/1086048856678084609/1547661573232009346) discord thread.
3. **Needle feedback:** the vocal needle outline adopts the color of the harmony part currently being matched. Per Leo's request in same discord thread, with another example video.
4. **Coordinator behavior dependency:** pins Core `71b8a3bc`, which broadcasts multi-part talkie matches and retains the previous target on ambiguous pitch ties. (handles ambiguity when talkies overlap and should prevent switching vocal lines when they intersect by sticking to the line the user was previously singing.
5. **Engine Preset editor:** adds **Vocals** to the Engine Preset preview-instrument selector. Vocal fields render; its unsupported static preview is intentionally blank. Switching to a tab that does not support Vocals resets the shared preview mode to guitar before construction, so its label and live preview remain synchronized immediately. Fixing a regression from the preset preview feature, but since it is directly vocals related and relatively small it fits here.

Vocals option appearing in Engine Presets:
Screenshot 2026-09-16 at 00 28 06

Vocals option not appearing in other presets:
Screenshot 2026-09-16 at 00 28 24

polson:
I still see two harmony sorts


CleanShot 2026-09-17 at 12 27 25

polson (changes requested):
See comments

XaiaX:
Oooh, I thought that was in the difficulty select, didn't notice it was in the "sort by" menu. Derp.

XaiaX:
The existing sorting issue is just that "Vocals" and "Harmony" are different instruments as far as YARG is concerned, so when sorting while "Harmony" is active if a song has only a solo part then for the harmony it has `NoPart`, which is why it doesn't do the reverse, since every song with a harmony also has a solo part.

We may want to rethink that or come up with one like "aggregate drums" that puts the scores together, but the old discrete harmony system only scores one single harmony part so those scores will generally not be directly analogous to free harmony particularly when a song has phrases that only contain a subset of harmonies, so the total number of phrases will be greater when the player can sing any part.

For now, I just added a label for Free Harmony to show up as a third sort option which will cover all songs with vocals since it falls back to solo when there is no harmony present. (If it has both, it will use the difficulty tier from harmony) Score and Stars will be tied to Free Harmony so will only be displayed if that's the active "instrument" but if a player hasn't played a song on free harmony it does *not* use their solo or harmony score. (We could tie that to the aggregate vocals). But, that is already the case with the existing vocal scores, so it's consistent.

It's also a little unclear if we actually want to do that, since we talked about maybe just replacing the existing vocals game mode with the free harmony mode with multiple mics, which will treat all vocal scores/stars/plays as the same, though the sorting for difficulty tier will still follow the harmonies-to-solo fallback path in that event.

polson:
> [!NOTE]
> *This review comment and reproduction report was generated with AI assistance.*

Found a few issues that need fixing before merge. Three have reproduction videos below.

#### 1. `.gitmodules` points at a private SSH alias and breaks clean checkouts

`.gitmodules` line 3 sets the `YARG.Core` submodule to `git@github-xaiax:XaiaX/YARG.Core.git`. `github-xaiax` is a personal SSH host alias that only exists on the author's machine, so nobody else can resolve it.

1. Clone this branch on a clean machine without that SSH alias.
2. Run `git submodule update --init --recursive`.
3. Git fails resolving host `github-xaiax` and aborts, so the game cannot be built.

Expected: `.gitmodules` uses the public `https://github.com/YARC-Official/YARG.Core`.
Actual: Points at a private SSH fork address; inferred from code, not run as a clean clone.
Fix: Revert `.gitmodules` to the public HTTPS URL.
Video: none (environment issue, no gameplay to record).

#### 2. Music list crashes with only computer players

Opening the music list with no human players makes the list fail.

1. Add only computer-controlled players with no human players.
2. Open the music library.
3. The library fails to show songs.

Expected: The music library shows normally with only computer players.
Actual: The library throws and fails to render.

https://github.com/user-attachments/assets/2d48d9a3-5e55-4410-8286-bd87db881802

#### 3. New sorting hides songs and shows raw numbers

Choosing the new harmony sorting shows raw numbers and hides songs without singing.

1. Open the music library.
2. Choose the new harmony sorting.
3. Headers show raw bracketed numbers and songs without singing are missing.

Expected: Normal tier names show and all songs appear grouped, including songs without singing.
Actual: Raw numbers show and instrumental songs are missing.

https://github.com/user-attachments/assets/0c57a589-1818-454e-a5e8-c927555dc47d

#### 4. Extra glow breaks glowing notes

Turning on the bonus glow for singing notes makes the glow shape disappear.

1. Start a song with singing notes.
2. Trigger the bonus glow during singing.
3. The glow shape around notes disappears.

Expected: The glow keeps its shape when bonus glow turns on or off.
Actual: The glow shape resets.

https://github.com/user-attachments/assets/f2921f30-1f71-498a-9639-2c3d3f78f432

XaiaX:
### Fixed issues ###

**Public Core submodule URL:**
- Restored the committed URL to the public canonical endpoint: `https://github.com/YARC-Official/    YARG.Core`

**Music Library failure with bot-only rosters:**
- `GetPreferredHighScoresForCurrentPlayers()` assumed at least one non-bot player and used `First(...)`.
- With only bots, refreshing song score information threw `InvalidOperationException: Sequence contains no matching element`, making the library unusable.
- Changed this to handle the no-human case and use the existing band/empty-score display path.
- Verified in the Editor with bot-only players: the library can now be entered and navigated normally.

**Free Harmony sort labels and omitted songs:**
- Removed custom Free Harmony sort (it was there to combine Vocals/Harmony into a single order with fallback from Harmony to Vocals) and replaced it with the normal localized intensity-category labels and a localized No Part category.
- The intended fallback remains: use Harmony tier when present, otherwise Vocals tier, otherwise No Part.
- Verified in the Editor.

**Star Power vocal-note glow shape:**
- Applying the Star Power glow color replaced the renderer’s material property block without preserving `_Dimensions`, so the glow reverted to shader defaults rather than matching the note’s dimensions.
- Read the renderer’s existing property block before updating glow color/intensity, preserving `_Dimensions`.
- Verified in the Editor.

**Party Vocals score/star behavior:**
- Removed an unintended Party Vocals exception that showed individual scores/stars for multiplayer Party Vocals sessions.
- Party Vocals now follows existing YARG semantics: one human uses individual records; multiple humans use the existing song-level band record and band stars.
- This does not change the Harmony → Vocals chart/tier fallback behavior.

### Validation ###

- Resolved and committed the `dev` merge conflicts.
- Updated the paired Core pin to the current coordinator branch commit.
- Unity 6000.3.5f2 batchmode compiled and editor-tested the bot-only Music Library, Free Harmony sort behavior, and vocal glow behavior.

polson:
> [!NOTE]
> *This review comment and reproduction report was generated with AI assistance.*

### Free Harmony search omits solo-vocal songs

Free Harmony search drops solo-vocal songs that gameplay can play through its lead fallback. In the library used for this reproduction, `Jenny B` appears in a title search, but adding the Free Harmony filter returns no matches for it.

1. In Quick Play > Music Library, search for `name:"Jenny B"` and confirm the song appears.
2. Add the Free Harmony filter so the search reads `freeharmony:;name:"Jenny B"`.

Expected: `Jenny B` remains available in the Free Harmony search through its lead-vocal fallback.

Actual: No songs match the criteria.

https://github.com/user-attachments/assets/9a8f87d3-a8d0-4945-99f2-0fc0016b367a

XaiaX:
#### Changes
- Fixed Free Harmony filtering for solo-vocal songs. The `freeharmony:` search was using a harmony-only song index, so songs playable through Free Harmony’s lead-vocal fallback disappeared from search results. It now includes those songs and uses harmony intensity when available, falling back to lead-vocal intensity otherwise.
- A follow-up audit found the same mismatch in the profile-scoped Intensity filter: solo-vocal songs were counted as “No Part” instead of under their playable Free Harmony intensity. That filter now follows the same harmony-first, lead-fallback rule. Harmony-only filtering remains unchanged. Easiest way to test this is to go to the filters, deselect all, and then select "no part" as the only filter, and you should only see instrumentals.

PR #1606, opened 2026-08-04
Fix Persistent UI Layering Regression

Dansla116:
With #1601, there is a regression that the `HelpBar` will appear in-front of the `DifficultiesContainer`, cutting off the bottom row of difficulty rings.

Because `DifficultiesContainer` needs to be in-front of `HelpBar`, and `HelpBar` needs to be in-front of `FiltersMenu` and `FiltersMenu` needs to be in-front of `DifficultiesContainer`, I've decided to hide the `DifficultiesContainer` when you open the `FiltersMenu` and show `DifficultiesContainer` when you close the `FiltersMenu`.


Supersonicboss1 (approved):
Seems to fix it. I did notice there's a flicker of the filters menu when going filters menu -> music library, but that doesn't seem to be caused by this change.

PR #1609, opened 2026-08-05
Support video playback via libVLC (with fallback to built in player)

theli-ua:
Only implemented for Linux/OpenGL in this PR
Widely extends range of supported video codecs and containers.
This does not bundle vlc itself and will only work if host system has system-wide libvlc 4.x installed.

Bundles small pre-built shared object for Linux/x86_64 as it looks like the established pattern (Looking at current audio related libraries). As well as pre-built libvlcsharp assembly (although would be possible to build with `dotnet` if required obviously)

Unfortunately upstream vlc-unity does not provide package.json so it cannot be added as a unity dependency so its integrated a bit awkwardly, git submodule in vlc/ and symlink of source files + asmdef into Assets/

Similarly there is a submodule for libvlcsharp in vlc/ , however its just there to indicate which commit was the libvlcsharp built from, or in case we want to build it as part of actual yarg build.


theli-ua:
If this ever would be merged I'd be interested to try and get vlc-unity support vulkan on linux

polson (approved):

theli-ua:
Build from this branch https://github.com/theli-ua/YARG-BleedingEdge/releases/tag/vlc-2026-08-07

wyrdough:
Can confirm this builds on the github runner, at least...

theli-ua:
> Can confirm this builds on the github runner, at least...

Yeah, I've included such a build right above your reply! (it was made to test github runner)

Assets/Script/Gameplay/BackgroundManager.cs
```diff
@@ -459,7 +457,7 @@ private void SetUpVideoTexture(BackgroundResult songBackGround)
                     //set venue source to song to enable video seeking/pausing features
                     _source = VenueSource.Song;
                     //set up videoPlayer to render to venue texture
-                    _videoPlayer.renderMode = VideoRenderMode.RenderTexture;
+                    //_videoPlayer.renderMode = VideoRenderMode.RenderTexture;
```
polson: Uncomment or remove?

theli-ua:
```suggestion
```

theli-ua: done

Assets/Script/Gameplay/YargVideoPlayer.cs
```diff
@@ -0,0 +1,347 @@
+#if !(UNITY_EDITOR_WIN || UNITY_STANDALONE_WIN || UNITY_WSA)
+#define VLC_SUPPORTED
```
polson, line 2: Should this be linux only with a #if ?

Assets/Script/Gameplay/YargVideoPlayer.cs
```diff
@@ -0,0 +1,347 @@
...
+{
+    [SerializeField] private VideoPlayer _unityVideoPlayer;
+
+#if VLC_SUPPORTED
+    [SerializeField] private LibVLCSharp.VLCMediaPlayer _vlcPlayer;
+    private bool _usingVLC = false;
+    private bool _vlcPrepared = false;
+    private bool _prepareCalled = false;
+    private float _prepareStartTime = 0f;
+    private const float VLC_PREPARE_TIMEOUT = 5f;
```
polson: I think this isn't used

Assets/Script/Gameplay/YargVideoPlayer.cs
```diff
@@ -0,0 +1,347 @@
...
+            else if (_unityVideoPlayer != null)
+                _unityVideoPlayer.playbackSpeed = value;
+#else
+            if (_unityVideoPlayer != null)
+                _unityVideoPlayer.playbackSpeed = value;
+#endif
+        }
+    }
+
+    public bool isLooping
```
polson, line 145: Looks unused

Assets/Script/Gameplay/YargVideoPlayer.cs
```diff
@@ -0,0 +1,347 @@
...
+            _usingVLC = false;
+            if (_vlcPlayer != null)
+            {
+                _vlcPlayer = null;
+            }
+            SwitchToVideoPlayerFallback();
+        }
+    }
+
+    private void OnVLCTextureResized(RenderTexture texture)
```
polson, line 302: I think this means video resizing causes the player to stop, is that ok?

Assets/Script/Gameplay/TextureManager.cs
```diff
@@ -35,6 +36,8 @@ public class TextureManager : GameplayBehaviour
 
         private bool _videoTexFound = false;
 
+        private List<Material> _videoMaterials = new List<Material>();
```
polson, line 39: Looks unused

Assets/Plugins/vlc/Linux/x86_64/libVLCUnityPlugin.so.meta
```diff
@@ -0,0 +1,50 @@
...
+      settings:
+        CPU: AnyCPU
+        DefaultValueInitialized: true
+        OS: AnyOS
+    Linux64:
+      enabled: 1
+      settings:
+        CPU: x86_64
+    OSXUniversal:
+      enabled: 0
```
polson, line 37: Should be linux only I think

polson (changes requested):
I get this on Windows:

DllNotFoundException: libVLCUnityPlugin assembly:<unknown assembly> type:<unknown type> member:(null)
LibVLCSharp.OnLoad.OnBeforeSceneLoadRuntimeMethod () (at Assets/VLCUnity/Internal/OnLoad.cs:56)

Assets/Script/Gameplay/YargVideoPlayer.cs
theli-ua: it should theoretically work on any platform although I don't have a way to test and debug build issues on windows which is why I disabled it.

Assets/Script/Gameplay/YargVideoPlayer.cs
theli-ua: Indeed

Assets/Script/Gameplay/TextureManager.cs
theli-ua: This is where it stores video materials so it can update them to point to correct video texture if it changes, it is used in

```csharp
        private void UpdateVideoMaterials()
        {
            foreach (var m in _videoMaterials)
            {
                m.SetTexture(_videoTexId, _videoTexture);
            }
        }
```

however you are right that I forgot to populate that one, let me fix that.

theli-ua:
> OnBeforeSceneLoadRuntimeMethod

I'll try to disable libvlcsharp for everything but linux for now since that's the only platform I bundled a shared object for, that should work

theli-ua:
Addressed feedback, tested to make sure the video texture in venues also is supported

Assets/Script/Gameplay/YargVideoPlayer.cs
theli-ua: This should only happen once, after video is loaded

Assets/Script/Gameplay/YargVideoPlayer.cs
theli-ua: its used by background manager

Assets/Script/Gameplay/YargVideoPlayer.cs
```diff
@@ -0,0 +1,338 @@
...
+
+#if VLC_SUPPORTED
+    [SerializeField] private LibVLCSharp.VLCMediaPlayer _vlcPlayer;
+    private bool _usingVLC = false;
+#endif
+
+    private string _url = "";
+    private VideoRenderMode _renderMode;
+    private double _time;
+    private float _playbackSpeed = 1.0f;
```
polson: Set but not used anywhere

Assets/Script/Gameplay/YargVideoPlayer.cs
```diff
@@ -0,0 +1,338 @@
...
+            }
+#endif
+            return _unityVideoPlayer.targetTexture;
+        }
+        set
+        {
+#if VLC_SUPPORTED
+            if (_usingVLC && _vlcPlayer != null && value != null)
+            {
+                Debug.Log($"[YargVideoPlayer] Ignoring external output texture in vlc mode");
```
polson: Should we be using YargLogger?

Assets/Script/Gameplay/TextureManager.cs
polson, line 39: Do we ever clear this list?

Assets/Plugins/vlc/Linux/x86_64/libVLCUnityPlugin.so.meta
```diff
@@ -0,0 +1,50 @@
...
+      settings:
+        Is16KbAligned: false
+    Any:
+      enabled: 1
+      settings:
+        Exclude Editor: 0
+        Exclude Linux64: 0
+        Exclude OSXUniversal: 1
+        Exclude Win: 0
```
polson: Should this be excluded also?

Assets/Scenes/Gameplay.unity
```diff
@@ -1789,7 +1868,7 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 2517693471320601084, guid: 039fee0ec42345d4f88c0625255415d4, type: 3}
       propertyPath: m_AnchoredPosition.x
-      value: -157.1503
```
polson, line 1792: Was this on purpose?

polson (changes requested):

Assets/Script/Gameplay/YargVideoPlayer.cs
theli-ua: Yes we should!

Assets/Script/Gameplay/TextureManager.cs
theli-ua: No, texture manager gets recreated, its part of the gameplay scene, it'll be fresh on each song

Assets/Plugins/vlc/Linux/x86_64/libVLCUnityPlugin.so.meta
theli-ua: I'm pretty sure it was disabled in the inspector, I will double check

Assets/Scenes/Gameplay.unity
theli-ua: Honestly, I have no idea what this one is

PR #1613, opened 2026-08-09
Black key glissandos

Frickitickitavi:
Conventionally, glissandos are used only for white keys. It's certainly not a good idea to chart a glissando that intermixes white keys and black keys, but there's really no reason why an all-black-key glissando shouldn't be charted if that's what the original keyboardist did.

This PR adds a new gem model for black keys in glissandos. Mechanically, glissandos are completely agnostic to specific keys - they accept all inputs and autohit all notes - so there's no need for a Core PR; they already work.

I took heed of the comment in `ThemeNote.cs` and inserted the new `BlackGlissando` type at `16` rather than pushing `Wildcard` and `DedicatedLaneKick` down to let it sit with the other PK types. However, I'm not really sure the warning needs to be heeded yet, since custom note themes aren't yet supported - if we can still get away with shuffling things around, I'd be happier putting it at `14` and bumping those other two down.

I also wasn't sure if renaming `Glissando` to `WhiteGlissando` would cause any serialization issues, so I left that alone, but would be happier to rename it if possible.

Test chart: 
[BlackKeyGlissandoTest.zip](https://github.com/user-attachments/files/30877362/BlackKeyGlissandoTest.zip)

The test chart contains glissando and nonglissando sequences of both white and black keys, both with and without OD, to test all aspects of the visuals and engine.

### DIFF
--- a/Assets/Prefabs/Gameplay/Visual/Themes/RectangularTheme.prefab
+++ b/Assets/Prefabs/Gameplay/Visual/Themes/RectangularTheme.prefab
@@ -155,7 +155,7 @@ Transform:
   m_GameObject: {fileID: 470677965324328071}
   serializedVersion: 2
   m_LocalRotation: {x: -0, y: -0, z: -0, w: 1}
-  m_LocalPosition: {x: 0, y: 0, z: 0}
+  m_LocalPosition: {x: -0.333, y: 0, z: 0}
   m_LocalScale: {x: 1, y: 1, z: 1}
   m_ConstrainProportionsScale: 0
   m_Children:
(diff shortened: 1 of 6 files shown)


Assets/Prefabs/Gameplay/Visual/Themes/RectangularTheme.prefab
```diff
@@ -155,7 +155,7 @@ Transform:
   m_GameObject: {fileID: 470677965324328071}
   serializedVersion: 2
   m_LocalRotation: {x: -0, y: -0, z: -0, w: 1}
-  m_LocalPosition: {x: 0, y: 0, z: 0}
+  m_LocalPosition: {x: -0.333, y: 0, z: 0}
```
wyrdough, line 158: Did you intend to move the black note? (Or am I misreading the diff)

Frickitickitavi: Yeah, so that the four keys gems in the prefab are equally spaced in the same range that previously held three (which is the same horizontal range that all the other note groups in the prefab are arranged between)

PR #1614, opened 2026-08-10
Harmony lipsync and singalongs

Supersonicboss1:
I should note that there is a lot of code that needs to be reworked to work properly with any number of performers on stage that is not 4, and this doesn't really try to fix that issue, though it is written in a way such that it shouldn't make it actively worse?

Anyway, most of what this does is just hook up lipsync to the correct `VRMCharacter` based on the `SingerPreference` array, as well as pass in the correct set of lipsync events for singalongs.

Requires YARC-Official/YARG.Core#441


Assets/Script/Venue/Characters/VRMCharacter.cs
```diff
@@ -103,9 +107,28 @@ public override void Initialize(CharacterManager characterManager = null)
...
+            if (_singalongLipsyncEvents.Count > 0 && _singalongEvents.Count > 0)
+            {
+                HasLipsyncAssigned = true;
+            }
+        }
+
         protected override void Update()
         {
-            if (_characterManager != null)
+            if (_characterManager)
```
wyrdough: Since `CharacterManager` derives (indirectly) from `MonoBehavior`, null checks need to be explicit so they trigger Unity to check the state of the C++ side since it is possible for the reference to not be valid there even though the C# reference is still valid. (Also, no pattern matching, Unity only checks when you use the traditional `==` and `!=` operators.

Assets/Script/Venue/Characters/VRMCharacter.cs
```diff
@@ -103,9 +107,28 @@ public override void Initialize(CharacterManager characterManager = null)
             base.Initialize(characterManager);
         }
 
+        public void InitializeLipsync(List<LipsyncEvent> lipsyncEvents)
```
wyrdough: I did not see any means of calling `InitializeLipsync` on non-bundled characters. (They are all-but-guaranteed to end up loading *after* `CharacterManager.Initialize()` runs)

Supersonicboss1: Fixed; we store the lipsync and singalong data on CharacterManager now, so we can copy it over to the new character even after init.

Assets/Script/Venue/Characters/VRMCharacter.cs
Supersonicboss1: If DotPeek is to be believed, they should behave pretty much identically, though I still changed it back for clarity.
Unity's Object class, for reference:
image

wyrdough: If Unity did really fix that stupid trap that is fantastic news

Supersonicboss1: Kind of? You still have the footgun with pattern matching; it's just that the object can be converted to a bool implicitly which checks C++ lifetime.

PR #1618, opened 2026-08-16
Testing framework

polson:
Adds Unit tests and a runner for the main YARG project.  Runs on every PR, no Unity editor required. 

Also added 3 tests.


polson:
Ok After working on this I have decided it's probably not worth it

Two things: 
1. We originally had a Unity EditMode test runner with GitHub actions.  This would use GameCI to download a unity docker image and run the EditMode tests.  This would let us test things that extend MonoBehavior etc.  This requires using our Unity licensing secrets to run, which presented security risks as it is easy for a malicious PR to leak the secrets if we allow it to run automatically.  We could gate the test run behind a button the maintainer has to push, but then contributors lose the benefits of being able to see tests fail on each push to a PR.

2. The simpler solution is to just run normal NUnit tests which is what this PR does now, but that means we have to extract out any testable logic from our MonoBehavior classes which means there's really not a lot we can test, and extracting it into a separate class makes the code annoying. 

So, I'm just going to close this and reconsider it at a later time.

PR #1620, opened 2026-08-18
feat: show built-in preset values read-only in the preset editor

XaiaX:
## Summary

Built-in (OOTB) presets now display their values read-only in the standard
preset editor UI instead of hiding them. The "You cannot modify a built-in
preset" notice still leads, with the preset's fields rendered after it, so
users can see what a default preset contains without copying it.

## Why

There was no way to inspect a built-in preset's values without copying it: selecting one showed only the notice and the preview controls. Answering "what does the default camera preset actually set?" required duplicating the preset.

## Design

- Fields build exactly as in edit mode; a read-only pass then disables interaction. Colors, swatches, and slider positions render exactly.
- Interaction is blocked via the `CanvasGroup` with alpha untouched, and control disabled-tints are overridden to opaque greys: sliders get a grey fill and a solid dark handle; an "on" toggle keeps a solid, darkened blue overlay.
- Value text (numeric inputs, hex fields) is gently dimmed to read as inert.
- Rows remain controller-navigable so long lists can be scrolled, but confirm is inert — this reuses the existing `IsEditable` guard in `BaseSettingNavigatable`.
- Color rows keep the stock navigatable (no color picker), and the "Copy from note" button is not created, since both could modify the preset.
- Preview controls (instrument dropdown, preview toggles) stay interactive on built-ins, as before.

## Changes

| File | Change |
|------|--------|
| `PresetsTab.cs` | Show the built-in notice *before* the fields; set the new flag. |
| `PresetSubTab.cs` | `HideFields` → `ReadOnlyFields` (fields render read-only instead of being skipped). |
| `PresetSubTab.Generic.cs` | Always build fields; `MakeFieldsReadOnly` pass; gate color-picker navigation and the copy-from-note button for read-only presets. |
| `BaseSettingVisual.cs` | `SetEditable` is virtual and takes a `dim` parameter (metadata rows keep the existing dimmed behavior). |
| `SliderSettingVisual.cs` | Read-only: grey fill, opaque dark handle (overrides the stock half-transparent disabled tint). |
| `ToggleSettingVisual.cs` | Read-only: opaque disabled tint, darkened blue "on" overlay. |
| `RuntimeNavigatable.cs` | New shared `DimmedSelectionGrey` (the selection color desaturated) so the color picker and read-only sliders use one definition. |
| `ColorPickerDialog.cs` | Slider-row selected grey now uses the shared constant instead of a local derivation. |
| `PresetDefaultText.prefab` | Fix "You can not" → "You cannot". |

## Notes

- Applies to every preset type (Camera, Color, Engine, Highway, Rock Meter) through the single `BuildSettingTab` path.
- `SetEditable`'s existing callers are unaffected; the new parameter defaults to the previous behavior.

## Testing done

- Unity batchmode compilation against the synced runtime: zero `error CS` on 2026-08-17.
- In-game verification on 2026-08-17 across Camera, Color, Engine, Highway, and Rock Meter presets: values visible and exact (colors unmodified, solid slider handles, single-state toggles), notice leading, no copy-from-note button, preview controls interactive, read-only rows inert under confirm and mouse.

### DIFF
--- a/Assets/Prefabs/Menu/Settings/Presets/PresetDefaultText.prefab
+++ b/Assets/Prefabs/Menu/Settings/Presets/PresetDefaultText.prefab
@@ -66,7 +66,7 @@ MonoBehaviour:
   m_OnCullStateChanged:
     m_PersistentCalls:
       m_Calls: []
-  m_text: '<font-weight=600>You can not modify a built-in preset.</font-weight>
+  m_text: '<font-weight=600>You cannot modify a built-in preset.</font-weight>
 
     In
     order to modify a preset, press on the copy button to make a preset based off
--- a/Assets/Script/Menu/Common/Dialogs/ColorPickerDialog.cs
+++ b/Assets/Script/Menu/Common/Dialogs/ColorPickerDialog.cs
@@ -228,7 +228,6 @@ namespace YARG.Menu.Dialogs
         /// </summary>
         private class SliderRowVisual
         {
-            private const float SELECTED_BAR_DARKEN = 0.7f;
             private const float EDITING_YELLOW_DIM = 0.8f;
 
             private static readonly Color SELECTED_HANDLE_GREY = new(0.82f, 0.82f, 0.82f, 1f);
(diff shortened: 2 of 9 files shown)


polson:
Include screenshots please, that will tell us more than the description about what this does

XaiaX:
Current view, no information provided to the user
Screenshot 2026-08-29 at 10 55 07

Modified views:

Default camera preset, showing dimmed / desaturated sliders and numeric values, still shows existing warning that you cannot edit default presets:
Screenshot 2026-08-29 at 13 26 24

Editable copy, sliders are colored, numeric values are brighter, "delete" button is now enabled:
Screenshot 2026-08-29 at 13 28 55

Dimmed / desaturated enabled checkboxes:
Screenshot 2026-08-29 at 13 27 25

Dimmed / desaturated color picker buttons, but color swatch is still original saturation/value. Also, the "copy from" button that is present in the editable version is removed, as it would not do anything:
Screenshot 2026-08-29 at 13 27 03

Editable version, saturated, "copy from" button appears.
Screenshot 2026-08-29 at 13 35 29

PR #1622, opened 2026-08-19
Preserve Playlist Entries when Songs are Updated in Place

Dansla116:
Currently, when a notes.mid file is modified and the library is rescanned, the song receives a new hash.  This is expected behavior and remains unchanged.  However, because playlists identify songs by hash, the song may disappear from any playlists containing it.

With this change, when a song's hash changes during a rescan but its directory remains the same, playlist entries using the old hash are updated to use the new hash.  Ambiguous matches, such as duplicate copies that produce different new hashes, are left unchanged.

### DIFF
--- a/Assets/Script/Song/SongContainer.cs
+++ b/Assets/Script/Song/SongContainer.cs
@@ -14,6 +14,7 @@ using YARG.Helpers.Extensions;
 using YARG.Localization;
 using YARG.Menu.MusicLibrary;
 using YARG.Player;
+using YARG.Playlists;
 using YARG.Scores;
 using YARG.Settings;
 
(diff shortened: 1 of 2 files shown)


Assets/Script/Song/SongContainer.cs
```diff
@@ -154,9 +155,11 @@ public static async UniTask RunRefresh(bool quick, LoadingContext? context = nul
             }
 
             var stopwatch = System.Diagnostics.Stopwatch.StartNew();
+            var previousSongCache = _songCache;
+            SongCache refreshedSongCache = null;
             var task = UniTask.RunOnThreadPool(() =>
             {
-                _songCache =  CacheHandler.RunScan(quick,
+                refreshedSongCache = CacheHandler.RunScan(quick,
```
wyrdough, line 162: I may just be blind, but it looks like the updated cache is never assigned to `_songCache`?

Dansla116: Right after this on line 180.  The assignment is delayed so both the previous and refreshed caches are available for determining hash replacements.

wyrdough: That's what I get for attempting a quick once over on my phone

PR #1624, opened 2026-08-20
Add Options for Song Duration Groups in Sort and Filter

Dansla116:
## Description
The Music Library's "Sort by Duration" and the Filters menu's "Song Lengths" currently use different groups.  This change synchronizes them and adds an advanced setting for selecting between:
- Range Labels: under 2, 2–5, 5–10, 10–15, 15–20, and 20+ minutes
- Legacy Labels: under 3, 3–5, 5–7, and 7+ minutes
Range Labels are the default.

This fixes: https://discord.com/channels/1086048856678084609/1528258447701971015

### Default Range Labels
Settings:
image

Music Library Menu:
image

Filters Menu:
image

### Optional Legacy Labels
Settings:
image

Music Library Menu:
image

Filters Menu:
image

### DIFF
--- a/Assets/Script/Song/SongContainer.cs
+++ b/Assets/Script/Song/SongContainer.cs
@@ -807,10 +807,7 @@ namespace YARG.Song
                     return Cast(_sortedSongs.SongLengths);
                 }
 
-                var groups = new List<SongEntry>[4]
-                {
-                    new(), new(), new(), new(),
-                };
+                var groups = new SortedDictionary<int, List<SongEntry>>();
 
                 // The range categories and their contents are already ordered by duration.
                 foreach (var range in _sortedSongs.SongLengths.Values)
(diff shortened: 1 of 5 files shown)


Assets/Script/Song/SongContainer.cs
```diff
@@ -800,6 +800,56 @@ CharacterGroup.Empty or
...
+            {
+                if (SettingsManager.Settings.SongLengthLabels.Value == SongLengthLabelMode.RangeLabels)
+                {
+                    return Cast(_sortedSongs.SongLengths);
+                }
+
+                var groups = new List<SongEntry>[4]
+                {
+                    new(), new(), new(), new(),
+                };
```
wyrdough: If the user doesn't actually have any songs in a given category, shouldn't we avoid displaying it?

Dansla116: Empty buckets weren't displayed because they were skipped when building the sections.  However, you're right that it's cleaner and more consistent with the existing range-label implementation to create buckets lazily, so I've updated it accordingly.

PR #1629, opened 2026-08-20
Add in new crowd settings

polson:
Replaces drop down with three crowd settings

Crowd idle means the mid-song crowd idling

With Crowd cheer on we hear the idle, then cheer, then silence

I do not personally like the crowd idle noise for the entire song, and I do not like claps, so I turn on crowd cheer only

Also includes a formal settings migration method

image

### DIFF
--- a/Assets/Script/Gameplay/GameManager.Audio.cs
+++ b/Assets/Script/Gameplay/GameManager.Audio.cs
@@ -5,7 +5,6 @@ using DG.Tweening.Core;
 using DG.Tweening.Plugins.Options;
 using UnityEngine;
 using YARG.Core.Audio;
-using YARG.Playback;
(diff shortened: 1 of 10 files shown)


wyrdough:
There seems to be no equivalent to the previous star power claps only setting? With that engaged, the crowd only claps during SP (and other crowd sounds are disabled), not also any time the player is performing well and the chart hasn't called for noclap.

polson:
> There seems to be no equivalent to the previous star power claps only setting? With that engaged, the crowd only claps during SP (and other crowd sounds are disabled), not also any time the player is performing well and the chart hasn't called for noclap.

Very good point you have made... Maybe we have to rethink this

polson:
Made it 4 toggles:

image

PR #1634, opened 2026-08-21
Add gameStateTexture to texture manager

theli-ua:
This is stacked on top of #1632 , leaving as draft for now.
Adds a game state texture with a sample shader displaying song progress included.

Allows pixel/vertex shader code based on some game state variables.


theli-ua:
Example usage in a venue:

image
image

PR #1635, opened 2026-08-22
Move delayed button arming into ShowCompactConfirmation, fix color flash

XaiaX:
## Summary

The compact confirmation dialog's delayed-delete pattern (used by preset
deletion) is now a built-in option (`armDelaySeconds`) on
`PresetSubTab.ShowCompactConfirmation`, instead of caller-side code. This also
fixes two defects in the existing pattern:

1. **One-frame color flash.** The confirm button rendered its enabled color for
   a frame before disabling, because the `Button` color-tint transition lerps
   over its fade duration. The arm path now zeroes the fade while switching, so
   the disabled color applies on the first rendered frame.
2. **Navigation bypass.** The dialog's Green button navigation entry invoked the
   confirm action directly, so a controller/keyboard confirm could trigger a
   destructive action during the "disabled" window. The entry is now gated on
   the armed state (mouse clicks were already guarded by `interactable`).

## Changes

| File | Change |
|------|--------|
| `PresetSubTab.cs` | `ShowCompactConfirmation` gains `armDelaySeconds` (default 0 — existing callers unaffected); new `ArmButtonAfterDelay` disables the button fade-free, arms it after the delay, and flips an `armed` flag that gates the Green scheme entry. |
| `PresetActions.cs` | `DeletePreset` collapses to a single `ShowCompactConfirmation(..., armDelaySeconds: 2f)` call. |

## Testing done

- Unity 6000.3.5f2 batchmode compile: zero `error CS`.
- Core unit tests unchanged (no Core code touched).
- Manual, in-editor (the Unity project has no test assembly; all dialog logic
  is Unity-side):
  - Preset delete: no enabled-color flash; Delete grey for ~2s then arms red;
    confirming deletes the preset and closes the dialog.
  - Keyboard `1` (Green) during the 2s window: no-op. `2` (Red): instant
    cancel.

Companion PR: profile load recovery (`bug/profile-load-recovery`) builds on
this and uses `armDelaySeconds` for its delete confirmations.

### DIFF
--- a/Assets/Script/Menu/Settings/Presets/PresetActions.cs
+++ b/Assets/Script/Menu/Settings/Presets/PresetActions.cs
@@ -68,46 +68,24 @@ namespace YARG.Menu.Settings
             if (preset.DefaultPreset) return;
 
             // Deleting is irreversible, so confirm first (same compact dialog
-            // as "Copy from note"). The delete button starts disabled/grey for
-            // a moment so it can't be hit by accidental mashing.
-            bool armed = false;
-
-            void Delete()
-            {
-                if (!armed) return;
-
-                DialogManager.Instance.ClearDialog();
-
-                _tab.SelectedContent.DeletePreset(preset);
-                _tab.ResetSelectedPreset();
-
-                SettingsMenu.Instance.Refresh();
-            }
-
-            // The cancel button keeps its brighter "safe" color here (delete is
-            // the destructive action, so it stays the red default).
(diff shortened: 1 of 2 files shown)


XaiaX:
This PR is a subset of https://github.com/YARC-Official/YARG/pull/1636 so merging this one first will reduce the diff size there, or that PR can be merged as-is and supersede this PR entirely.

polson (approved):

PR #1641, opened 2026-08-27
Performance Improvements when Sorting By Score Context

Dansla116:
## Description

Score selection previously used correlated SQLite subqueries that repeatedly searched the score history for each candidate record.
This change:
- Restricts score processing to hashes in the current song library.
- Uses ROW_NUMBER() window queries partitioned by song checksum to select the preferred score in one pass.
- Applies the window-query approach to best-stars, high-score-cache, and percentage-cache queries.
- Adds composite indexes for the score filtering and join patterns.
- Stores current library hashes in a temporary SQLite table, avoiding parameter-count limits.
- Reuses that table until SongContainer rebuilds the library.
- Uses an explicit library revision counter to invalidate the temporary table after a scan or container refresh.

This is strictly a query and caching optimization.  Existing score behavior and relevant implementation constraints remain unchanged:
- Preserves **all** existing score-history modes and their ordering rules.
- Does **not** delete historical scores.
- Does **not** introduce LINQ into SongContainer or song-cache paths.

Before/After Comparison with 6,380 songs and 5,916 total player-score records.
| Process | Before | After |
| :-- | --: | --: |
| High-Score Cache Query | **5,983ms** | 36ms |
| Populate 6,380 active hashes | 16ms | **18ms** |
| Ranked best-stars SQL query | **51ms** | 42ms |
| Map 3,268 returned scores | 1ms | 1ms |
| Map stars onto 6,380 songs | 1ms | 1ms |
| Group and sort songs | 6ms | 6ms |
| Create star categories | 1ms | 1ms |
| Total Stars-sort construction | **78ms** | 71ms |
| Build 6,380 song views and calculate star totals | **11,884ms** | 122ms |
| Total | **18,021ms** | 298ms |

~98.3% reduction or ~60x performance improvement


wyrdough:
Do you have any sense of how the update to the Linux version of the sqlite lib will impact compatibility with older distro versions?

Dansla116:
I checked the committed library's ELF symbol versions, and its newest glibc requirement is GLIBC_**2.14**.  We document glibc **2.31** as our compatibility floor in `Native/YargAudio/README.md`, so this should not reduce compatibility with supported Linux environments.

The update is required for SQLite window functions such as `ROW_NUMBER()` and `OVER()`.  The previously bundled Linux SQLite 3.13 rejected those queries on my Steam Deck, while Windows was already using a newer SQLite version that supports them.

wyrdough:
Thanks for checking.

PR #1645, opened 2026-08-29
Copy vlc-unity files instead of symlinks

theli-ua:
Symlinks too fragile and confusing


polson:
What do you think about just building your own small wrapper that calls into `LibVLCSharp.dll` and `libVLCUnityPlugin.so` directly, then you could just remove the entire `VLCUnity` directory and submodules?

theli-ua:
> What do you think about just building your own small wrapper that calls into `LibVLCSharp.dll` and `libVLCUnityPlugin.so` directly, then you could just remove the entire `VLCUnity` directory and submodules?

The submodules are not necessary at all, nothing is used from them. They are only there to indicate which commit hash we built all the vls stuff from.

Like, you can remove them without breaking anything.

As far as having our own wrapper - I see no reason to do that instead of simply copying vlcunity from upstream when we want to update

polson:
I guess I'm still confused about the submodules, do we need them still or no?  If we don't then I'd say remove them from the PR

theli-ua:
Removed submodules. Added build script to build both native plugin and libvlcsharp

polson:
Ok so I see three parts to this:

`libVLCUnityPlugin.so`
`LibVLCSharp.dll`
`VLCUnity` directory

It is not clear to me how all of this interacts.  Do you think we should add some documentation to ./docs?

polson:
Wait I think I get it, `VLCUnity` directory talks to `LibVLCSharp.dll` talks to `libVLCUnityPlugin.so` ?

polson:
I see these errors on windows:

```
DllNotFoundException: VLCUnityPlugin assembly:<unknown assembly> type:<unknown type> member:(null)
LibVLCSharp.OnLoad.OnBeforeSceneLoadRuntimeMethod () (at Assets/VLCUnity/Internal/OnLoad.cs:56)

DllNotFoundException: VLCUnityPlugin assembly:<unknown assembly> type:<unknown type> member:(null)
TrialNotice.LogTrialNotice () (at Assets/VLCUnity/Internal/TrialNotice.cs:66)

DllNotFoundException: VLCUnityPlugin assembly:<unknown assembly> type:<unknown type> member:(null)
TrialNotice.OnPlayModeChanged (UnityEditor.PlayModeStateChange state) (at Assets/VLCUnity/Internal/TrialNotice.cs:42)
UnityEditor.EditorApplication.Internal_PlayModeStateChanged (UnityEditor.PlayModeStateChange state) (at <b95741d8534d46b7b21ea8475977e676>:0)


```

theli-ua:
I don't have a way to test on Windows unfortunately. 
I think ideally native plugin would just be built for all 3 platforms (since it does support that), but I don't have a way to do that myself either.

And it looks like VLCUnity does unconditionally try to call into native plugin.

Worked around that by patching VLCUnity.

And to answer your earlier question:

* libVLCSharp - is just a C# wrapper for libvlc. We can't use the nuget though as it needs specific flag at build time
* VLCUnityPlugin is a native plugin that is mostly responsible to transfer data between vlc and unity textures, for output
* VLCUnity is a unity specific glue that ties everything together an provides Unity components

We're not expected to maintain either of those really, they are only pre-built/copied because there is no other way to consume them. VLC-unity does not expose itself as a unity package (by not having package.json and pre-build binaries)

polson:
I'll give this a test again, if the errors go away I'll merge

polson:
I get this error on quickplay when I start a song

```
DllNotFoundException: VLCUnityPlugin assembly:<unknown assembly> type:<unknown type> member:(null)
LibVLCSharp.Core.InitializeUnity (System.String libvlcDirectoryPath) (at <e3af4e1643c444e193dc9d2392e207cc>:0)
LibVLCSharp.Core.Initialize (System.String libvlcDirectoryPath) (at <e3af4e1643c444e193dc9d2392e207cc>:0)
LibVLCSharp.VLCMediaPlayer.CreateLibVLC () (at Assets/VLCUnity/Internal/VLCMediaPlayer.cs:388)
LibVLCSharp.VLCMediaPlayer.Awake () (at Assets/VLCUnity/Internal/VLCMediaPlayer.cs:91)
```

theli-ua:
I've made a test build off of latest commit ( [847d42e](https://github.com/YARC-Official/YARG/commit/847d42e36c461f6708ed03f11ef34026c10a37e5) ) 
https://github.com/theli-ua/YARG-BleedingEdge/releases/tag/vlc-2026-08-31-01

And verified (with WINE) that windows version works perfectly fine

theli-ua:
I've just added native plugins for osx and windows.
I have verified that the new [build](https://github.com/theli-ua/YARG-BleedingEdge/releases/tag/vlc-2026-09-01-00)  
works on windows and osx (falls back to unity player since I didn't add vlc libraries)

theli-ua:
I've now also verified that VLC path works on windows perfectly as well (just downloaded nightly vlc4 build and copied libs)

theli-ua:
Confirmed it works both with and without vlc libs on osx as well

PR #1646, opened 2026-08-29
Offer more options for end of song calibration

asimard1:
## Motivation

This is a PR that seeks to make my recently merged PR (#1535) better. The goal of this PR was to add an offset to songs which are not well calibrated during the chart creation to (ideally) make them perfect during gameplay. This was implemented as a button option on the histogram screen that permitted users to use their actual inputs for a song as an offset calibrator for that specific song. This makes it better by adding more options (like median/average, strums/no strums, kicks/no kicks) and also makes the end screen more detailed.

I have found during the last few weeks of working on #1535 that my strumming was a lot more precise than my HOPOs/taps. I think that most people will find this the case. My solution was this: Instead of calibrating blindly on every single input during a song (for both auto and end of song), would it be better to only use strums, and ignore HOPOs/taps. For instruments that don't have HOPOs and taps, the default is to use all notes. The same is true for guitar instruments playing songs that simply do not have strum notes.

Obviously, this is not something that should be on by default, so I added it to the settings.

Let me know if anything can be improved on :)

## The other changes

I also decided to add the option to calibrate using median instead of average at the end of a song. I saw that the auto calibration was already using median, and I added the option to set that too to average, however that can be removed if it doesn't please the dev team. (Obviously, it could very well lead to worse auto calibration since it's only based on 20 notes, but I thought it couldn't hurt to have the option somewhere.)

The last thing that this changes is the way that the "average average offset" (not a typo) is computed (this was added in #1535). Before, when playing with a band, the value that was added at the end of a song was simply the average of all the individual offset averages. However, I thought it weird to have a person playing on easy (who had like 100 notes) have the same impact on the average as someone playing on expert. (Also, the skill levels are different so the calibration accuracy is too.) I decided to have the average be weighted by the amount of notes that each person hit in the song. This gives more weight to the people who had good scores and had a lot of notes, and this also gives a better calibration overall.

## Visual representation

In order to see the difference between the strum and the hopos/taps offsets, I modified the histogram in the detailed results to show both separately. I also added median to the list, and sometimes it serves as a nice reference to see how skewed the input offset is. (Btw, I'm a math teacher so this is why I'm doing so much math in my PRs.) These changes look like this:

image

The white below is the offset graph for only the strum notes, and the gray above is everything. So the gray is what would have appeared without this PR, and the white is the new part that is now usable with these changes.

## A few notes

Looking at the graph above, it is clear that the average offset was normally brought down by the infinite frontend. The spike in inputs at -70ms is very telling. This has almost no chance of happening with strums, and so this PR mitigates this problem of aberrant values. Another thing that can be noted is that the median is closer to the average for strums than for all the notes. This in general has been true in my testing as well.

The thing to keep in mind is: calibrating using strums gives an average that is better distributed (mathematically speaking), and has less tendency to be affected by aberrant values that are caused by the infinite frontend. (And in case you were wondering, there is still a difference without infinite frontend as well, and the calibration does feel better when calibrating with only strums.)

## The only issue

If a user has chosen to calibrate using only strum inputs, and there are less than 20 strums hit before the end of the song, the auto calibration will just not happen. The logic is: If there are 20 notes without strum, calibrate using that, but as soon as there is one strum, wait until there are 20 strums. That is obviously solvable by opening the quick settings and setting "use strums only" to off, but it could also be improved. Let me know how you feel about this.


XaiaX:
I love the doubled graph, I have noticed a similar issue with my kicks vs. pads. If I calibrate with no kicks, and play a song with no kicks, I can nail a 0ms offset average. If I then add kicks back in, I'm consistently around -17ms.  So, if you could add a similar kicks vs. pads (not sure if distinguishing cymbals would matter but maybe even that?) I would love it.

asimard1:
I will do that, yes !

XaiaX:
Bonus for using median instead of average, which is the devil's measurement of centrality.

asimard1:
Well median does have a bit more variance in some cases, but that assumes that all strum offsets are distributed like normal variables, which should not be the case in general (we are not robots playing perfectly). So it should be a lot better, yes.

asimard1:
@XaiaX I have started adding the architecture to let users calibrate:
- For guitars: using only strums, only hopos/taps (if one so wishes) or every note.
- For drums: using only kicks, only pads/cymbals (I haven't split them) or every note.

The only thing that I have removed is the quick settings options to change these settings, because they are now dropdowns and dropdowns don't seem to exist for quick settings... I'll try to see if it's necessary to find a way, but otherwise it'll be controlled in the main menu, it's not so bad :)

asimard1:
I suck at drums, so I can't test in details, but that seems to have done the trick. Next step: The line that will be used for calibration should be in white rather than in gray, that way it's easy to tell which value will be used according to the settings... I'll add that soon and it'll be ready.

image

asimard1:
I reordered the option menus and reverted one to a toggle, just because it felt a bit overwhelming and didn't have a lot of positive impact. At the moment, it feels cleaner and much clearer.

Also, I added a different print color for the line that is of interest -- that is the line that will be used in the computation if the person chooses the calibrate with the button at the end of a song. It felt confusing to my friends when making parties which number was the one to look at, so hopefully this should make things more clear.

Let me know if there's anything to do !

XaiaX:
Maybe make the target calibration line be in the default UI blue?

neilweber:
> @XaiaX I have started adding the architecture to let users calibrate:
> 
>     * For guitars: using only strums, only hopos/taps (if one so wishes) or every note.
> 
>     * For drums: using only kicks, only pads/cymbals (I haven't split them) or every note.
> 
> 
> The only thing that I have removed is the quick settings options to change these settings, because they are now dropdowns and dropdowns don't seem to exist for quick settings... I'll try to see if it's necessary to find a way, but otherwise it'll be controlled in the main menu, it's not so bad :)

I'm an intermediate drummer and there is no way I'd want my kicks to be included in any calibration.

It seems to me the root problem is requiring the player to select a good chart to play.  It'd be simpler to have YARG approved easy chart that has no HOPOs/taps or kicks that everyone uses for calibration.  It shouldn't be fancy -- just thirty seconds long.

asimard1:
@neilweber The thing is that the goal of this is not only to calibrate correctly on one song. The goal of these changes is to detect more easily any chart that might be offset (from a track-editing point of view, that is a track that wasn't well timed in moonscraper for example) and correct it. So yes, calibration should be done correctly on an easy chart, but after that the feature extends further (and works very well I must say, I'm playing on my custom build and could not imagine playing without this feature).

The goal with the "only kick"/"no kicks"/"everything" was just to give more options, I haven't got an actual drum set for the moment (thus the bad graph that I posted which I played on keyboard), but I remember my kicks being more precise than my hands, so I wanted to leave that choice up to each player... Does that make sense?

asimard1:
> Maybe make the target calibration line be in the default UI blue?

I found the white to be fitting, but yes that can be changed. I'll try to upload an image later that shows it well, I really think the white works with the rest of the UI elements (especially the histogram which is in white also).

neilweber:
> @neilweber The thing is that the goal of this is not only to calibrate correctly on one song. The goal of these changes is to detect more easily any chart that might be offset (from a track-editing point of view, that is a track that wasn't well timed in moonscraper for example) and correct it.

Ah, that wasn't mentioned in the Motivation.  Anyway, when I get a chance I'll try it out with me ekit.

asimard1:
You're right, sorry about that :/ I'll add it to the first message.

Edit : done.

polson:
> > @XaiaX I have started adding the architecture to let users calibrate:
> > ```
> > * For guitars: using only strums, only hopos/taps (if one so wishes) or every note.
> > 
> > * For drums: using only kicks, only pads/cymbals (I haven't split them) or every note.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The only thing that I have removed is the quick settings options to change these settings, because they are now dropdowns and dropdowns don't seem to exist for quick settings... I'll try to see if it's necessary to find a way, but otherwise it'll be controlled in the main menu, it's not so bad :)
> 
> I'm an intermediate drummer and there is no way I'd want my kicks to be included in any calibration.
> 
> It seems to me the root problem is requiring the player to select a good chart to play. It'd be simpler to have YARG approved easy chart that has no HOPOs/taps or kicks that everyone uses for calibration. It shouldn't be fancy -- just thirty seconds long.

I would suggest just turning off kicks in the modifiers then

polson:
> @neilweber The thing is that the goal of this is not only to calibrate correctly on one song. The goal of these changes is to detect more easily any chart that might be offset (from a track-editing point of view, that is a track that wasn't well timed in moonscraper for example) and correct it. So yes, calibration should be done correctly on an easy chart, but after that the feature extends further (and works very well I must say, I'm playing on my custom build and could not imagine playing without this feature).
> 
> The goal with the "only kick"/"no kicks"/"everything" was just to give more options, I haven't got an actual drum set for the moment (thus the bad graph that I posted which I played on keyboard), but I remember my kicks being more precise than my hands, so I wanted to leave that choice up to each player... Does that make sense?

I'm not really seeing any point to this, if you want to calibrate with just kicks then just hit kicks and nothing else.  If you dont want to calibrate with kicks you can just not hit the kicks, or turn them off with modifiers.  What do you think?

Likewise for guitar / keys, just use the modifiers to auto calibrate how you want

neilweber:
> > > @XaiaX I have started adding the architecture to let users calibrate:
> > > ```
> > > * For guitars: using only strums, only hopos/taps (if one so wishes) or every note.
> > > 
> > > * For drums: using only kicks, only pads/cymbals (I haven't split them) or every note.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The only thing that I have removed is the quick settings options to change these settings, because they are now dropdowns and dropdowns don't seem to exist for quick settings... I'll try to see if it's necessary to find a way, but otherwise it'll be controlled in the main menu, it's not so bad :)
> > 
> > 
> > I'm an intermediate drummer and there is no way I'd want my kicks to be included in any calibration.
> > It seems to me the root problem is requiring the player to select a good chart to play. It'd be simpler to have YARG approved easy chart that has no HOPOs/taps or kicks that everyone uses for calibration. It shouldn't be fancy -- just thirty seconds long.
> 
> I would suggest just turning off kicks in the modifiers then

And how many people realize that?  Calibration is a big issue for many people and asking them to pick a good chart and then disable kicks or HOPOs seems to me to be the most complex and error prone approach.  IHMO, calibration should be streamlined and beginner-proofed as possible.

asimard1:
@polson The point of this PR and of the previous one was that calibration could become something that was done during normal gameplay. Play a song normally, then get direct feedback at the end and the option to adjust the offset so it's better next time. Turning off kicks or playing without hopos/taps would defeat that gameplay purpose, especially when playing in groups (for which my calibration method works really well since I take a weighted average).

It's way more fun to play like normal, without worrying about kicks/hopos messing up auto and end-of-song calibration if they're automatically excluded from the average/median.

The same goes for strums on guitar. I used to calibrate by skipping hopos/taps like you suggested, but then I'd have to restart the song anyway from a bad score. Now I can play normally, and while the calibration might not be perfect, it'll be better next time. This argument is even stronger when grinding out an FC on a song, since every time you play it it gets more precise.

This PR solves that problem, no need to change how you play just to calibrate. I really hope you consider it since it's made my experience with the game a lot better and I'm sure other people will appreciate it :)

polson:
> And how many people realize that? Calibration is a big issue for many people and asking them to pick a good chart and then disable kicks or HOPOs seems to me to be the most complex and error prone approach. IHMO, calibration should be streamlined and beginner-proofed as possible.

I agree, the goal is to eventually ship a calibration track, which would be just hitting notes along to a metronome in beginner mode (so you can hit anything).  This would replace the settings calibration

asimard1:
@polson I agree that your last proposition would solve calibration and would be an amazing change. I think this PR can complement this by also giving better in-song calibration.

In particular, I don't think the use of modifiers fully solves the problem I have in mind. What I'm seeing here is a gameplay friendly approach to calibrating that is also frequent feedback (every end of song gives you information about how precise you were).

Hitting only kicks or strum means missing every other note in the song which tanks your score, not something most people will do just to test if a song is offset correctly. Turning kicks off with modifiers has the same issue: it changes the song itself, not just how it's measured. What I'm going for is different: you play the song exactly as intended, and the end of song math uses only whatever inputs you chose to consider. This is also easier to set up than modifiers, since it's a one time setting and can be ignored completely (since the end of song calibration and auto calibration are only user input activated). It doesn't need to be toggled on or off on every song.

This is also a lot more useful when playing with other people. When you notice that a song is slightly off, it's not great to interrupt normal gameplay to set up modifiers just to figure out what's wrong. With my addition, you can simply play the song normally, get to the end screen, and use the results to solve the issue directly. It also addresses a secondary problem: my friends don't necessarily play the same way I do, so my calibration doesn't always work as well for them, and vice versa. The more detailed end-song information I added lets us adjust their profile instrument delay with much more precision. Even if I have friends who don't play as consistently as I do and whose results are therefore less reliable, the weighted average gives those results less influence, and in practice it works very well.

XaiaX asked for the kicks/pads split directly, and my own histogram (posted above) shows a clear -70ms cluster that's very likely the infinite frontend and not real timing error, something I can't really tell apart without this. So at least for some of us this solves a real problem, not just adds an option for the sake of it.

Does this address your concern, or does the existence of the option itself bother you? Feel free to let me know if there's something specific about the implementation that still doesn't sit right :)

(Sorry for the long comment D:)

polson:
> @polson I agree that your last proposition would solve calibration and would be an amazing change. I think this PR can complement this by also giving better in-song calibration.
> 
> In particular, I don't think the use of modifiers fully solves the problem I have in mind. What I'm seeing here is a gameplay friendly approach to calibrating that is also frequent feedback (every end of song gives you information about how precise you were).
> 
> Hitting only kicks or strum means missing every other note in the song which tanks your score, not something most people will do just to test if a song is offset correctly. Turning kicks off with modifiers has the same issue: it changes the song itself, not just how it's measured. What I'm going for is different: you play the song exactly as intended, and the end of song math uses only whatever inputs you chose to consider. This is also easier to set up than modifiers, since it's a one time setting and can be ignored completely (since the end of song calibration and auto calibration are only user input activated). It doesn't need to be toggled on or off on every song.
> 
> This is also a lot more useful when playing with other people. When you notice that a song is slightly off, it's not great to interrupt normal gameplay to set up modifiers just to figure out what's wrong. With my addition, you can simply play the song normally, get to the end screen, and use the results to solve the issue directly. It also addresses a secondary problem: my friends don't necessarily play the same way I do, so my calibration doesn't always work as well for them, and vice versa. The more detailed end-song information I added lets us adjust their profile instrument delay with much more precision. Even if I have friends who don't play as consistently as I do and whose results are therefore less reliable, the weighted average gives those results less influence, and in practice it works very well.
> 
> XaiaX asked for the kicks/pads split directly, and my own histogram (posted above) shows a clear -70ms cluster that's very likely the infinite frontend and not real timing error, something I can't really tell apart without this. So at least for some of us this solves a real problem, not just adds an option for the sake of it.
> 
> Does this address your concern, or does the existence of the option itself bother you? Feel free to let me know if there's something specific about the implementation that still doesn't sit right :)
> 
> (Sorry for the long comment D:)

If the option is just on the score screen then that's fine

asimard1:
> If the option is just on the score screen then that's fine.

There are some new settings in the main menu to control which notes are used for auto calibration (disabled by default), but apart from those, all of the visual changes happen on the score screen: the new histogram, the additional rows of information, and the optional calibration button that already exists in the game. Does that qualify for what you have in mind?

polson:
Auto calibration is the place where I don't like having tons of options.

Like I said before, if you want to calibrate without kicks, dont hit the kicks or turn them off in modifiers.  

Or just use beginner mode and you can calibrate hitting whatever you want.

I don't understand the concern about scores since auto calibration disables scores already.

What do you think?

asimard1:
I guess the main use case for the changes to the auto calibrator for me personally is:

If I'm a few minutes into a long song and realize that the song's timing is off and that's why I've been missing a lot of notes, with my change I could just turn on auto calibration and finish the song normally instead of restarting from the beginning just to turn on the modifier. I obviously lose the score, which isn't great, but if I'm just playing for fun, that's not really a concern. And honestly, if I have to quit out to enable the modifier, I might not play that same song anyway lol

I could also imagine a song that is well calibrated when starting but has some offset issues in the middle (I have seen that happen with customs found online), and then I would have to quit out, activate the modifier, go into practice mode to get the offset of that section... It just sounds like a very hands-on approach, while what I'm suggesting is more of a brainless way to do the same thing (which might be well received by more casual players).

I completely understand if that's still too niche to justify the change, though 🥲 What do you think?

Thank you for considering the PR, I really appreciate the feedback :)

polson:
Thanks for explaining your use case.  I think instead of putting this stuff as part of the calibration feature, given that we have a new per-track menu and will soon have the option to change difficulty on the fly, we should also be able to change modifiers on the fly as well at some point.  

I think it would be better then to do **Track Pause -> Change Modifiers -> Pause -> Enable auto calibration** for this use case

Calibration is already complicated for users with the three toggles.  Most users don't know even know what a song offset is, some don't even understand the difference between video and audio calibration.  So I am hesitant to make calibration any more complicated.

asimard1:
I completely agree with your message. What you propose would fully solve the in-song calibration issue. Long-term, I wonder if there's room for an in-game tutorial explaining non-intuitive mechanics (anchoring, the three calibration numbers, strums vs. HOPOs, BREs, trill lanes (even I don't fully know what those do lol)). Not something for now of course, just a thought.

I'll remove the auto calibration features (and locale references) so the PR can move forward. Thanks for working through this with me and let me know if anything else needs changing.

asimard1:
@polson I changed the auto calibration like you wanted, do you know when this PR might get merged? This is honestly the only thing keeping me from just playing on the nightly build and I just want to uninstall unity lol

polson:
> [!NOTE]
> *This review comment and reproduction report was generated with AI assistance.*

Found a few issues with the end-of-song calibration calculations and settings handling:

### 1. Filtered calibration ignores "No Strums" / "No Kicks" (`ExcludeSelected`)

When setting end-of-song calibration to exclude specific notes (such as choosing "No Strums" or "No Kicks"), `GameManager.GetWeightedFilterCategorySamples` hardcodes `if (filterCategory[i])`, which collects only the selected category (strums for guitar, kicks for drums). Saving the offset on the results screen saves the opposite of the selected setting and contradicts the highlighted row on the score card.

**Reproduction Steps:**
1. Open **Settings > Calibration > End of Song Calibration**.
2. Set **Strum Filter for Calibration** to **No Strums**.
3. Play a guitar chart containing both strums and HOPOs/taps.
4. On the score screen, observe that the score card displays and highlights **HOPO/TAP AVERAGE** in white.
5. Press the button to save the end-of-song offset.
6. Inspect `song_offsets.json`: the saved offset value matches **STRUM AVERAGE** rather than the highlighted **HOPO/TAP AVERAGE**.

- **Expected:** Choosing "No Strums" or "No Kicks" pools the excluded side (HOPOs/taps or pads) and saves that value.
- **Actual:** The calculation always pools strums or kicks, saving the notes intended to be ignored.

**Reproduction Video:**

https://github.com/user-attachments/assets/8b8cb111-27e2-4e35-8e15-5227c7cf9acd

*(Shows "No Strums" set, score screen highlighting HOPO/Tap average at -35ms, and saving offset writing +45ms strum average to song_offsets.json)*

---

### 2. Multi-instrument setting bleed in end-of-song offset saving

`ScoreScreenMenu.ToggleOffsetToJson` switches to `MeanAverageOffsetFilterCategoryOnly` if *either* the guitar strum filter or the drum kick filter is set to anything other than `Everything`. If a player sets guitar to "Only Strums" and later plays solo drums with kick filtering set to "Everything", saving the offset uses the kick-only average instead of the all-pads average displayed and highlighted on the drum scorecard. Additionally, `GetWeightedFilterCategorySamples` unconditionally filters any player with a category list to `filterCategory[i] == true`, ignoring per-instrument settings in band play.

**Reproduction Steps:**
1. Open **Settings > Calibration > End of Song Calibration**.
2. Set **Strum Filter for Calibration** to **Only Strums**, and **Kick Filter for Calibration** to **Everything**.
3. Play a song on solo drums.
4. On the score screen, observe that the drum score card highlights **AVERAGE OFFSET** (all pads) as the calibration row.
5. Press the button to save the end-of-song offset.
6. Inspect `song_offsets.json`: the saved offset uses the kick-only average instead of the all-pad average.

- **Expected:** Solo drum calibration respects the drum kick filter setting and saves the all-pad average when set to "Everything".
- **Actual:** The guitar strum setting triggers filtered mode and silently saves a kick-only offset.

**Reproduction Video:**

https://github.com/user-attachments/assets/f551352a-7017-444f-a925-87cf9e96f945

*(Shows guitar "Only Strums" + drum "Everything", drum score card highlighting all-pads average at -1ms, and saving offset writing kick-only -40ms)*

---

### 3. Chronological hit order desynchronizes drum chord kick/pad categorization

In drum chords, individual pads and pedals are timed and recorded chronologically into `OffsetSamples` as each input event arrives. However, post-song note categorization iterates chart notes where kick is always pad 0. When a drummer hits a cymbal or snare milliseconds before the kick on the same beat, the snare offset is recorded first in `OffsetSamples`, but the kick boolean is recorded first in `GetOffsetSampleFilterCategory()`. This swaps the kick and pad classification for those hits in the histogram and kick-filtered offset calculations.

- **Expected:** Timing samples remain linked to the exact drum pad or kick pedal that generated them.
- **Actual:** Offsets and pad categories swap positions whenever chord pads arrive out of chart order.

PR #1647, opened 2026-08-29
Add build preprocessor to fail on missing scripts in scenes and prefabs

polson:
We had a nightly build that was broken because it was missing scripts on windows

This preprocessor will treat that as an error instead of a warning.  Normally Unity allows it to be a warning for messy day-to-day editing.  For a nightly build we shouldn't be doing this.

This scans every prefab and scene in the project, looks for components with `Script: None (MonoScript)` which would be caused by a missing .cs file and throws an exception

Only runs when making a build, not when pressing play in editor


theli-ua (approved):
Looks straightforward and reasonable

PR #1654, opened 2026-09-02
Enable dedicated open lane on 5F Guitar

Frickitickitavi:
[Corresponding YARG.Core PR](https://github.com/YARC-Official/YARG.Core/pull/450)

Enables the previously-5LK-exclusive **Dedicated Open Lane** setting for the 5F Guitar game mode.

As a related bonus, also brings these changes for BREs:
* Open inputs during BREs now generate score, albeit shared with the green scoring zone
* Speaking of scoring zones, they were apparently broken on 5F (all just one big "green" zone) - that's been fixed
* Open inputs now produce open particle effects during BREs, including on 5LK
* Because 5LK doesn't make opens animate the green BRE lane anymore, I was able to simplify a lot of the convoluted <int,int> dictionaries in FiveLaneKeysPlayer
* Renamed various methods and other identifiers for clarity and consistency, like renaming the BRE-themed `InitializeSpawnedLane` to `InitializeBRELane`

### DIFF
--- a/Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
+++ b/Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
@@ -30,19 +30,21 @@ namespace YARG.Gameplay.Player
 
         private const int SHIFT_INDICATOR_MEASURES_BEFORE = 5;
 
-        public const int LANE_COUNT = 5;
+        public bool UsingOpenLane { get; private set; }
 
-        private static Dictionary<GuitarAction, FiveFretGuitarFret> _actionToFret = new() {
-            { GuitarAction.Fret1,    FiveFretGuitarFret.Green},
-            { GuitarAction.Fret2,     FiveFretGuitarFret.Red},
-            { GuitarAction.Fret3,   FiveFretGuitarFret.Yellow},
-            { GuitarAction.Fret4,     FiveFretGuitarFret.Blue},
-            { GuitarAction.Fret5,   FiveFretGuitarFret.Orange},
+        private static Dictionary<int, FiveFretGuitarFret> _actionToFret = new() {
+            { YargFiveFretGuitarEngine.OPEN_BRE_INPUT, FiveFretGuitarFret.Open },
+            { (int)GuitarAction.Fret1,                 FiveFretGuitarFret.Green },
+            { (int)GuitarAction.Fret2,                 FiveFretGuitarFret.Red },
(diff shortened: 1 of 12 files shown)


Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
```diff
@@ -392,7 +389,18 @@ private void UpdateFretArray()
...
+
+            if (UsingOpenLane) {
+                var openInputDelta = GameManager.VisualTime - _fretToMostRecentTime[FiveFretGuitarFret.Open];
+
+
+                _fretArray.SetPressedDrum(
+                    (int) FiveFretGuitarFret.Open,
+                    _openSustaining || (openInputDelta < DEDICATED_OPEN_FLASH_HOLD_DURATION),
+                    Fret.AnimType.CorrectNormal
+                );
```
Supersonicboss1: Why are we calling `SetPressedDrum` here and not `SetPressed`? If this is intentional, can you add a comment explaining why?

Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
```diff
@@ -873,20 +954,34 @@ private void SetDefaultActiveFrets()
...
             }
         }
+
+        // It's not possible to hold an open note (no, holding the strum bar down doesn't count),
+        // so the dedicated open lane needs the same visual trick we use on drums to make it look
+        // fully lit when hit
+        private void ZeroOutOpenHitTime()
+        {
+
+        }
```
Supersonicboss1: This function is unused, so I'm not clear on what the trick mentioned here is (is it the `SetPressedDrum` thing I mentioned earlier?)

Supersonicboss1 (changes requested):
Really just one minor thing, everything else seems fine.

Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
Frickitickitavi: Oh right, I meant to rename this to `SetPressedImpulse` or something since it's no longer drum-specific. Will fix

Assets/Script/Gameplay/Player/FiveFretGuitarPlayer.cs
Frickitickitavi: Oh whoops, that's a vestige from an earlier attempt at the flashing thing - will remove

PR #1656, opened 2026-09-03
Added editable vocal colors

TheFatBastid:
This should add the ability to edit the colors of vocals, harm1, harm2, and harm3 from a color profile.
This changes the background bars and the note/talkie colors.
Double check the shader stuff as I don't know that stuff super well.

Here, I changed harm2/3 to red, green respectively:
image


XaiaX:
What does the null preview mode look like in the presets editor? (Assuming I understand that block correctly)

Getting even more reason to develop a more robust preset editor page, maybe as a dedicated page in the UI with a "real" highway.

TheFatBastid:
> What does the null preview mode look like in the presets editor? (Assuming I understand that block correctly)
> 
> Getting even more reason to develop a more robust preset editor page, maybe as a dedicated page in the UI with a "real" highway.

It just refreshes and displays whatever other instrument was previously selected, default being guitar. The vocal track isn't shown, I kinda lost the plot with that part of it so changing vocal colors doesn't show in preview at all.

XaiaX:
Yeah I felt similar when I was tweaking the pro keys preview and gave up on trying to figure out how to add the keyboard strike line.
--edit--
I wonder if for the vocals preview we could just have a static, layered screenshot where we color some basic shapes / fill rectangles with the chosen colors. Wouldn't need animation for that.

XaiaX:
Does it have a way to set the background/star power texture the way the other instruments do?

PR #1657, opened 2026-09-03
Only recommend songs playable by all connected profiles.

neilweber:
As a drums only player, I sometimes get recommendations for songs that have no drum charts.  This fixes that.


polson:
Ignore the failing checks on this PR: The issue is pointing to `master`  means it will build the native libs because they show up as a change.  Then it failed on macos because the github macos builder got updated recently which causes a byte mismatch.  So both things happened, we can just ignore the error

PR #1665, opened 2026-09-07
6f visual fixes

theli-ua:
* removes black/white fret color in color preset in favor of just "fret" (but don't fret, inner is still 2 colors)
* allows to set open note colors
* plays correctly colored hit effects
* fixes sustain coloring
* removes wildcard from 6f preview

needs corresponding core pr merged


Supersonicboss1:
Hit particles only play for white/barre notes. Is this intentional? Edit: they do have particles, but they are almost completely impossible to see against the highway. I don't really see the point in having invisible particles, so IMO either don't play them at all to save resources, or choose a lighter color (i.e. a grey of some kind) to make them more visible.

theli-ua:
> Hit particles only play for white/barre notes. Is this intentional? Edit: they do have particles, but they are almost completely impossible to see against the highway. I don't really see the point in having invisible particles, so IMO either don't play them at all to save resources, or choose a lighter color (i.e. a grey of some kind) to make them more visible.

A different pr changes default theme's color from pure white to very dark grey that will address this issue. 
This just applies color consistently. If someone chooses to use red/green theme this will fix it.right now it's broken.

PR #1673, opened 2026-09-09
Implement track player menu

polson:
This adds a per-player menu that slides up when holding the pause button for that player.

Right now there is one option to drop out.

Drop out option is only enabled if there is more than one player.  If there is only one player, the option is disabled (greyed out)

Upon dropping out, the score is affected as follows:
  - Their highway drops down and their instrument audio stays unmuted.
  - Their solo score is not saved and will not count as a high score or full combo.
  - Their stats up to the moment they left still show on the end-of-song score screen.
  - The disconnected player is removed from the fail meter so they won't fail the band.
  - They stop contributing points and combo to the band score.
  - The remaining players can finish the song and still save a valid band score.

image

https://github.com/user-attachments/assets/32194598-16b5-418b-94fe-eb87df6fb63b


wyrdough (approved):
Nothing seems unreasonable to me at a glance

PR #1674, opened 2026-09-09
Add orthogonal projection to camera preset

theli-ua:
Core.PR:  https://github.com/YARC-Official/YARG.Core/pull/461


Assets/Prefabs/Gameplay/Visual/VocalTrack.prefab
wyrdough, line 1: Why prefab changes?

theli-ua: To move vocal track semantically "closer" to the screen than any of the highways, so that it's always drawn on top

wyrdough: ..we probably shouldn't allow the highway to extend all the way behind the vocals highway, but whatever

theli-ua: We didn't used to btw, but then we merged vocals/highways cameras

PR #1679, opened 2026-09-12
Add a setting to allow a user-directed VLC.app or libVLC path

taynaron:
Building on @theli-ua's VLC work, allow the user to set a path other than the default for looking for a VLC lib. Can use .app or lib on MacOS. Doesn't try to be "smart" or "tricky" about anything; if it's not in whatever the plugin thinks is the default path the user must set the path for VLC to be used.

Does involve one little workaround: The MacOS VLC plugin has a bit of weird pathing code in it that we're bypassing if the default path isn't used. Also, tried to make the path setting a bit resilient to players changing libs without a restart, but it's not 100% so added a "may need a restart" warning.

Screenshot 2026-09-12 at 12 01 27 PM
Screenshot 2026-09-12 at 12 01 56 PM
Screenshot 2026-09-12 at 12 02 15 PM
Screenshot 2026-09-12 at 12 43 52 PM

### DIFF
--- a/Assets/Prefabs/Menu/Settings/Visuals/FolderPathSetting.prefab
+++ b/Assets/Prefabs/Menu/Settings/Visuals/FolderPathSetting.prefab
@@ -1,0 +1,618 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1 &4184945013568993316
+GameObject:
+  m_ObjectHideFlags: 0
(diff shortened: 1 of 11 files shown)


theli-ua:
> Also, tried to make the path setting a bit resilient to players changing libs without a restart, but it's not 100% so added a "may need a restart" warning.

I would just skip this for simplicity. Just say that restart is needed

PR #1680, opened 2026-09-12
Make Recommendations follow Filter Results

Dansla116:
## Description

Make Recommendations follow the selected filters, including the persistent **Only Show Playable Songs** filter.  Refresh recommendations when `filter` or `game mode` changes.

This supersedes #1657 by incorporating its functionality into the broader filtering system.

## Screenshots
Recommendations with only the "Only Show Playable Songs" Filter
image

Recommendations after filtering out Rock Band Network
image


neilweber:
So with this change I'll never see a recommended song that I can't play with my instrument?

neilweber:
Ok.  I see that "Only Show Playable Songs" settings persists across sessions.  I don't why anybody would ever want to see charts they can't play, but functionally this should work.

I have a complete rewrite of RecommendedSongs.cs that is a whole lot cleaner than the existing weird implementation.  It also brought in Favorites (which I don't think existed when RecommendedSongs was first written) and added deduplication so the same song will never be recommended multiple times in a single session.  In addition, I envisioned allowing the user to configure how much each of the different song sources (i.e. Yarg source, most played, Favorites, etc) contributed to the recommended list.  That wasn't implemented but the code is structured to make that an easy modification.  I was waiting on my other PR to be approved and merged before submitting these changes.  I've attached my fully updated RecommendedSongs.cs.
[RecommendedSongs.txt](https://github.com/user-attachments/files/32153094/RecommendedSongs.txt)

Dansla116:
Thanks for sharing this!  The source-based structure is much cleaner, and Favorites plus session-level variety would be good additions.  My current PR differs in that `Recommendations` use the complete active `Filters` predicate, so playability is only required when "Only Show Playable Songs" is enabled, and all other selected filters are honored as well.  It also builds the eligible pool before selection to avoid retry limits and searches beyond the global top ten for eligible most-played songs.  I think your source structure could integrate well with that eligibility model, either in this PR or a follow-up.

neilweber:
Yeah, applying the Filters predicate on top of the rewritten code wouldn't be hard.  I'll let you decide which direction to take.  My goal was to eliminate a few easily fixable minor annoyances.

neilweber:
I found a bug in SortFilter.cs.  I normally never set filters but because of this PR I set the "Only Show Playable Songs" flag.  Since then, I've noticed that when going into Quickplay instead of the selection being positioned at "Playlists" it's positioned at the first song in my library.

Claude tracked the problem down to SortFilter.cs:

Right after the list rebuilds, selection is correctly sitting on the Playlists button (index 0) with no _currentSong — exactly right. But shouldApplyFilters=True (you have an active Filters-menu predicate — some filter is currently engaged), so EnsureValidSelectionAfterFilter() runs, and its check:

if (SelectedIndex < 0 || SelectedIndex >= ViewList.Count ||
    CurrentSelection is not SongViewType)   // <-- Playlists button trips this
{
    if (SetIndexTo(i => i is SongViewType, _primaryHeaderIndex))
        return;                              // forces jump to the first real song
    ...
}

It treats "the current selection isn't a SongViewType" as invalid, which is meant to catch "your previously-selected song got filtered out" — but it also catches the perfectly valid case of intentionally sitting on the Playlists or Recommended button. So any time you have a filter active in the Filters menu, this method yanks focus off Playlists and onto the first real song, then that leftover _currentSong short-circuits your later "default to Playlists" logic too.

Fix: don't treat a button/header selection as invalid — it's not subject to song filtering at all:

Update(Assets\Script\Menu\MusicLibrary\MusicLibraryMenu.SortFilter.cs)
Added 7 lines
                return;
            }

            // Playlists/Recommended entries aren't subject to song filters, so a selection
            // sitting on one of them is never invalidated by a filter change.
            if (CurrentSelection is ButtonViewType)
            {
                return;
            }

I've verified that it fixes the problem I'm seeing but did no more testing beyond that.

Dansla116:
Huh, it turns out this was already an issue in **Nightly**, and the problem code is actually in **Stable** too, although I couldn't get it to reproduce there.

If you have any filter active and the selection is sitting on one of the buttons above the song list, such as Playlists or Recommended, `EnsureValidSelectionAfterFilter()` treats that selection as invalid because it isn't a song and moves you down to the first song.

That became much easier to run into with #1640 because "Only Show Playable Songs" is now a persistent filter, and even more so with this PR because we no longer automatically turn "Show Recommendations" off when a filter is active.

All this to say regardless, this looks like a clean fix, so I rebased the PR on **dev** and incorporated your change.

Issue #1133, opened 2025-09-17
video_start_time not working on Linux


carabalb:
I have a song, a song.ini file, and a video.webm file all in the same directory. When the song starts, the video plays but the video_start_time setting has no effect. I've tried both positive an negative values (eg. video_start_time = 40000)

Tested on Fedora 42 using the latest v0.13.0 stable build.

ChristianJohnsten:
I can confirm this is also happening for me on Fedora 43 using the latest v0.15.0 stable build. Would really love to have this working for linux so that we don't need to trim or pad videos to sync them up with the song.

Issue #374, opened 2023-05-24
Video backgrounds loop when they end


MatharuG:
Hello again! Well, the title is pretty much self-explanatory, but to get into more detail: if the video background (video.mp4) is shorter than the chart, it starts over. I don't know if that is intentional or not, and that's why I'm creating an Issue.
I know video backgrounds are still very new, so I'd like to suggest two different "types" of video backgrounds: 

The ones that loop (like in CH, where they're supposed to loop, and are put in a Custom folder):
![image](https://github.com/YARC-Official/YARG/assets/26250877/d40f1726-6439-4654-9baf-54362ac4ccfc)

And the ones that don't loop (the song-specific ones.)

This could also apply to images, although that would require a priority system, such as
(Venue) > Video > Image > No BG

Once again, I don't have much experience with Unity, but I did some research and found that VideoPlayer.Unity does have a variable for looping: `videoPlayer.isLooping`.

===

While I'm at the topic of videos, I also would like to suggest for YARG to use the `video_start_time` variable that `song.ini` has, since at the moment, the videos start right away. `video_start_time` basically tells the game where to start the video (in milliseconds).
Unity has a similar "offset" variable to `video_start_time`, which is `videoPlayer.frame`, but it uses the amount of frames instead of milliseconds to define how much it skips of the video.

ok that's all, thanks!

EliteAsian123:
Yeah, `video_start_time` is actually cached, I just keep forgetting to implement it... That's my bad.

As for the other stuff, we'll probably be implementing more advanced settings soon!

TheNathannator:
This issue has been fixed for a while now, apologies for the delay!

- Global backgrounds will always loop.
- Song-specific video backgrounds will *not* loop if a video start/end time (other than 0 and <= 0 respectively) is specified, otherwise it will loop if its length is less than 85% of the song's.

PR #1681, opened 2026-09-13
Don't stop the Unity VideoPlayer right after it finishes preparing

matiasba:
`OnVideoPrepared()` calls `player.Stop()` before reading `player.length` and before seeking to the configured start offset.

That stop is only needed for the VLC backend, where `VLCMediaPlayer.OpenAsync()` starts playback as soon as the media opens. The Unity VideoPlayer does not auto-play (`playOnAwake` is disabled on the component), so there is nothing to stop — and stopping it returns it to an unprepared state, where `length` reports 0 and seeks are silently dropped.

On that path, that means:

- **Songs with a non-zero `video_start_time` never show their video.** The loop branch is skipped, `_videoEndTime` is set to `player.length` (0), and `Update()` then returns early on `if (_videoEndTime == 0)` every frame, so `Play()` is never reached. Without an offset the bug is masked: the loop branch is taken instead and `_videoEndTime` becomes `NaN`, which that check doesn't match.
- **Every video is treated as loopable**, regardless of its real length.
- **`video_start_time` itself is a no-op**, since the seek is dropped.

Stopping only on the VLC path leaves that backend unchanged and lets the Unity path keep the prepared state it just acquired.

Tested on Linux with two song folders sharing the same video file via hard link and differing only by `video_start_time`. Before: only the song without an offset played. After: videos play with positive, negative and absent offsets, the offset is actually applied, and a video as long as its song is no longer marked as looping. The VLC path was also compared against an unmodified build and behaves the same.

Fixes #1133 and #374.

### DIFF
--- a/Assets/Script/Gameplay/BackgroundManager.cs
+++ b/Assets/Script/Gameplay/BackgroundManager.cs
@@ -581,8 +581,6 @@ namespace YARG.Gameplay
             const double endTimeThreshold = 0;
             const double dontLoopThreshold = 0.85;
 
-            player.Stop();
-
             if (_source == VenueSource.Song && !GameManager.Song.VideoLoop)
             {
                 _videoStartTime = GameManager.Song.VideoStartTimeSeconds;
--- a/Assets/Script/Gameplay/YargVideoPlayer.cs
+++ b/Assets/Script/Gameplay/YargVideoPlayer.cs
@@ -317,6 +317,8 @@ public class YargVideoPlayer : MonoBehaviour
             return;
         }
 
+        // OpenAsync() starts playing, so stop before handing over to the caller.
+        Stop();
 
         prepareCompleted?.Invoke(this);
     }


Assets/Script/Gameplay/BackgroundManager.cs
```diff
@@ -581,7 +581,12 @@ private void OnVideoPrepared(YargVideoPlayer player)
             const double endTimeThreshold = 0;
             const double dontLoopThreshold = 0.85;
 
-            player.Stop();
+            // VLC starts playing as soon as the media opens; the Unity player doesn't, and
+            // stopping it discards the prepared state, so length reads 0 and seeks are dropped.
+            if (player.usingVLC)
+            {
+                player.Stop();
+            }
```
theli-ua: I would suggest moving this inside YargVideoPlayer's  OnVideoPrepared (which is what calls this)
So that we keep any unity/vlc logic in that file

theli-ua (changes requested):

Assets/Script/Gameplay/BackgroundManager.cs
matiasba: Changed as requested

theli-ua (approved):

PR #1683, opened 2026-09-13
iOS port

daviditsygin:
Draft, for direction rather than merge: the iOS port as it stands, rebased onto current `dev`. It runs signed on an iPhone 13 mini and on the simulator, playing songs end to end.

**This cannot merge as-is.** `.gitmodules` points YARG.Core at a fork (daviditsygin/YARG.Core) carrying two IL2CPP fixes: managed delegates for the cache collectors, and `__Internal` P/Invokes for STB2CSharp. Those want to go upstream first, then this drops back to the real submodule.

### What's here

**Platform.** iOS player compiles: non-iOS native plugins excluded, Discord RPC compiled out, ManagedBass patched for AOT. Audio through BASS + yarg_audio, album art via a static STB2CSharp build, song import through the Files app and the native folder picker.

**Input.** MIDI via a Minis fork (threaded port scanning and an RtMidi XCFramework, PR'd separately at TheNathannator/Minis#2), Bluetooth MIDI pairing, GameController-framework devices published into the Input System, and a lane-aware on-screen guitar with a tilt whammy.

**UI.** iPhone locked to landscape, safe-area and notch insets across the menus and the gameplay HUD, native touch scroll and tap, an on-screen pause button.

**Rendering.** The venue-invisible and note-pop-in bugs were one cause: `RenderTextureFormat.DefaultHDR` has no alpha on iOS Metal, so the highway composite hid the venue and discarded the fade mask. Venues also gained an `_ios` export target, and a GPU pipeline-state warm-up removes the mid-song stutter.

### Worth arguing about

- **Binary weight.** BASS and RtMidi libraries, plus two recorded pipeline-state collections in StreamingAssets (~700 KB each, one per quality level). The collections could ship out of band.
- **Test-only code.** `AutoPlayHarness` and `FrameHitchLogger` live behind `YARG_TEST_BUILD`. Happy to pull them out if you'd rather not carry them.
- **Touch input reaches into gameplay.** `TouchGuitarInput` reads the highway's fret positions to map lanes. It works, but a cleaner seam probably exists.
- Upstream has since added its own `TogglePause`, so that part of the iOS patch is already gone.

Unverified on hardware: vocals capture (the simulator has no audio input) and TestFlight (needs a real bundle id and an App Store Connect record).


wyrdough:
We can't do this officially thanks to Unity licensing.

PR #1688, opened 2026-09-16
Further simplify fade alpha calculation

theli-ua:
We already abandoned euclidian distance from camera previously. Go even further and simply use elements' Z coordinate directly since all highways are spawned at Z=0 and aligned with z axis.

This also fixes fading application for orthogonal projection

image


Assets/Resources/HighwaysAlphaMask.shader
```diff
@@ -43,10 +43,7 @@ Shader "HighwaysAlphaMask"
                 int index = WorldPosToIndex(IN.positionWS);
                 float fadeStartPos = _YargFadeParams[index * 2];
                 float fadeEndPos   = _YargFadeParams[index * 2 + 1];
-                // Euclidean distance from camera to this fragment
-                float3 camPos = YargWorldSpaceCameraPos(IN.positionWS);
-                // float dist = distance(camPos, IN.positionWS);
-                float dist = IN.positionWS.z - camPos.z;
+                float dist = IN.positionWS.z - 2;
```
wyrdough, line 51: It seems like it would be better if this followed `TrackPlayer`'s `STRIKE_LINE_POS` somehow so we don't have to remember to change it here as well if that ever gets modified.

wyrdough (changes requested):

Assets/Resources/HighwaysAlphaMask.shader
theli-ua: Not possible unless we want even more globals

theli-ua: FWIW the whole highways.hlsl relies on specific positioning of highways already, that is exactly how it figures out which of the transformation matrices to use. So it already needs to be kept in sync manually, this doesn't really have situation that worse.

theli-ua: *make

wyrdough: Can you add small comments in both places referencing the other and the need to keep them in sync, then?

theli-ua: You bet

theli-ua: done

PR #1689, opened 2026-09-17
Write fade alpha directly into highways rendertexture

theli-ua:
Get rid of additional rt that  has to be carried around. Step towards not having to patch UberPP

This doesn't do much, this is a cleanup/refactoring mostly. Although theoretically it is more efficient but most likely on the margin of error.


wyrdough (approved):
Approved pending testing

polson (approved):

PR #1690, opened 2026-09-17
Add Configurable Album Grouping to Artist Sort

Dansla116:
## Description

Adds optional secondary album grouping when the Music Library is sorted by artist.

Albums containing at least three visible songs are displayed under non-selectable album headers. Songs from smaller album groups remain directly beneath the artist header.

Adds distinct gray styling and a reduced row height.  Uses the sidebar album icon for album headers.  Displays album-level earned and total stars.

Adds a **Secondary Artist Sort** setting under Settings -> Songs -> Music Library
- Albums by Title / Songs by Title
- Albums by Title / Songs by Track
- Albums by Year / Songs by Title
- Albums by Year / Songs by Track
- Off

Defaults to albums and songs sorted by title

This resolves: https://discord.com/channels/1086048856678084609/1257842433194983424

## Screenshots

New "Secondary Artist Sort" in Settings Menu
image

New Default "Albums by Title / Songs by Title" Option
image

"Off" Option (same as today)
image

PR #1692, opened 2026-09-18
Show Gold Star Icons for Fully Gold-Starred Song Groups

Dansla116:
## Description

Improves star-progress indicators by distinguishing between groups that have earned the maximum number of regular stars and groups in which every song has been gold-starred.  Primary headers *(and secondary headers once #1690 is merged)* now display a gold star icon only when every song in the group has Gold Stars.  Otherwise, they retain the standard white star icon.

This distinction also applies to the total number of visible songs in the library or when using filters, browsing playlists, sorting by player context, and viewing collapsed headers.

## Screenshots

Guitar Hero: Smash Hits - 40/40 Stars Earned - Some songs have **Gold Stars**, while others are only **5 Stars**
image

Guitar Hero: TV - 15/15 Stars Earned - All songs have **Gold Stars**
image

Sorting by Stars and Headers Collapsed - **Gold Stars** and **5 Stars** headers show proper icon
image

Filtered Music Library - 160/160 Stars Earned – All visible songs have **Gold Stars**, so the overall library total and each displayed header use the **Gold Star** icon
image


Assets/Script/Menu/MusicLibrary/MusicLibraryMenu.Playlist.cs
```diff
@@ -126,6 +126,7 @@ private List<ViewType> CreatePlaylistViewList()
             bool allowdupes = SettingsManager.Settings.AllowDuplicateSongs.Value;
             _totalSongCount = 0;
             _totalStarCount = 0;
+            _allVisibleSongsGold = true;
```
wyrdough: I am just curious why you decided to have gold the default rather than the other way around?

Dansla116: Good shout.  I could go either way.

wyrdough:
So I'm kinda a dick here, but I think it was better before. My brain yesterday was thinking you could just swap to `false` and `|=` instead of `true` and `&=`, but that obviously won't work if you're testing that *all* the songs in a category have been GSed.

Sorry :(

Dansla116:
Happy medium, maybe?  Assume `false`, track whether any song isn't gold, then invert the result.  And the empty groups I was guarding against shouldn't appear anyway.

PR #1693, opened 2026-09-18
Move venue post processing into a "single" standalone pass

theli-ua:
Including alpha fixup that used to be a separate pass. Motivation is to get rid of necessity to patch UberPP and move to mainline one


theli-ua:
Confirmed visually with "miku" chart with scanlines and mirror effects, confirmed in rendergraph debugger as well:

image

theli-ua:
Note that effect's code itself is unchanged, just how its integrated

polson (approved):

